Issue #5, opened 2024-02-28
Automate dependencies update


ahg-g:
Example: https://github.com/kubernetes-sigs/jobset/pull/16

PR #10, opened 2024-02-29
Add dependabot

kerthcet:
fix https://github.com/kubernetes-sigs/lws/issues/5

Also Add the Issue/PR templates. This is mostly inspired by kueue.


kerthcet:
I guess you haven't joined the org yet, so no privilege?

liurupeng:
> I guess you haven't joined the org yet, so no privilege?

yes

kerthcet:
cc @ahg-g

ahg-g:
the test is not passing though, should we fix that first?

kerthcet:
> the test is not passing though, should we fix that first?

I'm digging.

Issue #13, opened 2024-02-29
Make generate error as `fatal: Invalid path` 


kerthcet:
When we run `git clone git@github.com:kerthcet/lws.git` directly, the repo name is **lws**, and when you run `make generate`, it will report error like: 
```
fatal: Invalid path '/Users/kerthcet/Workspaces/leader-worker-set/api/**': No such file or directory
```

This is because the project path is `sigs.k8s.io/lws` not `sigs.k8s.io/leader-worker-set`.

/kind bug

PR #11, opened 2024-02-29
Project refactor work

kerthcet:
What I did with this PR:
1. bump go version to v1.21 align with kubernetes
2. Rename foler internal to pkg as kueue does
3. Remove *.out and add a .gitignore file to this project.
4. change repo name leader-worker-set to lws
5. upgrade the controller-gen to v0.14.0 to fix test error and rerun `make manifest`.

fix https://github.com/kubernetes-sigs/lws/issues/13
part of https://github.com/kubernetes-sigs/lws/issues/8

PR #14, opened 2024-02-29
Fix make generate error with invalid repo path

kerthcet:
fix https://github.com/kubernetes-sigs/lws/issues/13

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -31,8 +31,8 @@ source "${CODEGEN_PKG}/kube_codegen.sh"
 # (https://github.com/kubernetes/code-generator/issues/165).
 # Here, we create the soft link named "x-k8s.io" to the parent directory of
 # LeaderWorkerSet to ensure the layout required by the kube_codegen.sh script.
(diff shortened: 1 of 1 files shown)


ahg-g:
how about changing things to lws instead? I guess here https://github.com/kubernetes-sigs/lws/blob/9af20f61ba6d4e4ebd97b43bfa6ffca5aa618743/PROJECT#L8-L9 and that should be reflected in all other autogenerated stuff https://github.com/search?q=repo%3Akubernetes-sigs%2Flws%20%22leader-worker-set%22&type=code

kerthcet:
That's what came to my mind first - **Changing leader-workder-set to lws**, however, I think we have to change all, more than a part, maybe the test-infra as well?

ahg-g:
The test-infra seems aligned with lws: https://github.com/kubernetes/test-infra/pull/32081/files

kerthcet:
SG then.

kerthcet:
In favor of https://github.com/kubernetes-sigs/lws/pull/11.
/close

PR #16, opened 2024-02-29
added unit tests for webhook utils and controller utils

Edwinhr716:
Adding Unit tests

### DIFF
--- a/internal/controller/controller_utils_test.go
+++ b/internal/controller/controller_utils_test.go
@@ -1,0 +1,357 @@
+package controller
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	appsapplyv1 "k8s.io/client-go/applyconfigurations/apps/v1"
+	coreapplyv1 "k8s.io/client-go/applyconfigurations/core/v1"
+	metaapplyv1 "k8s.io/client-go/applyconfigurations/meta/v1"
+	"k8s.io/utils/ptr"
+	leaderworkerset "sigs.k8s.io/leader-worker-set/api/leaderworkerset/v1"
+	testutils "sigs.k8s.io/leader-worker-set/test/testutils"
+)
+
+func TestConstructWorkerStatefulSetApplyConfiguration(t *testing.T) {
+	tests := []struct {
+		name                  string
+		pod                   *corev1.Pod
+		lws                   *leaderworkerset.LeaderWorkerSet
+		wantStatefulSetConfig *appsapplyv1.StatefulSetApplyConfiguration
+	}{
+		{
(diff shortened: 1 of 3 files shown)


internal/controller/controller_utils_test.go
```diff
@@ -0,0 +1,357 @@
...
+			},
+		},
+		{
+			name: "1 replica, size 2, exclusive placement enabled",
+			pod: &corev1.Pod{
+				ObjectMeta: v1.ObjectMeta{
+					Name:      "test-sample",
+					Namespace: "default",
+					Labels: map[string]string{
+						"leaderworkerset.sigs.k8s.io/worker-index": "0",
```
ahg-g, line 103: why is this not showing in wantStatefulSetConfig?

internal/webhook/webhook_utils_test.go
```diff
@@ -0,0 +1,460 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			appliedExclusivePlacement := exclusiveAffinityApplied(tc.pod)
+			if appliedExclusivePlacement != tc.expectedAppliedExclusivePlacement {
+				t.Errorf("Expected value %t, got %t", tc.expectedAppliedExclusivePlacement, appliedExclusivePlacement)
+			}
+		})
+	}
+}
+
+func MakeLeaderPodSpec() corev1.PodSpec {
```
ahg-g, line 407: can we add this and the functions below to the wrappers.go file and apply the wrapper pattern please?

internal/controller/controller_utils_test.go
```diff
@@ -0,0 +1,357 @@
+package controller
```
liurupeng, line 1: could you put the test under the existing pkg/controller/controller_test.go

internal/controller/controller_utils_test.go
```diff
@@ -0,0 +1,357 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	appsapplyv1 "k8s.io/client-go/applyconfigurations/apps/v1"
+	coreapplyv1 "k8s.io/client-go/applyconfigurations/core/v1"
+	metaapplyv1 "k8s.io/client-go/applyconfigurations/meta/v1"
+	"k8s.io/utils/ptr"
+	leaderworkerset "sigs.k8s.io/leader-worker-set/api/leaderworkerset/v1"
+	testutils "sigs.k8s.io/leader-worker-set/test/testutils"
```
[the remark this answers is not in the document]
Edwinhr716: Done

internal/webhook/webhook_utils_test.go
Edwinhr716: The testutils package imports the webhook package, so if I import testutils it creates an import cycle. That is why I needed to copy the functions into this file

internal/controller/controller_utils_test.go
Edwinhr716: The function never adds that label to the statefulsetconfig. I can remove the label from the input if needed

liurupeng:
@Edwinhr716 I think you need to git pull the changes from main and rebase

PR #38, opened 2024-03-05
Fix Available and Processing condition should not both be true 

kerthcet:
#### What type of PR is this?
/kind bug

Fixes #

```release-note
None
```


kerthcet:
cc @RupengLiu

liurupeng:
@kerthcet nice catch

Issue #3, opened 2024-02-28
Add support for rolling updates 


ahg-g:
As discussed in https://bit.ly/k8s-LWS

kerthcet:
About maxSurge, we know that deployment is stateless, so we can burst out extra replicas for service availability, and no need to delete them in the final as long as meet the replicas. But for statefulset, it doesn't has this capability because the replicas of statefulsets is indexed, then maxSurge seems not fit for lws as well. e.g. we have a lws with 3 replicas, and the maxSurge is 2, rolling update looks like:
- replica-0 (old)
- replica-1 (old)
- replica-2 (rolling update)
- replica-3 (new)
- replica-4 (new)

Then when rolling update finished, we'll remove the replica-3 and replica-4? This seems like a waste especially for LLM. @ahg-g @liurupeng any other opinions?

liurupeng:
@kerthcet we could add maxSurge support later.  But we may need the maxSurge feature. Right now for GPU/TPU customers, one of most critical issues is hardware stockout which means the accelerators in the zone has all been used up. If the customer use maxUnavailable, with cluster-autoscaler/nodepool-auto-provisioning enabled. It's often during the rollingUpdate the hardware of the updating podgroup will be released. But to provision a new accelerator node, it will be pretty hard(due to stockout), and this may cause issues if the total number of nodes is not large. With maxSurge, this could be mitigated since we always ensure there is at least lws.spec.replicas number of pod groups.

kerthcet:
Thanks for the explanation, got the idea about **Reserve**. Several concerns come into my mind:
- This would require additional costs for rolling update, like we have to cleanup the extra replicas, or even trigger the unnecessary autoscaling
- From a system/cluster level, will this intensify the resource competition because hardware is scarce and we required extra resources
- If maxSurge is set and resource only fit for `spec.Replicas`, will this block the rolling update. But we can scale down the maxSurge to 0  to solve this.
- Maybe we should support **reserve pod** in kubernetes or at least at kube-scheduler.
- Would this fit the Stetefulset? Do we have similar scenarios?

Sorry, I need to think more deep about this feature before I start the work.

kerthcet:
I'm inclined to rename `rolloutStrategy` to `updateStrategy` to align with Statefulset and rollout may confused with https://github.com/kubernetes-sigs/lws/issues/89 in the future.

liurupeng:
[...]

PR #39, opened 2024-03-06
Feat: Rollout API

kerthcet:
#### What type of PR is this?
/kind feature

part of https://github.com/kubernetes-sigs/lws/issues/3

```release-note
Support Rollout in leaderWorkerSet.
```


kerthcet:
This is API part, you can take a look if you want. cc @ahg-g @liurupeng

ahg-g:
/lgtm

Please wait until @liurupeng reviews as well.

Dockerfile
```diff
@@ -17,7 +17,7 @@ COPY api/ api/
 COPY pkg/controller/ pkg/controller/
 COPY pkg/cert/ pkg/cert/
 COPY pkg/webhook/ pkg/webhook/
-COPY pkg/commonutils pkg/commonutils
```
liurupeng, line 20: may need to do a rebase

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -89,12 +97,71 @@ type LeaderWorkerTemplate struct {
...
+	// +kubebuilder:default=1
+	MaxUnavailable intstr.IntOrString `json:"maxUnavailable,omitempty"`
+
+	// The maximum number of replicas that can be scheduled above the original number of
+	// replicas.
+	// Value can be an absolute number (ex: 5) or a percentage of total replicas at
+	// the start of the update (ex: 10%).
+	// This can not be 0 if MaxUnavailable is 0.
+	// Absolute number is calculated from percentage by rounding up.
+	// By default, a value of 1 is used.
```
liurupeng, line 145: if we didn't specify MaxUnavailable and MaxSurge, the default value for them are both 1? which means that actually we could have 2 replicas upgrading at the same time?

liurupeng: But I think it is reasonable to have both as 1 at the same time

ahg-g:
Before we unhold, lets discuss if we want to include this in 0.1.0 release. If not, then we should wait until we release.

kerthcet:
Talked offline as we will release 0.1.0 first.

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -89,12 +97,71 @@ type LeaderWorkerTemplate struct {
...
+	// By default, a fixed value of 1 is used.
+	// Example: when this is set to 30%, the old replicas can be scaled down by 30%
+	// immediately when the rolling update starts. Once new replicas are ready, old replicas
+	// can be scaled down further, followed by scaling up the new replicas, ensuring
+	// that at least 70% of original number of replicas are available at all times
+	// during the update.
+	//
+	// +kubebuilder:validation:XIntOrString
+	// +kubebuilder:default=1
+	MaxUnavailable intstr.IntOrString `json:"maxUnavailable,omitempty"`
```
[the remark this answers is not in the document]
liurupeng: yep, this is good

ahg-g:
Is there an implementation PR in the works? if so, I guess we can merge this.

kerthcet:
Yes, the implementation for `maxUnavailable & rollout` has already implemented, I'm still adding more testcases and test locally. 

For maxSurge, I'll leave for next step to reduce the review burden. We can merge now. 

/hold cancel

PR #61, opened 2024-03-18
Fix TPU env var assignment when leader doesn't request TPUs

ahg-g:
#### What type of PR is this?
/kind bug

LWS injects TPU env vars necessary for multi-host workloads, this includes TPU_WORKER_ID and TPU_HOST_NAMES. Currently we inject that into the workers if they request TPUs assuming that the leader does too, and so TPU_WORKER_ID=0 is always assumed to be the leader pod, and the leader hostname is also added to TPU_HOST_NAMES list. 

In some cases, the leader may not be requesting TPUs and is not acting as a TPU worker, in which case its hostname should not be listed in the host names and the TPU_WORKER_ID should be assigned to the LWS workers starting from 0 (and so TPU_WORKER_ID = LWS_WORKER_INDEX-1)

Fixes #

```release-note

```

### DIFF
--- a/pkg/utils/accelerators/tpu.go
+++ b/pkg/utils/accelerators/tpu.go
@@ -22,14 +22,15 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 
-	leaderworkerset "sigs.k8s.io/lws/api/leaderworkerset/v1"
+	podutils "sigs.k8s.io/lws/pkg/utils/pod"
 	statefulsetutils "sigs.k8s.io/lws/pkg/utils/statefulset"
 )
 
 var (
-	TpuResourceName    = corev1.ResourceName("google.com/tpu")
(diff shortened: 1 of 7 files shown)


ahg-g:
/hold 

just so it doesn't get merged by mistake

pkg/utils/accelerators/tpu.go
```diff
@@ -89,38 +90,49 @@ func AddTPUVariables(pod *corev1.Pod, size int) error {
...
 		corev1.EnvVar{
 			Name:  TpuWorkerId,
-			Value: pod.Labels[leaderworkerset.WorkerIndexLabelKey],
+			Value: fmt.Sprint(tpuWorkerId),
 		},
 	)
 	return nil
 }
+
+// AddTPUAnnotations adds TPU specific annotations.
```
liurupeng, line 133: where is this function invoked?

PR #65, opened 2024-03-19
Support maxUnavailable in rollout

kerthcet:
#### What type of PR is this?
/kind feature

part of https://github.com/kubernetes-sigs/lws/issues/3

```release-note
Support maxUnavailable in rollout
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -240,7 +240,7 @@ $(CONTROLLER_GEN): $(LOCALBIN)
 .PHONY: envtest
 envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(LOCALBIN)
-	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20240320141353-395cfc7486e6
 
 GINKGO = $(shell pwd)/bin/ginkgo
 .PHONY: ginkgo
--- a/Makefile
+++ b/Makefile
@@ -240,7 +240,7 @@ $(CONTROLLER_GEN): $(LOCALBIN)
 .PHONY: envtest
 envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(LOCALBIN)
-	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20240320141353-395cfc7486e6
+	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
 GINKGO = $(shell pwd)/bin/ginkgo
 .PHONY: ginkgo
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -23,23 +23,27 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	appsapplyv1 "k8s.io/client-go/applyconfigurations/apps/v1"
 	coreapplyv1 "k8s.io/client-go/applyconfigurations/core/v1"
 	metaapplyv1 "k8s.io/client-go/applyconfigurations/meta/v1"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
-	"k8s.io/utils/ptr"
+	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	leaderworkerset "sigs.k8s.io/lws/api/leaderworkerset/v1"
+	"sigs.k8s.io/lws/pkg/utils"
 	podutils "sigs.k8s.io/lws/pkg/utils/pod"
+	statefulsetutils "sigs.k8s.io/lws/pkg/utils/statefulset"
 )
 
 // LeaderWorkerSetReconciler reconciles a LeaderWorkerSet object
@@ -50,13 +54,20 @@ type LeaderWorkerSetReconciler struct {
 }
 
 var (
+	apiGVStr = leaderworkerset.GroupVersion.String()
+)
+
+const (
 	lwsOwnerKey  = ".metadata.controller"
-	apiGVStr     = leaderworkerset.GroupVersion.String()
-	fieldManager = "lws"
+	fieldManager = "leader-worker-set"
 )
 
 func NewLeaderWorkerSetReconciler(client client.Client, scheme *runtime.Scheme, record record.EventRecorder) *LeaderWorkerSetReconciler {
-	return &LeaderWorkerSetReconciler{Client: client, Scheme: scheme, Record: record}
+	return &LeaderWorkerSetReconciler{
+		Client: client,
+		Scheme: scheme,
+		Record: record,
+	}
 }
 
 //+kubebuilder:rbac:groups="",resources=events,verbs=create;watch;update;patch
@@ -77,44 +88,14 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	log := ctrl.LoggerFrom(ctx).WithValues("leaderworkerset", klog.KObj(lws))
 	ctx = ctrl.LoggerInto(ctx, log)
 
-	// construct the statefulset apply configuration
-	leaderStatefulSetApplyConfig, err := constructLeaderStatefulSetApplyConfiguration(lws)
+	partition, err := r.rollPartition(ctx, lws)
 	if err != nil {
-		log.Error(err, "Constructing StatefulSet apply configuration.")
-		return ctrl.Result{}, err
-	}
-	if err := setControllerReferenceWithStatefulSet(lws, leaderStatefulSetApplyConfig, r.Scheme); err != nil {
-		log.Error(err, "Setting controller reference.")
-		return ctrl.Result{}, nil
-	}
-	obj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(leaderStatefulSetApplyConfig)
-	if err != nil {
-		log.Error(err, "Converting StatefulSet configuration to json.")
-		return ctrl.Result{}, err
-	}
-	patch := &unstructured.Unstructured{
-		Object: obj,
-	}
-	// Use server side apply and add fieldmanager to the lws owned fields
-	// If there are conflicts in the fields owned by the lws controller, lws will obtain the ownership and force override
-	// these fields to the ones desired by the lws controller
-	// TODO b/316776287 add E2E test for SSA
-	err = r.Patch(ctx, patch, client.Apply, &client.PatchOptions{
-		FieldManager: fieldManager,
-		Force:        ptr.To[bool](true),
-	})
-	if err != nil {
-		log.Error(err, "Using server side apply to update leader statefulset")
+		log.Error(err, "rolling partition error")
 		return ctrl.Result{}, err
 	}
 
-	stsSelector := client.MatchingLabels(map[string]string{
-		leaderworkerset.SetNameLabelKey: lws.Name,
-	})
-	var lwssts appsv1.StatefulSetList
-	if err := r.List(ctx, &lwssts, stsSelector, client.InNamespace(lws.Namespace)); err != nil {
-		log.Error(err, "Fetching statefulsets managed by leaderworkerset instance")
-		return ctrl.Result{}, nil
+	if err := r.SSAWithStatefulset(ctx, lws, partition); err != nil {
+		return ctrl.Result{}, err
 	}
 
 	// Create headless service if it does not exist.
@@ -200,6 +181,139 @@ func SetupIndexes(indexer client.FieldIndexer) error {
 	})
 }
 
+// leaderTemplate() here indicates the leaderTemplate or the workerTemplate
+// if leaderTemplate is nil.
+// - if leader template updated, we'll roll out the whole lws.
+// - if leader template is nil, and worker template updated, we'll roll out the whole lws.
+// - if leader template is nil, but we update with
+
+// Rolling update will always wait for the former replica ready then process the next one,
+// we didn't consider rollout strategy type here for we only support rollingUpdate now,
+// once we have more policies, we should update the logics here.
+// Possible scenarios:
+//   - When sts is under creation, partition is always 0 because pods are created in parallelism, rolling update is not fit here.
+//   - When sts is rolling update, the partition will start from the last index to the 0 index step by maxUnavailable.
+//   - When sts is rolling update, replicas scaled up, the partition will not change until scaling ended.
+//   - When sts is rolling update, replicas scaled down, the partition will not change unless (replicas-scaleDownedSize) < partition,
+//     partition will be reset to (replicas-scaleDownedSize).
+//
+// When all set, the partition should always be zero.
+func (r *LeaderWorkerSetReconciler) rollPartition(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet) (int32, error) {
+	sts := &appsv1.StatefulSet{}
+	err := r.Get(ctx, types.NamespacedName{Name: lws.Name, Namespace: lws.Namespace}, sts)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return 0, err
+	}
+
+	// If sts not created yet, all partitions should be updated.
+	if apierrors.IsNotFound(err) {
+		return 0, nil
+	}
+
+	partition := *sts.Spec.UpdateStrategy.RollingUpdate.Partition
+	replicas := *lws.Spec.Replicas
+	rollingStep, err := intstr.GetValueFromIntOrPercent(&lws.Spec.RolloutStrategy.RollingUpdateConfiguration.MaxUnavailable, int(replicas), false)
+	if err != nil {
+		return 0, err
+	}
+
+	// Indicates a new rolling update here.
+	if templateUpdated(sts, lws) {
+		if replicasUpdated(sts, lws) {
+			return min(*lws.Spec.Replicas, *sts.Spec.Replicas), nil
+		} else {
+			return replicas - int32(rollingStep), nil
+		}
+	}
+
+	// Otherwise always return 0.
+	if partition == 0 {
+		return 0, nil
+	}
+
+	stsSelector := client.MatchingLabels(map[string]string{
+		leaderworkerset.SetNameLabelKey: lws.Name,
+	})
+	var stsList appsv1.StatefulSetList
+	if err := r.List(ctx, &stsList, stsSelector, client.InNamespace(lws.Namespace)); err != nil {
+		return 0, err
+	}
+
+	var finishedPart int32
+	templateHash := utils.LeaderWorkerTemplateHash(lws)
+stop:
+	for i := replicas - 1; i >= 0; i-- {
+		for j := len(stsList.Items) - 1; j >= 0; j-- {
+			nominatedName := fmt.Sprintf("%s-%d", lws.Name, i)
+			if stsList.Items[j].Name == nominatedName {
+				// TODO: quick sort the pods and statefulsets first, which will be faster than API requests.
+				// Or only construct workerStatefulset until leaderPod is ready.
+
+				// Rolling update happens only when the former group of pods(leaderPod+workerStatefulset) are all ready
+				// then we'll go process the next group.
+				// Compare the templateRevisionHash to make sure that we're checking the right object.
+
+				stsTemplateHash := stsList.Items[j].Labels[leaderworkerset.TemplateRevisionHashKey]
+				if !(stsTemplateHash == templateHash && statefulsetutils.StatefulsetReady(stsList.Items[j])) {
+					break stop
+				}
+
+				pod := &corev1.Pod{}
+				if err := r.Get(ctx, types.NamespacedName{Name: nominatedName, Namespace: lws.Namespace}, pod); err != nil {
+					return 0, err
+				}
+
+				podTemplateHash := pod.Labels[leaderworkerset.TemplateRevisionHashKey]
+				if !(podTemplateHash == templateHash && podutils.PodRunningAndReady(*pod)) {
+					break stop
+				}
+
+				finishedPart++
+				break
+			}
+		}
+	}
+
+	return utils.NonZeroValue(replicas - int32(rollingStep) - finishedPart), nil
+}
+
+func (r *LeaderWorkerSetReconciler) SSAWithStatefulset(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, partition int32) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	// construct the statefulset apply configuration
+	leaderStatefulSetApplyConfig, err := constructLeaderStatefulSetApplyConfiguration(lws, partition)
+	if err != nil {
+		log.Error(err, "Constructing StatefulSet apply configuration.")
+		return err
+	}
+	if err := setControllerReferenceWithStatefulSet(lws, leaderStatefulSetApplyConfig, r.Scheme); err != nil {
+		log.Error(err, "Setting controller reference.")
+		return err
+	}
+	obj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(leaderStatefulSetApplyConfig)
+	if err != nil {
+		log.Error(err, "Converting StatefulSet configuration to json.")
+		return err
+	}
+	patch := &unstructured.Unstructured{
+		Object: obj,
+	}
+	// Use server side apply and add fieldmanager to the lws owned fields
+	// If there are conflicts in the fields owned by the lws controller, lws will obtain the ownership and force override
+	// these fields to the ones desired by the lws controller
+	// TODO b/316776287 add E2E test for SSA
+	err = r.Patch(ctx, patch, client.Apply, &client.PatchOptions{
+		FieldManager: fieldManager,
+		Force:        pointer.Bool(true),
+	})
+	if err != nil {
+		log.Error(err, "Using server side apply to update leader statefulset")
+		return err
+	}
+
+	return nil
+}
+
 // updates the condition of the leaderworkerset to either Progressing or Available.
 func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet) (bool, error) {
 	log := ctrl.LoggerFrom(ctx)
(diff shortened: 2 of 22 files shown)


kerthcet:
Main changes:
- The general mechanism is we'll have a `templateRevisionHash` label for leaderStatefulst, leaderPod, workerStatefulset, workerPod, it will be used to verify that this is the new revision, so when leaderTemplate or workerTemplate is updated, we'll trigger a rolling update of the leader statefulset, which means a group of pods(leader+worker) will be updated the same time.
- The rolling update number of replicas is defined by `maxUnavailable`, which means when we set `maxUnavailable` to 2, and we have a lws with 5 replicas, the rolling update index will looks like: `4,3 -> 2,1, -> 0`.

Other updates:
- the maxUnavailable is default to 1 because 0 is not allowed in Statefulset
- the maxSurge(implementation not included in this PR) is default to 0, then the default behavior will rolling update a replica each time, I think this is more cautious.

Another concern is about the `leaderWorkerTemplate.size`, I have an implementation locally,  it behaves as when we increate or decrease the size, it will not lead to the statefulset rollingupdate, only scalilng up/down the worker pods. But I may need your inputs for this, and this is now conflict with several things:
- we have a size annotation in the leaderPod, which will lead to the statefulset rolling update each time we change the size
- the restart policy `RecreateGroupOnPodRestart` is somehow conflict with decreasing the size because when a pod is deleted, the leaderPod will also be deleted. 

so **I remove the rolling update for size update** right now, and make it immutable as original.

Please take a look and sorry for the slow response because I'm right at kubecon last week. cc @ahg-g @liurupeng

kerthcet:
/hold cancel
we may need e2e test as well. I'll add the e2e test framework later.

liurupeng:
@kerthcet we have someone working on the e2e test already, that work is in progress, thanks!

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -200,6 +181,139 @@ func SetupIndexes(indexer client.FieldIndexer) error {
...
+	replicas := *lws.Spec.Replicas
+	rollingStep, err := intstr.GetValueFromIntOrPercent(&lws.Spec.RolloutStrategy.RollingUpdateConfiguration.MaxUnavailable, int(replicas), false)
+	if err != nil {
+		return 0, err
+	}
+
+	// Indicates a new rolling update here.
+	if templateUpdated(sts, lws) {
+		if replicasUpdated(sts, lws) {
+			return min(*lws.Spec.Replicas, *sts.Spec.Replicas), nil
```
liurupeng, line 223: Probably add a comment here:
"When scaling and upgrade happened at the same time, scale up will have extra replicas created with new template. Scale down will decrease the number of replica at first and then trigger rollingUpdate."

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -200,6 +181,139 @@ func SetupIndexes(indexer client.FieldIndexer) error {
...
+	templateHash := utils.LeaderWorkerTemplateHash(lws)
+stop:
+	for i := replicas - 1; i >= 0; i-- {
+		for j := len(stsList.Items) - 1; j >= 0; j-- {
+			nominatedName := fmt.Sprintf("%s-%d", lws.Name, i)
+			if stsList.Items[j].Name == nominatedName {
+				// TODO: quick sort the pods and statefulsets first, which will be faster than API requests.
+				// Or only construct workerStatefulset until leaderPod is ready.
+
+				// Rolling update happens only when the former group of pods(leaderPod+workerStatefulset) are all ready
```
liurupeng, line 252:
```suggestion
				// Rolling update happens only when all the former groups of pods(leaderPod+workerStatefulset) are all ready
```

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -200,6 +181,139 @@ func SetupIndexes(indexer client.FieldIndexer) error {
...
+					break stop
+				}
+
+				finishedPart++
+				break
+			}
+		}
+	}
+
+	return utils.NonZeroValue(replicas - int32(rollingStep) - finishedPart), nil
```
liurupeng, line 277: the upgrade will stop when any of the updated group is not ready?

pkg/utils/utils.go
```diff
@@ -0,0 +1,43 @@
...
+
+func NonZeroValue(value int32) int32 {
+	if value < 0 {
+		return 0
+	}
+	return value
+}
+
+func LeaderWorkerTemplateHash(lws *leaderworkerset.LeaderWorkerSet) string {
+	return Sha1Hash(lws.Spec.LeaderWorkerTemplate.LeaderTemplate.String() +
```
liurupeng, line 41: Should update to the RestartPolicy triggers the rolling update? what do you think? @kerthcet @ahg-g

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -200,6 +181,139 @@ func SetupIndexes(indexer client.FieldIndexer) error {
...
+	if err != nil {
+		log.Error(err, "Using server side apply to update leader statefulset")
+		return err
+	}
+
+	return nil
+}
+
 // updates the condition of the leaderworkerset to either Progressing or Available.
 func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet) (bool, error) {
```
liurupeng, line 318: the lws.status.updatedReplica is in the next PR?

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -200,6 +181,139 @@ func SetupIndexes(indexer client.FieldIndexer) error {
...
+				// Rolling update happens only when the former group of pods(leaderPod+workerStatefulset) are all ready
+				// then we'll go process the next group.
+				// Compare the templateRevisionHash to make sure that we're checking the right object.
+
+				stsTemplateHash := stsList.Items[j].Labels[leaderworkerset.TemplateRevisionHashKey]
+				if !(stsTemplateHash == templateHash && statefulsetutils.StatefulsetReady(stsList.Items[j])) {
+					break stop
+				}
+
+				pod := &corev1.Pod{}
```
liurupeng, line 261:
```suggestion
				leaderPod := &corev1.Pod{}
```
this would be slightly more straightforward

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -200,6 +181,139 @@ func SetupIndexes(indexer client.FieldIndexer) error {
...
+				if err := r.Get(ctx, types.NamespacedName{Name: nominatedName, Namespace: lws.Namespace}, pod); err != nil {
+					return 0, err
+				}
+
+				podTemplateHash := pod.Labels[leaderworkerset.TemplateRevisionHashKey]
+				if !(podTemplateHash == templateHash && podutils.PodRunningAndReady(*pod)) {
+					break stop
+				}
+
+				finishedPart++
```
liurupeng, line 271: we calculate the upgraded and ready groups here. If any updated group suddenly failed, for example the first upgraded group is not ready, will the below calculation makes the upgrade to start from scratch again?

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -200,6 +181,139 @@ func SetupIndexes(indexer client.FieldIndexer) error {
...
+	var stsList appsv1.StatefulSetList
+	if err := r.List(ctx, &stsList, stsSelector, client.InNamespace(lws.Namespace)); err != nil {
+		return 0, err
+	}
+
+	var finishedPart int32
+	templateHash := utils.LeaderWorkerTemplateHash(lws)
+stop:
+	for i := replicas - 1; i >= 0; i-- {
+		for j := len(stsList.Items) - 1; j >= 0; j-- {
```
liurupeng, line 246: qq, why we traverse the stsList from the last to the first one as well?

Makefile
```diff
@@ -221,7 +221,7 @@ $(CONTROLLER_GEN): $(LOCALBIN)
 .PHONY: envtest
 envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(LOCALBIN)
-	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20240320141353-395cfc7486e6
```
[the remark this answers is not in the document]
kerthcet: The latest setup-envtest requires go1.22, but seems we already updated to the 1.22. Let me revert this.

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: gcr.io/k8s-staging-lws/lws
-  newTag: main
+  newName: registry.cn-shanghai.aliyuncs.com/kerthcet-public/lws
```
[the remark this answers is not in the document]
kerthcet: Seems you're reviewed at an old version, solved already.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -200,6 +181,139 @@ func SetupIndexes(indexer client.FieldIndexer) error {
 	})
 }
 
+// leaderTemplate() here indicates the leaderTemplate or the workerTemplate
+// if leaderTemplate is nil.
+// - if leader template updated, we'll roll out the whole lws.
+// - if leader template is nil, and worker template updated, we'll roll out the whole lws.
+// - if leader template is nil, but we update with
```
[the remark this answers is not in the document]
kerthcet: Updated.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -320,44 +436,51 @@ func constructLeaderStatefulSetApplyConfiguration(lws *leaderworkerset.LeaderWor
...
 	podTemplateApplyConfiguration.WithAnnotations(podAnnotations)
+
 	// construct statefulset apply configuration
 	statefulSetConfig := appsapplyv1.StatefulSet(lws.Name, lws.Namespace).
 		WithSpec(appsapplyv1.StatefulSetSpec().
 			WithServiceName(lws.Name).
 			WithReplicas(*lws.Spec.Replicas).
 			WithPodManagementPolicy(appsv1.ParallelPodManagement).
 			WithTemplate(&podTemplateApplyConfiguration).
-			WithServiceName(lws.Name).
```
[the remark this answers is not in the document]
kerthcet: We have the service name already, see L456.

kerthcet:
I'll push the commits tomorrow. Only parts of the comments solved.

pkg/controllers/leaderworkerset_controller.go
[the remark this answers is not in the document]
liurupeng: ok, make sense!

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -200,6 +181,139 @@ func SetupIndexes(indexer client.FieldIndexer) error {
...
+// - if leader template is nil, and worker template updated, we'll roll out the whole lws.
+// - if leader template is nil, but we update with
+
+// Rolling update will always wait for the former replica ready then process the next one,
+// we didn't consider rollout strategy type here for we only support rollingUpdate now,
+// once we have more policies, we should update the logics here.
+// Possible scenarios:
+//   - When sts is under creation, partition is always 0 because pods are created in parallelism, rolling update is not fit here.
+//   - When sts is rolling update, the partition will start from the last index to the 0 index step by maxUnavailable.
+//   - When sts is rolling update, replicas scaled up, the partition will not change until scaling ended.
```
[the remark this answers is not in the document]
kerthcet: Actually, we'll use new template.

pkg/controllers/leaderworkerset_controller.go
kerthcet: make sense.

pkg/controllers/leaderworkerset_controller.go
kerthcet: Changed the logics here, right now, the Partition is one-transition, which means it can only become smaller in rolling update. This is to avoid the bug like when rolling update and Replicas increases, we'll update the old replicas with an old template. Because in sts, when the index is less than the partition, rolling update the replica will apply with the old template. You can refer to the code for more details, I added some comments.

pkg/controllers/leaderworkerset_controller.go
kerthcet: For functionality, seems not that necessary, but for visibility, seems useful, maybe I'll add it later, but on my original design, I didn't have the plan.

kerthcet: TL;DR: I'll add the field.

pkg/controllers/leaderworkerset_controller.go
kerthcet: Before: we'll calculate from the scratch again, but will not restart the replica.
Now: we'll frozen the Partition.

pkg/controllers/leaderworkerset_controller.go
kerthcet: Because at first I though the items will start from the last index to 0, which will accelerate the calculation, but seems not that case, but I didn't change the implementation for lazy reason.
Anyway I think it doesn't make any difference? 😅

kerthcet:
> Regarding updating the size will trigger the StatefulSet upgrade, could you elaborate more on this?

As you said, we have an annotation on the leaderPod, so each time we update the size we'll trigger the whole rolling update of the pods. My assumption is when updating the size should we trigger the rolling update, or just scaling up and down, because the template is not updated.

ahg-g:
/hold

Just so it doesn't get merged by mistake, I would like to take a look at this once the forst round of review is done

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -200,6 +181,146 @@ func SetupIndexes(indexer client.FieldIndexer) error {
...
+				}
+
+				finishedPart++
+				break
+			}
+		}
+	}
+
+	// When updated replicas become not ready again or scaled up replicas haven't ready yet,
+	// we'll not modify the Partition field. That means Partition is one-transition to make it simple.
```
ahg-g, line 283:
```suggestion
	// When updated replicas become not ready again or scaled up replicas are not ready yet,
	// we'll not modify the Partition field. That means Partition moves in one direction to make it simple.
```

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -200,6 +181,146 @@ func SetupIndexes(indexer client.FieldIndexer) error {
...
+	if err := r.List(ctx, &stsList, stsSelector, client.InNamespace(lws.Namespace)); err != nil {
+		return 0, err
+	}
+
+	var finishedPart int32
+	templateHash := utils.LeaderWorkerTemplateHash(lws)
+
+stop:
+	for i := replicas - 1; i >= 0; i-- {
+		for j := len(stsList.Items) - 1; j >= 0; j-- {
```
ahg-g, line 251: This nested loop is expensive, how about split them into two linear loops: 

loop 1 iterates over the items and places them in a list based on their indexes (remember each item has an index).
loop 2 iterate over the indexed list in reverse order to check which ones are ready.

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -80,42 +79,50 @@ var _ webhook.CustomValidator = &LeaderWorkerSetWebhook{}
...
+func (r *LeaderWorkerSetWebhook) generalValidate(ctx context.Context, obj runtime.Object) (admission.Warnings, field.ErrorList) {
 	lws := obj.(*v1.LeaderWorkerSet)
-	var allErrs []error
+	specPath := field.NewPath("spec")
+
+	var allErrs field.ErrorList
 	// Ensure replicas and groups number are valid
 	if lws.Spec.Replicas != nil && *lws.Spec.Replicas < 0 {
-		allErrs = append(allErrs, fmt.Errorf("the Replicas %d is invalid", lws.Spec.Replicas))
+		allErrs = append(allErrs, field.Invalid(specPath.Child("replicas"), lws.Spec.Replicas, "replicas must greater or equal to 0"))
```
ahg-g, line 108:
```suggestion
		allErrs = append(allErrs, field.Invalid(specPath.Child("replicas"), lws.Spec.Replicas, "replicas must be greater or equal to 0"))
```

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -80,42 +79,50 @@ var _ webhook.CustomValidator = &LeaderWorkerSetWebhook{}
...
 	}
 	if *lws.Spec.LeaderWorkerTemplate.Size < 1 {
-		allErrs = append(allErrs, fmt.Errorf("the Size %d is invalid", lws.Spec.LeaderWorkerTemplate.Size))
+		allErrs = append(allErrs, field.Invalid(specPath.Child("leaderWorkerTemplate", "size"), lws.Spec.LeaderWorkerTemplate.Size, "size must greater than 1"))
```
ahg-g, line 111:
```suggestion
		allErrs = append(allErrs, field.Invalid(specPath.Child("leaderWorkerTemplate", "size"), lws.Spec.LeaderWorkerTemplate.Size, "size must be equal or greater than 1"))
```

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -80,42 +79,50 @@ var _ webhook.CustomValidator = &LeaderWorkerSetWebhook{}
...
 	}
 	if int64(*lws.Spec.Replicas)*int64(*lws.Spec.LeaderWorkerTemplate.Size) > math.MaxInt32 {
-		allErrs = append(allErrs, fmt.Errorf("the product of replicas and worker replicas must not exceed %d for lws '%s'", math.MaxInt32, lws.Name))
+		// allErrs = append(allErrs, fmt.Errorf("the product of replicas and worker replicas must not exceed %d for lws '%s'", math.MaxInt32, lws.Name))
```
ahg-g, line 114: remove

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -80,42 +79,50 @@ var _ webhook.CustomValidator = &LeaderWorkerSetWebhook{}
...
-	var allErrs []error
-	newLws := newObj.(*v1.LeaderWorkerSet)
-	oldLws := oldObj.(*v1.LeaderWorkerSet)
-	if newLws.Spec.Replicas != nil && *newLws.Spec.Replicas < 0 {
-		allErrs = append(allErrs, fmt.Errorf("the Replicas %d is invalid", newLws.Spec.Replicas))
 	}
-	newLwsClone := newLws.DeepCopy()
-	// Replicas can be mutated
-	newLwsClone.Spec.Replicas = oldLws.Spec.Replicas
-	if !apiequality.Semantic.DeepEqual(oldLws.Spec, newLwsClone.Spec) {
```
ahg-g, line 112: Why are we removing this logic? it might be better to do the opposite, by default a field is immutable unless we explicitly make it not.

test/integration/webhooks/leaderworkerset_test.go
```diff
@@ -193,15 +193,6 @@ var _ = ginkgo.Describe("leaderworkerset defaulting, creation and update", func(
 			},
 			lwsCreationShouldFail: true,
 		}),
-		ginkgo.Entry("update with invalid replicas should fail (larger than maxInt32)", &testValidationCase{
```
ahg-g, line 196: why was this removed?

ahg-g (commented):
This is really good, thanks Kante!

Makefile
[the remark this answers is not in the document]
kerthcet: Already reverted.

pkg/utils/utils.go
```diff
@@ -0,0 +1,43 @@
...
+
+package utils
+
+import (
+	"crypto/sha1"
+	"encoding/hex"
+
+	leaderworkerset "sigs.k8s.io/lws/api/leaderworkerset/v1"
+)
+
```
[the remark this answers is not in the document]
kerthcet: Not yet, most are regular functions. We can migrate later if necessary.

test/testutils/util.go
```diff
@@ -247,3 +298,30 @@ func UpdateReplicaCount(ctx context.Context, k8sClient client.Client, lws *leade
 		return k8sClient.Update(ctx, lws)
 	}, Timeout, Interval).Should(gomega.Succeed())
 }
+
+func UpdateSize(ctx context.Context, k8sClient client.Client, lws *leaderworkerset.LeaderWorkerSet, size int32) {
+	gomega.Eventually(func() error {
+		lws.Spec.LeaderWorkerTemplate.Size = ptr.To[int32](size)
```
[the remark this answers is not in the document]
kerthcet: Yes, I used to enable update the size as well, but some ideas still need to be discussed.

I didn't clean up all the codes.

test/integration/controllers/leaderworkerset_test.go
```diff
@@ -538,8 +518,389 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
...
+
+						var leaderSts appsv1.StatefulSet
+						gomega.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: leaderworkerset.Name, Namespace: leaderworkerset.Namespace}, &leaderSts)).To(gomega.Succeed())
+						// Manually create leader pods here because we have no statefulset controller.
+						gomega.Expect(testing.CreateLeaderPods(ctx, leaderSts, k8sClient, lws, 4, 6)).To(gomega.Succeed())
+					},
+					checkLWSState: func(lws *leaderworkerset.LeaderWorkerSet) {
+						testing.ExpectValidLeaderStatefulSet(ctx, lws, k8sClient)
+						testing.ExpectLeaderWorkerSetUnavailable(ctx, k8sClient, lws, "All replicas are ready")
+						testing.ExpectStatefulsetPartitionEqualTo(ctx, k8sClient, lws, 4)
```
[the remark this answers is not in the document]
kerthcet: Yes, when scaling up, Partition will not change, new created replicas will be created with new remplates. updated.

pkg/controllers/leaderworkerset_controller.go
kerthcet: Let me follow this next for performance issue.

ahg-g: Thanks, the approach of splitting into two loops is more intuitive too and easier to follow.

ahg-g:
> so I remove the rolling update for size update right now, and make it immutable as original.

I think an immutable size is fine and a reasonable tradeoff to simplify the rolling update logic

pkg/controllers/leaderworkerset_controller.go
liurupeng: during upgrade, if the upgraded replicas is consistently in not ready state. This means the upgrade will frozen forever. In this case, would we suggest customer to re-apply the working template themselves, or we provide a mechanism to rollback to the previous one.

pkg/webhooks/leaderworkerset_webhook.go
ahg-g: Is RolloutStrategy mutable in statefulsets?

pkg/controllers/leaderworkerset_controller.go
ahg-g: > Let me follow this next for performance issue.

In this PR, right?

pkg/webhooks/leaderworkerset_webhook.go
kerthcet: yes, we apply the same strategy.

pkg/controllers/leaderworkerset_controller.go
kerthcet: yes, I'm finishing the last part.

kerthcet:
Last 2 commit:
- Add on e2e test
- Solve the performance issue when calculating the partition field

pkg/controllers/leaderworkerset_controller.go
kerthcet: That's a good question, I think currently we're doing the same way as sts right now, sts will stuck in this state forever, I used to ticket a proposal about `EnforceRollingUpdate` in the upstream, but didn't continue https://github.com/kubernetes/enhancements/issues/3541, it's a similar question.

Now user can only reapply a fixed template, but maybe later we should support rollback in lws as well to mitigate this pain point.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -236,7 +236,21 @@ func (r *LeaderWorkerSetReconciler) rollingUpdatePartition(ctx context.Context,
...
+	podSelector := client.MatchingLabels(map[string]string{
+		leaderworkerset.SetNameLabelKey:     lws.Name,
+		leaderworkerset.WorkerIndexLabelKey: "0",
+	})
+	var leaderPodList corev1.PodList
+	if err := r.List(ctx, &leaderPodList, podSelector, client.InNamespace(lws.Namespace)); err != nil {
+		return 0, err
+	}
+	// Got a sorted leader pod list matches with the following sorted statefulsets one by one, which means
+	// the leader pod and the corresponding work statefulset has the same index.
```
ahg-g, line 250:
```suggestion
	// the leader pod and the corresponding worker statefulset has the same index.
```

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -246,40 +260,32 @@ func (r *LeaderWorkerSetReconciler) rollingUpdatePartition(ctx context.Context,
...
-				if !(stsTemplateHash == templateHash && statefulsetutils.StatefulsetReady(stsList.Items[j])) {
-					break stop
-				}
-
-				leaderPod := &corev1.Pod{}
-				if err := r.Get(ctx, types.NamespacedName{Name: nominatedName, Namespace: lws.Namespace}, leaderPod); err != nil {
-					return 0, err
-				}
+		nominatedName := fmt.Sprintf("%s-%d", lws.Name, i)
+		if nominatedName != sortedPods[i].Name || nominatedName != sortedSts[i].Name {
```
ahg-g, line 274: maybe add a comment here (or move the one above) saying this indicates that the leader or the worker statefulset isn't created yet.

test/e2e/e2e_test.go
```diff
@@ -53,6 +53,21 @@ var _ = Describe("leaderWorkerSet e2e tests", func() {
...
+	It("Can rolling update lws", func() {
+		lws := testing.BuildLeaderWorkerSet(ns.Name).Obj()
+		Expect(k8sClient.Create(ctx, lws)).To(Succeed())
+
+		// Wait for leaderWorkerSet ready then update it.
+		testing.ExpectLeaderWorkerSetAvailable(ctx, k8sClient, lws, "All replicas are ready")
+		testing.UpdateWorkerTemplate(ctx, k8sClient, lws)
+
+		// Wait for leaderWorkerSet ready again.
+		testing.ExpectLeaderWorkerSetAvailable(ctx, k8sClient, lws, "All replicas are ready")
```
ahg-g, line 68: how do we ensure that this is checking the availability status after the update? isn't it possible that we are still checking the previous status before the lws got a chance to pick the update and change the availability status to false?

test/integration/controllers/leaderworkerset_test.go
```diff
@@ -791,6 +791,79 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
...
+						gomega.Eventually(func() error {
+							if err := k8sClient.Get(ctx, types.NamespacedName{Name: lws.Name, Namespace: lws.Namespace}, &leaderworkerset); err != nil {
+								return err
+							}
+							leaderworkerset.Spec.Replicas = ptr.To[int32](3)
+							return k8sClient.Update(ctx, &leaderworkerset)
+						}, testing.Timeout, testing.Interval).Should(gomega.Succeed())
+
+						var leaderSts appsv1.StatefulSet
+						gomega.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: leaderworkerset.Name, Namespace: leaderworkerset.Namespace}, &leaderSts)).To(gomega.Succeed())
```
ahg-g, line 843: what are we using this for?

ahg-g (commented):
A couple of nits, otherwise this is good from my end!

test/integration/controllers/leaderworkerset_test.go
kerthcet: removed

test/e2e/e2e_test.go
kerthcet: Move this down then we'll check the hash first to make sure we're validating the new revision.

kerthcet:
All set @ahg-g

liurupeng:
lgtm as well, I will wait for @ahg-g to approve

ahg-g:
/lgtm
/hold cancel

Great work!

test/e2e/e2e_test.go
ahg-g: I am not sure this will make a difference, it would be nice if we can check that the template hash changed, and then check that the lws is available.

kerthcet: I think we already do as you said, so the process is:
- update the lws
- `ExpectValidLeaderStatefulSet` will check that the template hash is already the updated one, see https://github.com/kubernetes-sigs/lws/blob/59f5a761c23609bbc28b9aa0493d6c2b0722a608/test/testutils/validators.go#L121-L124
- `ExpectLeaderWorkerSetAvailable` validate the lws is available.

PR #76, opened 2024-03-29
E2E test now waits for webhook to complete setup

Bslabe123:
#### What type of PR is this?
/kind bug

```release-note
NONE
```

### DIFF
--- a/test/e2e/suite_test.go
+++ b/test/e2e/suite_test.go
@@ -91,17 +91,17 @@ func LwsReadyForTesting(client client.Client) {
 	By("waiting for webhooks to come up")
 
 	// To verify that webhooks are ready, let's create a simple lws.
-	leaderWorkerSetSpec := testutils.BuildLeaderWorkerSet(metav1.NamespaceDefault).Replica(3).Obj()
+	lws := testutils.BuildLeaderWorkerSet(metav1.NamespaceDefault).Replica(3).Obj()
 
 	// Once the creation succeeds, that means the webhooks are ready
 	// and we can begin testing.
 	Eventually(func() error {
-		return client.Create(ctx, leaderWorkerSetSpec)
+		return client.Create(ctx, lws)
 	}, timeout, interval).Should(Succeed())
 
 	// Delete this leaderworkerset before beginning tests.
-	Expect(client.Delete(ctx, leaderWorkerSetSpec))
+	Expect(client.Delete(ctx, lws))
 	Eventually(func() error {
-		return client.Get(ctx, types.NamespacedName{Name: leaderWorkerSetSpec.Name, Namespace: leaderWorkerSetSpec.Namespace}, &leaderworkerset.LeaderWorkerSet{})
+		return client.Get(ctx, types.NamespacedName{Name: lws.Name, Namespace: lws.Namespace}, &leaderworkerset.LeaderWorkerSet{})
 	}).ShouldNot(Succeed())
 }
--- a/test/e2e/suite_test.go
+++ b/test/e2e/suite_test.go
@@ -101,8 +101,7 @@ func LwsReadyForTesting(client client.Client) {
 
 	// Delete this leaderworkerset before beginning tests.
 	Expect(client.Delete(ctx, leaderWorkerSetSpec))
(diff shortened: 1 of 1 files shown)


test/e2e/suite_test.go
```diff
@@ -93,11 +79,30 @@ var _ = BeforeSuite(func() {
...
 
 var _ = AfterSuite(func() {
 	cancel()
-	By("tearing down the test environment")
-	err := testEnv.Stop()
-	Expect(err).NotTo(HaveOccurred())
 })
+
+func LwsReadyForTesting(client client.Client) {
+	By("waiting for ")
```
liurupeng, line 91:
```suggestion
	By("waiting for webhooks to come up")
```

test/e2e/suite_test.go
```diff
@@ -93,11 +79,30 @@ var _ = BeforeSuite(func() {
...
+
+	// To verify that webhooks are ready, let's create a simple lws.
+	lws := testutils.BuildLeaderWorkerSet(metav1.NamespaceDefault).Replica(3).Obj()
+
+	// Once the creation succeeds, that means the webhooks are ready
+	// and we can begin testing.
+	Eventually(func() error {
+		return client.Create(ctx, lws)
+	}, timeout, interval).Should(Succeed())
+
```
liurupeng, line 101: could you verify that the some of the lws's spec is as expected?

test/e2e/suite_test.go
```diff
@@ -93,11 +79,30 @@ var _ = BeforeSuite(func() {
...
+	// Delete this leaderworkerset before beginning tests.
+	Expect(client.Delete(ctx, lws))
```
liurupeng, line 103: why do we delete the lws here before get?

test/e2e/suite_test.go
Bslabe123: We could but it would duplicate the functionality of the actual e2e test, creating this lws is a consequence of needing to check that the webhooks are ready

test/e2e/suite_test.go
Bslabe123: This lws serves no other purpose other than to ensure the webhooks are ready, once they are the lws is no longer needed here

test/e2e/suite_test.go
```diff
@@ -93,11 +79,30 @@ var _ = BeforeSuite(func() {
...
+	var leaderWorkerSetStruct leaderworkerset.LeaderWorkerSet
```
liurupeng, line 104: let's rename it to leaderworkerset

test/e2e/suite_test.go
liurupeng: sg, then we need to add several test cases after setup the suite

test/e2e/suite_test.go
Bslabe123: renamed it to leaderworkersetspec to avoid confusion with imported leaderworkerset

liurupeng: then reuse "lws" instead of having a new variable?

test/e2e/suite_test.go
```diff
@@ -93,11 +79,29 @@ var _ = BeforeSuite(func() {
...
-	By("tearing down the test environment")
-	err := testEnv.Stop()
-	Expect(err).NotTo(HaveOccurred())
 })
+
+func LwsReadyForTesting(client client.Client) {
+	By("waiting for webhooks to come up")
+
+	// To verify that webhooks are ready, let's create a simple lws.
+	leaderWorkerSetSpec := testutils.BuildLeaderWorkerSet(metav1.NamespaceDefault).Replica(3).Obj()
```
liurupeng, line 94: we use spec usually to define the context of a template, like podSpec for the spec of the pod template, which is a sub field of the pod template, here, I think lws may make more sense, wdyt?

ahg-g:
how did this merge and the e2e tests didn't pass?

liurupeng:
I merged it, thought that the e2e test will not run with the latest changes, my bad.

liurupeng:
@Bslabe123 is working on fix it right now

PR #82, opened 2024-04-02
Cleanup after e2e test

kerthcet:
#### What type of PR is this?
/kind cleanup

Benefit to using existing kind cluster.

Fixes #

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ ARTIFACTS ?= $(PROJECT_DIR)/bin
 
 INTEGRATION_TARGET ?= ./test/integration/...
 
-E2E_KIND_VERSION ?= kindest/node:v1.25.1
+E2E_KIND_VERSION ?= kindest/node:v1.27.1
 USE_EXISTING_CLUSTER ?= false
 
 # For local testing, we should allow user to use different kind cluster name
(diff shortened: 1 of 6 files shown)


test/testutils/wrappers.go
```diff
@@ -168,7 +168,7 @@ func MakeLeaderPodSpec() corev1.PodSpec {
 		Containers: []corev1.Container{
 			{
 				Name:  "worker",
-				Image: "busybox",
+				Image: "nginx:1.14.2",
```
kerthcet, line 171: busybox will lead to the pod completed, which will make the statefulset not ready.

Makefile
```diff
@@ -50,7 +50,7 @@ ARTIFACTS ?= $(PROJECT_DIR)/bin
 
 INTEGRATION_TARGET ?= ./test/integration/...
 
-E2E_KIND_VERSION ?= kindest/node:v1.25.1
+E2E_KIND_VERSION ?= kindest/node:v1.26.0
```
kerthcet, line 53: This is required since only version greater than 1.26 then sts support  `Ordinals`.

test/testutils/validators.go
```diff
@@ -36,7 +36,7 @@ import (
 )
 
 const (
-	Timeout  = 3 * time.Second
+	Timeout  = 1 * time.Minute
```
kerthcet, line 39: 3s not enough for e2e test.

PR #91, opened 2024-04-10
cleanup: use kubebuilder marker instead of webhook

### LINKED ISSUE #15: Use kubebuilder validation & default for simplification
No longer to add specific logics to webhook, See https://book.kubebuilder.io/reference/markers/crd-validation
/kind feature

nayihz:
#### What type of PR is this?
/kind cleanup

Fixes #15 

```release-note
NONE
```

### DIFF
--- a/api/leaderworkerset/v1/leaderworkerset_types.go
+++ b/api/leaderworkerset/v1/leaderworkerset_types.go
@@ -110,7 +110,9 @@ type LeaderWorkerTemplate struct {
 	Size *int32 `json:"size,omitempty"`
 
 	// RestartPolicy defines the restart policy when pod failures happen.
-	RestartPolicy RestartPolicyType `json:"restartPolicy,omitempty"`
+	// +kubebuilder:default=Default
+	// +kubebuilder:validation:Enum={Default,RecreateGroupOnPodRestart}
+	RestartPolicy RestartPolicyType `json:"restartPolicy"`
 }
 
 // RolloutStrategy defines the strategy that the leaderWorkerSet controller
--- a/pkg/webhooks/leaderworkerset_webhook.go
+++ b/pkg/webhooks/leaderworkerset_webhook.go
@@ -50,7 +50,6 @@ var _ webhook.CustomDefaulter = &LeaderWorkerSetWebhook{}
 // Default implements webhook.Defaulter so a webhook will be registered for the type
 func (r *LeaderWorkerSetWebhook) Default(ctx context.Context, obj runtime.Object) error {
 	lws := obj.(*v1.LeaderWorkerSet)
-
 	if lws.Spec.LeaderWorkerTemplate.RestartPolicy == "" {
 		lws.Spec.LeaderWorkerTemplate.RestartPolicy = v1.DefaultRestartPolicy
 	}
@@ -74,13 +73,13 @@ var _ webhook.CustomValidator = &LeaderWorkerSetWebhook{}
 
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (r *LeaderWorkerSetWebhook) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
-	warnings, allErrs := r.generalValidate(ctx, obj)
+	warnings, allErrs := r.generalValidate(obj)
 	return warnings, allErrs.ToAggregate()
 }
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (r *LeaderWorkerSetWebhook) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) (admission.Warnings, error) {
-	warnings, allErrs := r.generalValidate(ctx, newObj)
+	warnings, allErrs := r.generalValidate(newObj)
 
 	oldLws := oldObj.(*v1.LeaderWorkerSet)
 	newLws := newObj.(*v1.LeaderWorkerSet)
@@ -93,7 +92,7 @@ func (r *LeaderWorkerSetWebhook) ValidateDelete(ctx context.Context, obj runtime
 	return nil, nil
 }
 
-func (r *LeaderWorkerSetWebhook) generalValidate(ctx context.Context, obj runtime.Object) (admission.Warnings, field.ErrorList) {
+func (r *LeaderWorkerSetWebhook) generalValidate(obj runtime.Object) (admission.Warnings, field.ErrorList) {
 	lws := obj.(*v1.LeaderWorkerSet)
 	specPath := field.NewPath("spec")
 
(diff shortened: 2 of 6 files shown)


pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -51,14 +51,6 @@ var _ webhook.CustomDefaulter = &LeaderWorkerSetWebhook{}
 func (r *LeaderWorkerSetWebhook) Default(ctx context.Context, obj runtime.Object) error {
 	lws := obj.(*v1.LeaderWorkerSet)
 
-	if lws.Spec.LeaderWorkerTemplate.RestartPolicy == "" {
```
kerthcet, line 54: Default only works for optional field, plz revert this change.

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -110,6 +110,8 @@ type LeaderWorkerTemplate struct {
 	Size *int32 `json:"size,omitempty"`
 
 	// RestartPolicy defines the restart policy when pod failures happen.
+	// +kubebuilder:default=Default
+	// +kubebuilder:validation:Enum=Default;RecreateGroupOnPodRestart
```
kerthcet, line 114:
```suggestion
	// +kubebuilder:validation:Enum={Default,RecreateGroupOnPodRestart}
```

kerthcet (commented):
Also we forgot to add a validation for maxUnavailable, which can be less t than 1. Can you help to fix that? Don't forget to add a test.

Thanks.

pkg/webhooks/leaderworkerset_webhook.go
nayihz: `// +kubebuilder:default=Default` also works for `RestartPolicy`

nayihz:
> Also we forgot to add a validation for maxUnavailable, which can be less t than 1.

`maxUnavailable` could be int or string. IIRC we have no marker to validate it. Maybe we can only validate this field in the webhook?

test/integration/webhooks/leaderworkerset_test.go
```diff
@@ -111,10 +111,15 @@ var _ = ginkgo.Describe("leaderworkerset defaulting, creation and update", func(
 		}),
 		ginkgo.Entry("apply default rollout strategy", &testDefaultingCase{
 			makeLeaderWorkerSet: func(ns *corev1.Namespace) *testutils.LeaderWorkerSetWrapper {
-				return testutils.BuildLeaderWorkerSet(ns.Name).RolloutStrategy(leaderworkerset.RolloutStrategy{}) // unset rollout strategy
```
kerthcet, line 114: Why change this? BuilderLeaderWorkerSet also set the rolloutStrategy, so what you changed here will not validate the defaulter for rollout strategy.

pkg/webhooks/leaderworkerset_webhook.go
kerthcet: You can, but it doesn't work. Default works after admission of the request, so `Required` will check first.

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -108,13 +100,10 @@ func (r *LeaderWorkerSetWebhook) generalValidate(ctx context.Context, obj runtim
...
 	}
-	if lws.Spec.LeaderWorkerTemplate.RestartPolicy != v1.DefaultRestartPolicy && lws.Spec.LeaderWorkerTemplate.RestartPolicy != v1.RecreateGroupOnPodRestart {
-		allErrs = append(allErrs, field.Invalid(specPath.Child("leaderWorkerTemplate", "restartPolicy"), lws.Spec.LeaderWorkerTemplate.RestartPolicy, "only support [Default, RecreateGroupOnPodRestart]"))
 
-	}
 	if lws.Spec.RolloutStrategy.RollingUpdateConfiguration != nil {
 		value, err := intstr.GetScaledValueFromIntOrPercent(&lws.Spec.RolloutStrategy.RollingUpdateConfiguration.MaxUnavailable, int(*lws.Spec.Replicas), false)
-		if err != nil || value > int(*lws.Spec.Replicas) {
+		if err != nil || value > int(*lws.Spec.Replicas) || value < 1 {
 			allErrs = append(allErrs, field.Invalid(specPath.Child("rolloutStrategy", "rollingUpdateConfiguration", "maxUnavailable"), lws.Spec.RolloutStrategy.RollingUpdateConfiguration.MaxUnavailable, "maxUnavailable must not greater than replicas"))
```
kerthcet, line 107:
```suggestion
allErrs = append(allErrs, field.Invalid(specPath.Child("rolloutStrategy", "rollingUpdateConfiguration", "maxUnavailable"), lws.Spec.RolloutStrategy.RollingUpdateConfiguration.MaxUnavailable, "maxUnavailable can neither greater than replicas nor less than 1"))
```

pkg/webhooks/pod_webhook_test.go
```diff
@@ -23,7 +23,6 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
kerthcet, line 26: good catch

pkg/webhooks/leaderworkerset_webhook.go
kerthcet: You successful here is because you set the `emitempty` to `Type` then this field is a **Optional** field now, but actually we want to make it a **Required** field. 

The problem with your approach here is the `RollingUpdateConfiguration` might be nil and panic when rolling update because `RollingUpdateConfiguration` can't be nil, but you didn't validate this below when spec.rolloutStrategy.type is empty.

My suggestion is:
- Remove the `omitempty` from `RestartPolicy` and `Type` to make them required fields
- Revert the codes here

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -51,14 +51,6 @@ var _ webhook.CustomDefaulter = &LeaderWorkerSetWebhook{}
...
 
-	if lws.Spec.LeaderWorkerTemplate.RestartPolicy == "" {
-		lws.Spec.LeaderWorkerTemplate.RestartPolicy = v1.DefaultRestartPolicy
-	}
-
-	if lws.Spec.RolloutStrategy.Type == "" {
-		lws.Spec.RolloutStrategy.Type = v1.RollingUpdateStrategyType
-	}
-
 	if lws.Spec.RolloutStrategy.Type == v1.RollingUpdateStrategyType && lws.Spec.RolloutStrategy.RollingUpdateConfiguration == nil {
```
kerthcet, line 54: I mean here lws.Spec.RolloutStrategy.Type might be empty if we didn't set the value.

pkg/webhooks/leaderworkerset_webhook.go
kerthcet: See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

nayihz: Thanks for detailed explanation.

kerthcet (commented):
/lgtm
/approve

Thanks!

pkg/webhooks/leaderworkerset_webhook.go
nayihz: I'm curious that why users can set `RestartPolicy == ""` if `RestartPolicy` field is `required`

nayihz: > RestartPolicy is required by the Schema because it can't be empty.

Since RestartPolicy cannot be configured as empty, why is it still possible to be set to`""`(`RestartPolicy == ""`)?

nayihz: IIUC, RestartPolicy field is `required` and could be only set to `Default` or `RecreateGroupOnPodRestart`. So we should not allow user to set `RestartPolicy = ""`.

PR #97, opened 2024-04-11
Adding system metrics support

### LINKED ISSUE #87: System metrics support

We should add some metrics for tracking the health of the system, some of them like:
- Rolling update duration
- Recreate group times
- HPA related? Like scaling up/down times & duration?

Please add more if you think so.

**Why is this needed**:

Providing insight of the system

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [x] Docs update

Edwinhr716:
#### What type of PR is this?
/kind feature

Fixes #87 

```release-note
adding recreate_group_times, rolling_update_duration metrics, and replica_ready_status_duration
```

### DIFF
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -41,6 +42,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	leaderworkerset "sigs.k8s.io/lws/api/leaderworkerset/v1"
+	"sigs.k8s.io/lws/pkg/metrics"
 	"sigs.k8s.io/lws/pkg/utils"
 	podutils "sigs.k8s.io/lws/pkg/utils/pod"
 	statefulsetutils "sigs.k8s.io/lws/pkg/utils/statefulset"
@@ -367,7 +369,8 @@ func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *l
 			}
 			if podutils.PodRunningAndReady(leaderPod) {
 				readyCount++
-
+				waitTime := getLastTransitionTime(string(leaderworkerset.LeaderWorkerSetProgressing), lws)
+				metrics.ReplicaReadyStatus(sts.Name, time.Since(waitTime.Time))
 				if sts.Labels[leaderworkerset.TemplateRevisionHashKey] == templateHash && leaderPod.Labels[leaderworkerset.TemplateRevisionHashKey] == templateHash {
 					updatedCount++
 				}
@@ -568,4 +571,13 @@ func templateUpdated(sts *appsv1.StatefulSet, lws *leaderworkerset.LeaderWorkerS
 
 func replicasUpdated(sts *appsv1.StatefulSet, lws *leaderworkerset.LeaderWorkerSet) bool {
 	return *sts.Spec.Replicas != *lws.Spec.Replicas
+}
+
+func getLastTransitionTime(conditionType string, lws *leaderworkerset.LeaderWorkerSet) metav1.Time {
+	for _, condition := range lws.Status.Conditions {
+		if condition.Type == conditionType {
+			return condition.LastTransitionTime
+		}
+	}
+	return metav1.Now()
 }
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"strconv"
-	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -350,7 +349,7 @@ func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *l
 	readyCount := 0
 	updatedCount := 0
 	templateHash := utils.LeaderWorkerTemplateHash(lws)
-
+	var readyLeaderPods []corev1.Pod
 	// Iterate through all statefulsets.
 	for _, sts := range lwssts.Items {
 		if sts.Name == lws.Name {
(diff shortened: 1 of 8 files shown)


ahg-g:
is this ready for review? asking because it is still marked as WIP

Edwinhr716:
not yet, currently working on adding a metric for latency for pods to be scheduled and become ready, and also plan on adding unit tests

Edwinhr716:
Done with most of it, PTAL. Only missing unit test for `ReadyReplicaStatus` but want to make sure the implementation looks good first

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+
+	recreateGroupTimes = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: "lws",
+			Name:      "recreate_group_times",
+			Help:      "number of times a group has been recreated",
+		}, []string{"leadername"},
+	)
+
+	replicaReadyStatusDuration = prometheus.NewHistogramVec(
```
liurupeng, line 47: add a comment here about what is the metrics record, like "the latency for a pod group to be ready after creation"

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+	rollingUpdateDuration.WithLabelValues(hash).Observe(duration.Seconds())
+}
+
+func RecreatingGroup(leaderName string) {
+	recreateGroupTimes.WithLabelValues(leaderName).Inc()
+}
+
+func ReplicaReadyStatus(readyPods []corev1.Pod, startTime metav1.Time) {
+	for _, pod := range readyPods {
+		readyTime := podutils.PodReadyConditionLastTransitionTime(pod).Time
```
liurupeng, line 66: when the leader pod is ready, its worker statefulset may not be ready right? the latency here calculates the leader pods' latency to be ready

liurupeng: we may need to pass the worker sts and leader pod as a pair and make sure that latency=worker_sts_ready-leader_pod_created, wdyt

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -569,3 +573,12 @@ func templateUpdated(sts *appsv1.StatefulSet, lws *leaderworkerset.LeaderWorkerS
 func replicasUpdated(sts *appsv1.StatefulSet, lws *leaderworkerset.LeaderWorkerSet) bool {
 	return *sts.Spec.Replicas != *lws.Spec.Replicas
 }
+
+func getLastTransitionTime(conditionType string, lws *leaderworkerset.LeaderWorkerSet) metav1.Time {
```
liurupeng, line 577: maybe move this to a util file?

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -389,6 +390,9 @@ func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *l
 	updateCondition := setCondition(lws, condition)
 	// if condition changed, record events
 	if updateCondition {
+		if updatedCount == int(*lws.Spec.Replicas) {
+			metrics.ReplicaReadyStatus(readyLeaderPods, getLastTransitionTime(string(leaderworkerset.LeaderWorkerSetProgressing), lws))
```
liurupeng, line 394: the start time here is when the lws just progress, there might be a possibility that the leader pod is not scheduled.

pkg/metrics/metrics.go
liurupeng: but overall, I start to think about whether it's necessary to have this metrics for now.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+var (
+	rollingUpdateDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: "lws",
+			Name:      "rolling_update_duration",
+			Help:      "Duration of rolling updates",
+		}, []string{"hash"},
+	)
+
+	recreateGroupTimes = prometheus.NewCounterVec(
```
liurupeng, line 39: this is not called anywhere else right?

Edwinhr716: Yeah, adding it to register it, but waiting on #106 to implement it.

pkg/metrics/metrics.go
Edwinhr716: The LeaderPod is only added into the list that is passed if the worker sts is ready, that is checked [here](https://github.com/kubernetes-sigs/lws/pull/97/files#diff-636763c4aab72861dbe05856e71e11abf551830d79fdcd848f3133ffc9c2d6e1L359) 

I guess I'm a bit confused with what the replica latency measures,  I thought it was from when the lws object was created/edited to when the replica is ready. So is it from when the leader statefulset is created to when the worker sts is ready?

pkg/controllers/leaderworkerset_controller.go
Edwinhr716: Asked a similar question in other comment: what is the replica latency actually measuring? I thought it was from lws creation to when the replica is ready

pkg/metrics/metrics.go
Edwinhr716: the help attribute of the object describes what it records, but I can add a comment for it as well

pkg/controllers/leaderworkerset_controller.go
Edwinhr716: Was thinking about that as well, can create an lws utils file similar to the one that exists for pods, and move this function, `templateUpdated` and `replicasUpdated` to it in a different PR.

pkg/metrics/metrics.go
ahg-g: I don't think we need such a metric. The time it takes for the pod to become ready is not a characteristic of the lws controller.

ahg-g (commented):
I don't know if we really need to add our own metrics at this stage, I would focus on making sure that we are getting the metrics that controller-runtime publishes by default; see https://book.kubebuilder.io/reference/metrics-reference

Those metrics cover a lot of what we need already for monitoring the health of the operator, like how long a reconcile call takes, and I think the first PR should focus on that only and it should be included in the 0.2.0 release.

kerthcet:
+1 to add these metrics later since not part of the 0.2.0.

liurupeng:
/hold
As discussed with Edwin, hold the PR for now, thanks @Edwinhr716

liurupeng:
close for now, @Edwinhr716

PR #98, opened 2024-04-12
Support maxSurge in rolling update

kerthcet:
#### What type of PR is this?
/kind feature

Fixes https://github.com/kubernetes-sigs/lws/issues/3

It works like:
- When rolling update, lws will burst out maxSurge replicas 
- When rolling update enter into the final stage, we'll update a new replica and reclaim a bursted replica
- Finally the rolling update succeeded.

```release-note
Support maxSurge in rolling update.
```


kerthcet:
This is the general prototype, I need more tests.

kerthcet:
So when rolling update, lws will request for hardware as well, there's no difference with `lws without maxSurge`, this can not help with your case, right? @liurupeng But when reclaiming the bursted replicas, it can somehow help here ...

kerthcet:
You can take a look if you want, only webhook validation and e2e tests not added. But seems @ahg-g has question about this feature. We can't achieve the similar behavior as deployment does.

kerthcet:
All set now. PTAL.

kerthcet:
Would you like to take a look of this PR? Because it changes a lot and once other PR merges, lots of conflicts have to be solve. @ahg-g @liurupeng

ahg-g:
I will try to find sometime this Friday

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -206,77 +206,111 @@ func SetupIndexes(indexer client.FieldIndexer) error {
...
+	replicasUpdated := stsReplicas != int(*lws.Spec.Replicas)
+
 	// Indicates a new rolling update here.
 	if templateUpdated(sts, lws) {
-		// Processing scaling up/down first prior to rolling update.
-		if replicasUpdated(sts, lws) {
-			return min(*lws.Spec.Replicas, *sts.Spec.Replicas), nil
+		if replicasUpdated {
+			// Processing scaling up/down first prior to rolling update.
+			return min(lwsReplicas, int32(stsReplicas)), burstReplicas, nil
```
liurupeng, line 261: When MaxSurge is specified, and *lws.spec.replica changed, we scale lws to *lws.spec.replicas + maxSurge right? The partition here is the number of replicas needs to be updated considering scale up/down. And each step we upgrade MaxUnavailable number of replicas.

liurupeng: this is actually slightly different from what I imaged. We have the SurgeUpgrade on GKE which we upgrade MaxSurge+MaxUnavailable number of nodes each time. so rollingStep = MaxSurge+MaxUnavailable. @ahg-g do you think we should do sth similar here as well? Otherwise it feels to me this feature more looks like enabling burst replicas during upgrade

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -379,6 +412,14 @@ func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *l
 
 				if sts.Labels[leaderworkerset.TemplateRevisionHashKey] == templateHash && leaderPod.Labels[leaderworkerset.TemplateRevisionHashKey] == templateHash {
 					updatedCount++
+
+					index, err := strconv.Atoi(sts.Labels[leaderworkerset.GroupIndexLabelKey])
+					if err != nil {
+						return false, err
+					}
+					if index < int(*lws.Spec.Replicas) {
+						currentCount++
```
liurupeng, line 421: the currentCount here is the ready&updated replicas that will not be scaled down after upgrade right. If with maxSurge=2 and lws.spec.replicas=12, then once all 12 replicas is upgraded, the currentCount here is 10 and the condition will change to available based on line 437. But actually we need to perform a scale down from 12 to 10. Is there such a possibility?

test/e2e/e2e_test.go
```diff
@@ -99,7 +99,26 @@ var _ = ginkgo.Describe("leaderWorkerSet e2e tests", func() {
...
 
-		testing.ExpectValidLeaderStatefulSet(ctx, lws, k8sClient)
+		testing.ExpectValidLeaderStatefulSet(ctx, k8sClient, lws, 2)
+		testing.ExpectValidWorkerStatefulSets(ctx, lws, k8sClient, true)
+		testing.ExpectValidPods(ctx, k8sClient, lws, &corev1.PodList{})
+		// Wait for leaderWorkerSet to be ready again.
+		testing.ExpectLeaderWorkerSetAvailable(ctx, k8sClient, lws, "All replicas are ready")
+	})
+
+	ginkgo.It("Can perform a rolling update with maxSurge set", func() {
```
liurupeng, line 109: could we have one test to cover scale up/down while doing rolling update with maxSurge and another one to perform two rolling update at the same time

test/integration/controllers/leaderworkerset_test.go
```diff
@@ -896,18 +890,283 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
...
+						testing.ExpectValidLeaderStatefulSet(ctx, k8sClient, lws, 3)
+						testing.ExpectValidWorkerStatefulSets(ctx, lws, k8sClient, true)
 						testing.ExpectLeaderWorkerSetAvailable(ctx, k8sClient, lws, "All replicas are ready")
+						testing.ExpectStatefulsetPartitionEqualTo(ctx, k8sClient, lws, 0)
 						testing.ExpectLeaderWorkerSetStatusReplicas(ctx, k8sClient, lws, 3, 3)
 					},
 				},
 			},
 		}),
+		ginkgo.Entry("rolling update with maxSurge greater than replicas", &testCase{
```
liurupeng, line 1106: probably add one test case to check when two upgrades are started at the same time

liurupeng:
Sorry for the delay, added two comments about the semantics of the MaxSurge support, once that's finalized, will check more detailed, thanks Kante!

pkg/controllers/leaderworkerset_controller.go
kerthcet: Make sense to me, also aligned with the comment: `followed by scaling up the new replicas, ensuring
	// that at least 70% of original number of replicas are available at all times
	// during the update.`

test/integration/controllers/leaderworkerset_test.go
kerthcet: Added at integration-tests.Easier to validate.

test/e2e/e2e_test.go
kerthcet: Added at integration test.

kerthcet:
How rolling update looks like right now with maxSurge, let's say we have a lws with replicas=4, maxUnavailable=2, maxSurge=2, rollinStep = 2+2 = 4:

- ✅ replica updated successfully
- ❎ replica hasn't updated yet
- ⏳ rolling update in progress

|      | Partition | Replicas | R-0 |  R-1 | R-2 | R-3 | R-4 | R-5 | Note | 
| ----------- | ----------- | ----------- | ----------- | ----------- | ----------- | ----------- | ----------- | ----------- | ----------- |
| Stage1      | 0 | 4 |  ✅   |  ✅ | ✅ | ✅ |  |  | Before rolling update |
| Stage2   | 4 | 6 |  ❎ | ❎ | ❎ | ❎ | ⏳ | ⏳ | Rolling update started, bursted 2 replicas, Partition=lws.replicas | 
| Stage3      | 2 | 6 |  ❎  |  ❎ | ⏳ | ⏳ | ⏳ | ⏳ | Partition changes to 2 (6-4) | 
| Stage4      | 2 | 6 |  ❎  |  ❎ | ⏳ | ⏳ | ✅ | ⏳ | Since the last replica is not ready, Partition will not change | 
| Stage5   | 0 | 6 |  ⏳ | ⏳ | ⏳ | ⏳ | ✅ | ✅ | Partition = 6-4-2=0 |
| Stage6      | 0 | 6 |  ⏳  |  ⏳ | ⏳ | ✅ | ✅ | ✅ |  | 
| Stage7   | 0 | 5 |  ⏳ | ✅ | ⏳ | ✅ | ✅ | | Release a replica when unavailableReplicas == maxSurge |
| Stage8     | 0 | 4 |  ✅  | ⏳ |  ✅ | ✅ | | | Release another replica for accommodation | 
| Stage9     | 0 | 4 |  ✅  | ✅ |  ✅ | ✅ | | | Rolling update finished |

kerthcet:
Have no idea why https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_lws/98/pull-lws-test-integration-main/1785199425849659392 is failed, maybe the controller is stuck.

ahg-g:
This is 

> How rolling update looks like right now with maxSurge, let's say we have a lws with replicas=4, maxUnavailable=2, maxSurge=2, rollinStep = 2+2 = 4:
> 
> * ✅ replica updated successfully
> * ❎ replica hasn't updated yet
> * ⏳ rolling update in progress
> 
> Partition	Replicas	R-0	R-1	R-2	R-3	R-4	R-5	Note
> Stage1	0	4	✅	✅	✅	✅			Before rolling update
> Stage2	4	6	❎	❎	❎	❎	⏳	⏳	Rolling update started, bursted 2 replicas, Partition=lws.replicas
> Stage3	2	6	❎	❎	⏳	⏳	⏳	⏳	Partition changes to 2 (6-4)
> Stage4	2	6	❎	❎	⏳	⏳	✅	⏳	Since the last replica is not ready, Partition will not change
> Stage5	0	6	⏳	⏳	⏳	⏳	✅	✅	Partition = 6-4-2=0
> Stage6	0	6	⏳	⏳	⏳	✅	✅	✅	
> Stage7	0	5	⏳	✅	⏳	✅	✅		Release a replica when unavailableReplicas == maxSurge
> Stage8	0	4	✅	⏳	✅	✅			Release another replica for accommodation
> Stage9	0	4	✅	✅	✅	✅			Rolling update finished

This is fantastics and makes a lot of sense, we need to make sure this is well documented!

test/integration/controllers/leaderworkerset_test.go
```diff
@@ -876,18 +870,557 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
...
+						testing.ExpectValidLeaderStatefulSet(ctx, k8sClient, lws, 2)
+						testing.ExpectValidWorkerStatefulSets(ctx, lws, k8sClient, true)
+						testing.ExpectLeaderWorkerSetAvailable(ctx, k8sClient, lws, "All replicas are ready")
+						testing.ExpectStatefulsetPartitionEqualTo(ctx, k8sClient, lws, 0)
+						testing.ExpectLeaderWorkerSetStatusReplicas(ctx, k8sClient, lws, 2, 2)
+					},
+				},
+			},
+		}),
+		ginkgo.Entry("multiple rolling update with maxSurge set", &testCase{
```
liurupeng, line 1326: I know we already have many test case covered. but could you also add one to test when during upgrade, one replica is consistently unready, thus the upgrade is stuck.

liurupeng:
overall lgtm! Thanks Kante!

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -206,107 +206,103 @@ func SetupIndexes(indexer client.FieldIndexer) error {
...
+	if maxSurge > int(lwsReplicas) {
+		maxSurge = int(lwsReplicas)
 	}
+	burstReplicas := lwsReplicas + int32(maxSurge)
+
+	var continuousReadyReplicas, lwsUnreadyReplicas int32
+
+	defer func() {
+		// Reclaim the bursted replicas gradually.
+		if lwsUnreadyReplicas <= int32(maxSurge) {
```
ahg-g, line 245: I find it hard to reason about when using a defer function and named return variables, I suggest to directly put the value in the return statements below and remove the named return variable and defer func. 

Also, I am not sure the calculation here is correct, shouldn't it always be the following (without the if statement above):

```replicas = lwsReplicas + min(utils.NonZeroValue(lwsUnreadyReplicas-maxUnavailable), maxSurge)```

File: api/leaderworkerset/v1/leaderworkerset_types.go [hunk 7]
```diff
@@ -145,6 +149,22 @@ type RollingUpdateConfiguration struct {
 	// +kubebuilder:validation:XIntOrString
 	// +kubebuilder:default=1
 	MaxUnavailable intstr.IntOrString `json:"maxUnavailable,omitempty"`
```
ahg-g, line 151: We should now allow MaxUnavailable=0 if maxSurge>0

ahg-g (commented):
This is great, really impressive!

test/integration/controllers/leaderworkerset_test.go
kerthcet: It's ok, more covered test case, more safe the code. But I guess we already covered this in each integration test because during rolling update, we'll verify the leaderWorkerSet is unavailable also the value of Partition field, which indicates it's stuck, or you have other metrics?

liurupeng: I see, make sense

kerthcet:
kindly ping @ahg-g @liurupeng can we push this forward. Hope not to rebase again. 😅

PR #99, opened 2024-04-12
Fix maxUnavailable couldn't be bigger than replicas

kerthcet:
#### What type of PR is this?
/kind bug

With HPA enabled, the replicas could scale to zero, if we limit the maxUnavailable must less than replicas, it's impossible here.

Also add more strict policies to the validation.

Fixes #

```release-note
None
```


ahg-g:
/lgtm

make sense, thanks!

Issue #100, opened 2024-04-12
Add some EventRecorder events


googs1025:
What would you like to be added:
I noticed that the events generated by EventRecorder are not comprehensive enough. I plan to add more events in the controller, such as events for handling deletion errors or generating some common errors
Why is this needed:
The events generated by EventRecorder can be viewed for more critical events when executing kubectl describe.

googs1025:
done

PR #101, opened 2024-04-12
feat: add event recorder event for lws controller

googs1025:
#### What type of PR is this?
/kind feature

- add event for lws controller 

I have only added an event recorder for the LWS controller and did not add any events for the pod controller. This decision was made because I believe that users will only be interested in events generated by the LWS controller and not those generated by the pod controller.
If there is a need to add events for the pod controller, please let me know and I will add them accordingly.

refer to: https://github.com/kubernetes-sigs/lws/issues/100

```release-note
None
```

### DIFF
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -42,6 +42,7 @@ import (
 
 	leaderworkerset "sigs.k8s.io/lws/api/leaderworkerset/v1"
 	"sigs.k8s.io/lws/pkg/utils"
+	"sigs.k8s.io/lws/pkg/utils/constants"
 	podutils "sigs.k8s.io/lws/pkg/utils/pod"
 	statefulsetutils "sigs.k8s.io/lws/pkg/utils/statefulset"
 )
@@ -90,22 +91,26 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 	partition, err := r.rollingUpdatePartition(ctx, lws)
 	if err != nil {
-		log.Error(err, "rolling partition error")
+		r.Record.Eventf(lws, corev1.EventTypeWarning, constants.RollingUpdateFailedReason, err.Error())
+		log.Error(err, "Rolling partition error")
 		return ctrl.Result{}, err
 	}
 
 	if err := r.SSAWithStatefulset(ctx, lws, partition); err != nil {
+		r.Record.Eventf(lws, corev1.EventTypeWarning, constants.SSALeaderWorkerSetFailedReason, err.Error())
 		return ctrl.Result{}, err
 	}
 
 	// Create headless service if it does not exist.
 	if err := r.createHeadlessServiceIfNotExists(ctx, lws); err != nil {
 		log.Error(err, "Creating headless service.")
+		r.Record.Eventf(lws, corev1.EventTypeWarning, constants.HeadlessServiceCreationFailedReason, err.Error())
 		return ctrl.Result{}, err
 	}
 
 	err = r.updateStatus(ctx, lws)
 	if err != nil {
+		r.Record.Eventf(lws, corev1.EventTypeWarning, constants.UpdateStatusFailedReason, err.Error())
 		return ctrl.Result{}, err
 	}
 
--- a/pkg/utils/constants/constants.go
+++ b/pkg/utils/constants/constants.go
@@ -1,0 +1,36 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


pkg/controllers/leaderworkerset_controller.go
```diff
@@ -363,7 +368,7 @@ func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *l
 			var leaderPod corev1.Pod
 			if err := r.Get(ctx, client.ObjectKey{Namespace: lws.Namespace, Name: sts.Name}, &leaderPod); err != nil {
 				log.Error(err, "Fetching leader pod")
-				return false, err
+				return false, client.IgnoreNotFound(err)
```
liurupeng, line 371: if the leader pod doesn't exist, the worker statefulset will be deleted automatically as well. So if the worker statefulset exist, and its leader pod is not found, we definitely needs to throw errors

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -403,7 +408,7 @@ func (r *LeaderWorkerSetReconciler) updateStatus(ctx context.Context, lws *leade
 	sts := &appsv1.StatefulSet{}
 	if err := r.Get(ctx, types.NamespacedName{Name: lws.Name, Namespace: lws.Namespace}, sts); err != nil {
 		log.Error(err, "Error retrieving leader StatefulSet")
-		return err
+		return client.IgnoreNotFound(err)
```
liurupeng, line 411: I think we need to throw the error here if the leader statefulset exist, since it's supposed to be created already at this point?

pkg/controllers/pod_controller.go
```diff
@@ -76,7 +76,7 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 	// get the leaderWorkerSet object
 	var leaderWorkerSet leaderworkerset.LeaderWorkerSet
 	if err := r.Get(ctx, types.NamespacedName{Name: lwsName, Namespace: pod.Namespace}, &leaderWorkerSet); err != nil {
-		return ctrl.Result{}, err
+		return ctrl.Result{}, client.IgnoreNotFound(err)
```
liurupeng, line 79: if the lws object for this pod doesn't exist, it will be surprising, let's make sure the error is returned

pkg/controllers/pod_controller.go
```diff
@@ -159,7 +159,7 @@ func (r *PodReconciler) handleRestartPolicy(ctx context.Context, pod corev1.Pod,
 			return false, fmt.Errorf("parsing pod name for pod %s", pod.Name)
 		}
 		if err := r.Get(ctx, types.NamespacedName{Name: leaderPodName, Namespace: pod.Namespace}, &leader); err != nil {
-			return false, err
+			return false, client.IgnoreNotFound(err)
```
liurupeng, line 162: throw as well

pkg/controllers/leaderworkerset_controller.go
googs1025: @liurupeng Oh, that's embarrassing. I apologize for not considering that when looking at the code. I thought we were supposed to ignore the errors if they couldn't be found.

pkg/controllers/leaderworkerset_controller.go
googs1025: done

pkg/controllers/pod_controller.go
googs1025: done

pkg/controllers/pod_controller.go
googs1025: done

googs1025:
@liurupeng @kerthcet When I was running the CI, I noticed that errors were occurring, which was not expected. It seems like there might be some issues in the code. I feel like adding the event recorder alone shouldn't cause this error.

googs1025:
Wait for a while and check again. There haven't been any error messages.

kerthcet:
> /retest

Please don't retest if you have a flaky, will lead to the code baseline unstable.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -90,22 +91,26 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 	partition, err := r.rollingUpdatePartition(ctx, lws)
```
kerthcet, line 92: I don't think these events make sense to the system, event usually generated for state transitions, or specific reasons for diagnosis, not logging. 

Please revert them. The only one makes sense to me is the `createHeadlessService` failed event, which is an important signal.

No need to created a new file to container the constants, they're not shared, located at this file is enough.

googs1025: ok, i fix it right now

kerthcet: Discussed offline, hope to have a event to tell us that the rolling update is failed but there's no specific signal indicates the rolling update is failed. So maybe keep the same as sts right now.(Sts doesn't have a event to report that the rolling update is failed as well.)

Fix me if not.

PR #105, opened 2024-04-15
Add publish not ready addresses

### LINKED ISSUE #85: Allow setting  publishNotReadyAddresses=True when creating headless service

An option to set  `publishNotReadyAddresses=True` when creating the headless service of pod hostnames.

**Why is this needed**:

The headless service is used to simplify discovery of pod network IDs within a group for distributed computing, in most cases this is not related to any particular readiness status.

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

kannon92:
#### What type of PR is this?
/kind feature

Adds PublishNotReady to headless services.

Fixes #85 

```release-note
add publishNotReadyAddresses to headless services. defaults to true.
```

### DIFF
--- a/api/leaderworkerset/v1/leaderworkerset_types.go
+++ b/api/leaderworkerset/v1/leaderworkerset_types.go
@@ -85,6 +85,11 @@ type LeaderWorkerSetSpec struct {
 	// RolloutStrategy defines the strategy that will be applied to update replicas
 	// when a revision is made to the leaderWorkerTemplate.
 	RolloutStrategy RolloutStrategy `json:"rolloutStrategy"`
+
+	// PublishNotReadyAddresses indicates that any agent
+	// which deals with endpoints for this Service should disregard any indications of ready/not-ready.
+	// +kubebuilder:default=true
+	PublishNotReadyAddresses bool `json:"publishNotReadyAddresses"`
 }
 
 // Template of the leader/worker pods, the group will include at least one leader pod.
--- a/api/leaderworkerset/v1/leaderworkerset_types.go
+++ b/api/leaderworkerset/v1/leaderworkerset_types.go
@@ -85,11 +85,6 @@ type LeaderWorkerSetSpec struct {
 	// RolloutStrategy defines the strategy that will be applied to update replicas
 	// when a revision is made to the leaderWorkerTemplate.
 	RolloutStrategy RolloutStrategy `json:"rolloutStrategy"`
-
-	// PublishNotReadyAddresses indicates that any agent
-	// which deals with endpoints for this Service should disregard any indications of ready/not-ready.
-	// +kubebuilder:default=true
-	PublishNotReadyAddresses bool `json:"publishNotReadyAddresses"`
 }
 
 // Template of the leader/worker pods, the group will include at least one leader pod.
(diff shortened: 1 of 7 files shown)


api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -85,6 +85,11 @@ type LeaderWorkerSetSpec struct {
 	// RolloutStrategy defines the strategy that will be applied to update replicas
 	// when a revision is made to the leaderWorkerTemplate.
 	RolloutStrategy RolloutStrategy `json:"rolloutStrategy"`
+
+	// PublishNotReadyAddresses indicates that any agent
+	// which deals with endpoints for this Service should disregard any indications of ready/not-ready.
+	// +kubebuilder:default=true
+	PublishNotReadyAddresses bool `json:"publishNotReadyAddresses"`
```
liurupeng, line 92: hi @kannon92 As discussed with @ahg-g  I think we would make it enabled by default for now.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -131,6 +131,7 @@ func (r *LeaderWorkerSetReconciler) createHeadlessServiceIfNotExists(ctx context
 				Selector: map[string]string{
 					leaderworkerset.SetNameLabelKey: lws.Name,
 				},
+				PublishNotReadyAddresses: lws.Spec.PublishNotReadyAddresses,
```
liurupeng, line 134: could you add a test case in the integration/e2e test that verifies the field? thanks

api/leaderworkerset/v1/leaderworkerset_types.go
kannon92: Do you mean that you don't want anyone to set it to false?

pkg/controllers/leaderworkerset_controller.go
kannon92: Added an integration test. 

I don't know if we need a e2e test here.

api/leaderworkerset/v1/leaderworkerset_types.go
kannon92: You may have commented on the Network API. I realized that adding that made it more complicated than it needed to be. I added a default setting to main LWS spec and I default to true in this case.

liurupeng: yes for now, we could add the API field later when there is an ask. wdyt

liurupeng: the reason for not adding a new field for this is bc we can't remove it later when we found out that exposing a new field is not necessary

pkg/controllers/leaderworkerset_controller.go
liurupeng: k, sg

api/leaderworkerset/v1/leaderworkerset_types.go
ahg-g: +1; lets not expose an api field for now until we get a case where we need to make it configurai

ahg-g:
/hold

As @liurupeng mentioned, I don't think we want an explicit api field at this point

api/leaderworkerset/v1/leaderworkerset_types.go
kannon92: sounds good. I updated the PR.

ahg-g:
/lgtm
/approve

Thanks!

PR #121, opened 2024-04-28
Simplify the partition update model

kerthcet:
#### What type of PR is this?

/kind cleanup

Fixes #

```release-note
None
```

### DIFF
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -227,11 +227,7 @@ func (r *LeaderWorkerSetReconciler) rollingUpdatePartition(ctx context.Context, 
 	// Indicates a new rolling update here.
 	if templateUpdated(sts, lws) {
 		// Processing scaling up/down first prior to rolling update.
-		if replicasUpdated(sts, lws) {
-			return min(*lws.Spec.Replicas, *sts.Spec.Replicas), nil
-		} else {
-			return utils.NonZeroValue(replicas - int32(rollingStep)), nil
-		}
+		return min(*lws.Spec.Replicas, *sts.Spec.Replicas), nil
 	}
 
 	// Replicas changes during rolling update.
--- a/test/integration/controllers/leaderworkerset_test.go
+++ b/test/integration/controllers/leaderworkerset_test.go
@@ -501,6 +501,8 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
 					checkLWSState: func(lws *leaderworkerset.LeaderWorkerSet) {
 						testing.ExpectValidLeaderStatefulSet(ctx, lws, k8sClient)
 						testing.ExpectLeaderWorkerSetUnavailable(ctx, k8sClient, lws, "All replicas are ready")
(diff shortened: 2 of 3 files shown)


pkg/controllers/leaderworkerset_controller.go
```diff
@@ -227,11 +227,7 @@ func (r *LeaderWorkerSetReconciler) rollingUpdatePartition(ctx context.Context,
 	// Indicates a new rolling update here.
 	if templateUpdated(sts, lws) {
 		// Processing scaling up/down first prior to rolling update.
-		if replicasUpdated(sts, lws) {
-			return min(*lws.Spec.Replicas, *sts.Spec.Replicas), nil
-		} else {
-			return utils.NonZeroValue(replicas - int32(rollingStep)), nil
-		}
+		return min(*lws.Spec.Replicas, *sts.Spec.Replicas), nil
```
kerthcet, line 230: All the partition rolling update logics move to the the segment of `Calculating the Partition during rolling update`.

kerthcet: Yes, you're right, this will waste another reconciliation but make it a bit more simple. Actually we have the same question right now, e.g. :

We have a lws with replicas=4 and maxUnavailable=3, then we update the lws template and increase the replicas to 6, when rolling update, the Partition will change from 4(first reconciliation) to 3(6-3=3, next reconciliation). 

We can get the right Partition in the first reconciliation, but we need to add more logics, like comparing the value of maxUnavailable and Replicas, same to other places, makes it a bit complex. So maybe this is not a big problem?

kerthcet: So the partition here is to make suer that the new scaling replicas will not use the old template.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -227,11 +227,7 @@ func (r *LeaderWorkerSetReconciler) rollingUpdatePartition(ctx context.Context,
...
 	}
 
 	// Replicas changes during rolling update.
```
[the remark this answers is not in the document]
kerthcet: We just split the `replicasUpdate` from `templateUpdate` apart. So `templateUpdate` will update from scratch but `replicasUpdate` will only perform scaling up/down, when in rolling update, we try to scale the replicas, we hope the replicas in rolling update will not use the old template.

pkg/controllers/leaderworkerset_controller.go
ahg-g: I am ok to proceed, but I don't get your concern. So in the case where the replicas changed, we will execute the first branch, which will set partition to the `min(*lws.Spec.Replicas, *sts.Spec.Replicas)`, if the number of replicas didn't change, then there are no new replicas to worry about, hence setting partition to replicas-maxUnavailable should be safe, what am I missing?

kerthcet: No, we didn't miss anything, what I mean is there're a lot of cases to consider here, so we move the core logic to the last part, just to make sure the Partition maybe not that perfect but it's a safe value.

PR #127, opened 2024-05-03
add kep-115 Subgroup support

### LINKED ISSUE #115: subgroup support for disaggregated serving on TPU

Edwinhr716:
#### What type of PR is this?
/documentation

Fixes #115 

```release-note

```


kep/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,201 @@
...
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+### LeaderWorkerSet API
+We extend the LeaderWorkerSet API to introduce a new field: subGroupSize to opt in and set the number of pods that each subgroup will contain. Current behavior is kept if not set. 
+
+```
+type LeaderWorkerSetSpec struct {
+	 // Number of pods per subgroup
```
ahg-g, line 106: pls document the the constraints on this, 1) it must not be greater than size, 2) size must be divisible by this value 3) it is immutable

kep/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,201 @@
...
+    - Tracks which subgroup the pod is part of 
+  - `leaderworkerset.gke.io/subgroup-worker-index = worker-index%subGroupSize`
+    - index/identity of a pod inside the pod's subgroup
+  - `leaderworkerset.gke.io/subgroup-key` 
+    - Pods that are part of the same subgroup will have an annotation that is a unique hash value will be generated from the name of the leader, and the subgroup-index
+
+To support exclusive placement at the subgroup level, the pod webhook will inject the new labels, and set the pod affinity/anti-affinity on all the pods. If both levels of pod affinity are set, then the leader pod will contain two pod affinities, while the workers will have a node selector, and a single pod affinity set. 
+
+### TPU Environment Variable Injection
+Because the value of the TPU environmental variables will vary between subgroups, the way they are injected will be extended. `TPU_WOKRER_ID` will be the value of `subgroup-worker-index`, while the value of `TPU_HOSTNAMES` will only be a list of the pods in the same subgroup.
```
ahg-g, line 129:
```suggestion
Because the value of the TPU environmental variables will vary between subgroups, the way they are injected will be extended. `TPU_WORKER_ID` will be the value of `subgroup-worker-index`, while the value of `TPU_HOSTNAMES` will only be a list of the pods in the same subgroup.
```

kep/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+LeaderWorkerSet supports the leader not requesting TPU resources. This raises a problem when determining the values of `subgroup-index` and s`ubgroup-worker-index`. Suppose we have two TPU slices with two hosts each. Since the leader doesn’t request TPU resources, there will be four workers + the leader, meaning that one of the workers will have a worker index of four. Because of the way the new labels are calculated, this worker will have a subgroup-index of two, creating three subgroup indices (0,1,2) even though there are only two TPU slices.
```
ahg-g, line 131:
```suggestion
LeaderWorkerSet supports the leader not requesting TPU resources. This raises a problem when determining the values of `subgroup-index` and `subgroup-worker-index`. Suppose we have two TPU slices with two hosts each. Since the leader doesn’t request TPU resources, there will be four workers + the leader, meaning that one of the workers will have a worker index of four. Because of the way the new labels are calculated, this worker will have a subgroup-index of two, creating three subgroup indices (0,1,2) even though there are only two TPU slices.
```

Edwinhr716: fixed

kep/115-Subgroup-support/README.md
Edwinhr716: fixed

kep/115-Subgroup-support/README.md
Edwinhr716: Added

kep/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+<!--
+Detail the things that people will be able to do if this KEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
+As a user, I should be able to have smaller groups within my LeaderWorkerSet deployment and be able to deploy each subgroup on a different topology.
```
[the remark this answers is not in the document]
Edwinhr716: fixed

kep/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,202 @@
...
+this KEP.  Describe why the change is important and the benefits to users. The
+motivation section can optionally provide links to [experience reports] to
+demonstrate the interest in a KEP within the wider Kubernetes community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+Disaggregated serving is an optimization made for LLM inference workloads. It takes advantage of the fact that the two phases of inference have different characteristics and thus it can be beneficial to run them on different machines. State of the art LLM serving frameworks such as [vLLM](https://github.com/vllm-project/vllm/issues/2472) are already adding support for this optimization based on the paper released by [Microsoft](https://www.microsoft.com/en-us/research/publication/splitwise-efficient-generative-llm-inference-using-phase-splitting/). 
+
+LeaderWorkerSet does not currently support having a group for each phase.
```
liurupeng, line 50:
```suggestion
LeaderWorkerSet does not currently support having a group for each phase.
```
This KEP is to have LeaderWorkerSet to support running both prefill/decode servers in one pod group (aka one replica)

kep/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,202 @@
...
+type LeaderWorkerSetSpec struct {
+	 // Number of pods per subgroup. This value is immutable,
+   // and must not be greater than LeaderWorkerSet.Spec.Size.
+	SubGroupSize *int32 'json:"subGroupSize,omitempty'
+} 
+```
+
+### Exclusive Topology Support
+LeaderWorkerSet can guarantee that the leader and the workers are placed in the same topology if the `leaderworkerset.gke.io/exclusive-topology` annotation is set. Similarly, we will support that the pods within the same subgroup will be placed in the same topology with a new annotation `leaderworkerset.gke.io/exclusive-topology-subgroup`, so that the new changes can support up to two levels of pod affinity. 
+
```
liurupeng, line 114: We could give an example for two levels of exclusive placement and describe how pods will be scheduled at last.

kep/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,202 @@
...
+The goal is to ensure that we don't accept enhancements with inadequate testing.
+
+All code is expected to have adequate tests (eventually with coverage
+expectations). Please adhere to the [Kubernetes testing guidelines][testing-guidelines]
+when drafting this test plan.
+
+[testing-guidelines]: https://git.k8s.io/community/contributors/devel/sig-testing/testing.md
+-->
+
+[ ] I/we understand the owners of the involved components may require updates to
```
liurupeng, line 151: qq, what is the "[ ]" at the start of the line?

kep/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,202 @@
...
+Describe what tests will be added to ensure proper quality of the enhancement.
+
+After the implementation PR is merged, add the names of the tests here.
+-->
+
+- The pod-webhook integration tests should test
+  - Pod Affinity/Anti-Affinity is injected properly in leader and workers
+  - The new labels are only added if `SubGroupSize` is set
+  - The expected TPU environment variables are injected given different combinations of leader requests resources and subgroup sizes. 
+
```
liurupeng, line 189: e2e tests can be added as well if we don't request TPU resources just to test the subgrouping and exclusive placement right?

kep/115-Subgroup-support/README.md
Edwinhr716: Fixed

kep/115-Subgroup-support/README.md
Edwinhr716: A checkbox which is part of the template

kep/115-Subgroup-support/README.md
Edwinhr716: I added to test the subgrouping, but can e2e be added for exclusive placement? I don't see an existing e2e test for that

liurupeng: e2e test means we will really schedule the pods into the nodes. As long as there are more than 1 nodes, then this should be possible right? even if only one nodes, we could still verify the pods have correct labels, and a certain pod is scheduled while others are pending?

kep/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,208 @@
...
+#### End to End Tests
+
+- Test that LWS deployment with subgrouping enabled can be deployed
```
liurupeng, line 194:
```suggestion
- Test that LWS deployment with subgrouping enabled will have correct pod labels, pod exclusive placement and work well with other features enabled, like failure handling and rolling update.
```

kep/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,208 @@
...
+```
+type LeaderWorkerSetSpec struct {
+	 // Number of pods per subgroup. This value is immutable,
+   // and must not be greater than LeaderWorkerSet.Spec.Size.
+	SubGroupSize *int32 'json:"subGroupSize,omitempty'
+} 
+```
+
+### Exclusive Topology Support
+LeaderWorkerSet can guarantee that the leader and the workers are placed in the same topology if the `leaderworkerset.gke.io/exclusive-topology` annotation is set. Similarly, we will support that the pods within the same subgroup will be placed in the same topology with a new annotation `leaderworkerset.gke.io/exclusive-topology-subgroup`, so that the new changes can support up to two levels of pod affinity.
```
kerthcet, line 115: Avoid to use vendored words, we're now `leaderworkerset.sigs.k8s.io/exclusive-topology`

kep/115-Subgroup-support/README.md
kerthcet: +1 to have an example.

kep/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,208 @@
...
+proposal will be implemented, this is the place to discuss them.
+-->
+### LeaderWorkerSet API
+We extend the LeaderWorkerSet API to introduce a new field: subGroupSize to opt in and set the number of pods that each subgroup will contain. Current behavior is kept if not set. 
+
+```
+type LeaderWorkerSetSpec struct {
+	 // Number of pods per subgroup. This value is immutable,
+   // and must not be greater than LeaderWorkerSet.Spec.Size.
+	SubGroupSize *int32 'json:"subGroupSize,omitempty'
```
kerthcet, line 110: Should we put under `leaderWorkerTemplate`, parallel to size? Because it represents the topology of the template.

kep/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,208 @@
...
+} 
+```
+
+### Exclusive Topology Support
+LeaderWorkerSet can guarantee that the leader and the workers are placed in the same topology if the `leaderworkerset.gke.io/exclusive-topology` annotation is set. Similarly, we will support that the pods within the same subgroup will be placed in the same topology with a new annotation `leaderworkerset.gke.io/exclusive-topology-subgroup`, so that the new changes can support up to two levels of pod affinity. 
+
+### Implementation
+- Two new annotations will be added
+  - `leaderworkerset.gke.io/exclusive-topology-subgroup `
```
kerthcet, line 119: Let's keep the name consistent as word subgroup always **at the head** or **at the end**, head maybe?

kep/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+### TPU Environment Variable Injection
+Because the value of the TPU environmental variables will vary between subgroups, the way they are injected will be extended. `TPU_WORKER_ID` will be the value of `subgroup-worker-index`, while the value of `TPU_HOSTNAMES` will only be a list of the pods in the same subgroup.
+
+LeaderWorkerSet supports the leader not requesting TPU resources. This raises a problem when determining the values of `subgroup-index` and `subgroup-worker-index`. Suppose we have two TPU slices with two hosts each. Since the leader doesn’t request TPU resources, there will be four workers + the leader, meaning that one of the workers will have a worker index of four. Because of the way the new labels are calculated, this worker will have a subgroup-index of two, creating three subgroup indices (0,1,2) even though there are only two TPU slices.
+
+If the leader does not request TPU resources, then the labels will have the following values
```
kerthcet, line 136: Will this lead to vendor problem, because the logic is special as normal. Hope not here.

kep/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,208 @@
...
+    - Tracks which subgroup the pod is part of 
+  - `leaderworkerset.gke.io/subgroup-worker-index = worker-index%subGroupSize`
+    - index/identity of a pod inside the pod's subgroup
+  - `leaderworkerset.gke.io/subgroup-key` 
+    - Pods that are part of the same subgroup will have an annotation that is a unique hash value will be generated from the name of the leader, and the subgroup-index
+
+To support exclusive placement at the subgroup level, the pod webhook will inject the new labels, and set the pod affinity/anti-affinity on all the pods. If both levels of pod affinity are set, then the leader pod will contain two pod affinities, while the workers will have a node selector, and a single pod affinity set. 
+
+### TPU Environment Variable Injection
+Because the value of the TPU environmental variables will vary between subgroups, the way they are injected will be extended. `TPU_WORKER_ID` will be the value of `subgroup-worker-index`, while the value of `TPU_HOSTNAMES` will only be a list of the pods in the same subgroup.
```
kerthcet, line 132: Do you mean TPU_WORKER_HOSTNAMES rather than TPU_HOSTNAMES

Edwinhr716: Yes, fixed

kep/115-Subgroup-support/README.md
Edwinhr716: Switched it to always be head

kep/115-Subgroup-support/README.md
Edwinhr716: added example

keps/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,223 @@
...
+- `leaderworkerset.sigs.k8s.io/subgroup-exclusive-topology: cloud.google.com/gke-nodepool`
+
+The leader pod will first be scheduled on a placement group. Once a worker pod is created, it will follow the leader to the placement group. 
+Afterward, it will be scheduled into a nodepool that has other pods with the same subgroup index. So, the placement will look something like this
+
+![kep-115 scheduling](https://github.com/kubernetes-sigs/lws/assets/86417275/8127f005-2f50-4648-aa1a-6b37cb3b813a)
+
+### Implementation
+- Two new annotations will be added
+  - `leaderworkerset.sigs.k8s.io/ubgroup-exclusive-topology`
```
kerthcet, line 134:
```suggestion
  - `leaderworkerset.sigs.k8s.io/subgroup-exclusive-topology`
```

keps/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,223 @@
...
+proposal will be implemented, this is the place to discuss them.
+-->
+### LeaderWorkerSet API
+We extend the LeaderWorkerSet API to introduce a new field: subGroupSize to opt in and set the number of pods that each subgroup will contain. Current behavior is kept if not set. 
+
+```
+type LeaderWorkerSetSpec struct {
+	 // Number of pods per subgroup. This value is immutable,
+   // and must not be greater than LeaderWorkerSet.Spec.Size.
+	SubGroupSize *int32 'json:"subGroupSize,omitempty'
```
kerthcet, line 112: Put former question here: 

Should we put this field under leaderWorkerTemplate, parallel to `Size`? Because it represents the topology of the template.

kerthcet (commented):
Thanks for the effort. I have a general question here about the splitwise approach:

As we know, prefill is much slower than decoding phase (I mean for a sentence) because we only process prefill once, but decode for multi times , so does the division in average really make sense? What if we want the ratio like 3:7 or even more complicated 2:3:5? The latter is somehow like YAGNI, just for imaging.

kerthcet:
BTW, I'm debugging the flaky test, just forget it. Sorry for that.

keps/115-Subgroup-support/README.md
Edwinhr716: Yeah that makes sense to me. I'll move it to the LeaderWorkerTemplate to parallel Size

Edwinhr716:
A 3:7 ratio could be achieved by having 10 subgroups, 3 for prefill and 7 for decode. There doesn't necessarily have to be a subgroup per stage. Similar thing with the 2:3:5 ratio. The only requirement is that each subgroup has the same number of machines. So in the 3:7 ratio, if you have 10 machines, you can have each subgroup have a machine each and achieve the 3:7 ratio.

kep/115-Subgroup-support/README.md
Edwinhr716: By vendor problem you mean the logic changes based on whether or not the user uses TPUs?

kerthcet: I mean we should not change the label values because of different scenarios, they're identifiers, but  we can leave the logics to the injecting part as we do today. I guess it's possible.

keps/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,223 @@
+# KEP #115 Subgroup Support
```
kerthcet, line 1:
```suggestion
# KEP-115: Subgroup Support
```

For formatting.

kerthcet: Also we still miss a kep.yaml, this is important for later tracking. Sorry for the additional works but it's a standard way for kube.

keps/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,223 @@
...
+
+### TPU Environment Variable Injection
+Because the value of the TPU environmental variables will vary between subgroups, the way they are injected will be extended. `TPU_WORKER_ID` will be the value of `subgroup-worker-index`, while the value of `TPU_WORKER_HOSTNAMES` will only be a list of the pods in the same subgroup.
+
+LeaderWorkerSet supports the leader not requesting TPU resources. This raises a problem when determining the values of `subgroup-index` and `subgroup-worker-index`. Suppose we have two TPU slices with two hosts each. Since the leader doesn’t request TPU resources, there will be four workers + the leader, meaning that one of the workers will have a worker index of four. Because of the way the new labels are calculated, this worker will have a subgroup-index of two, creating three subgroup indices (0,1,2) even though there are only two TPU slices.
```
kerthcet, line 149: I can somehow know what you want to do here, but can't fully understand the words, like `we have two TPU slices with two hosts each`, maybe an example would be great or something else to make this more clear, this will benefit the future readers a lot. Thanks.

kerthcet:
> A 3:7 ratio could be achieved by having 10 subgroups, 3 for prefill and 7 for decode.

Got your idea, but then how we can inject different scheduling directives for different subgroups? Like I set the Size=10, and subGroupSize=1, then we got 10 subGroups, and I want to set the first three subgroups for prefill, the rest for decoding. This is not exposed by the API right now I think.

keps/115-Subgroup-support/README.md
Edwinhr716: I'll add a graph for this as well

liurupeng:
@kerthcet this is how we image the prefill/decode will work. Ex: we create a lws with one replica (1 leader, 11 workers), and use the subgroup feature to create three subgroups, each subgroup will have an exclusive placement label to allow only being scheduled on one TPU node pool. So at last, there will be 1 pod group with 12 pods scheduled on 3 three TPU pod slices node pool. And the servers that doing prefill and decode is decided after the workload started. A container on the leader pod will connect to all the worker pods that within the same group, and there should be a script to predefine the ratio of prefill:decode (1:2). Then the container in the leader pod will assign one node pool to be prefill and 2 node pools to be decode.

keps/115-Subgroup-support/README.md
Edwinhr716: Fixed and added kep.yaml

liurupeng:
So in the API field, we only need to define what is the subgroupsize, that's bc we expect the model server container in the leader pod to assign the prefill/decode slices during runtime.

kerthcet:
Fully understand your intension here, thanks for the detailed explanation, still two questions:
- We divided into three subgroups, why the last two subgroups should be exclusive, doesn't the TPU node pool the same pool, I think the TPU node pool you mean here is a node pool with the same type of TPUs.
- Since what we wanna in the end is prefill:decode=1:2, can't we just achieve this from lws side, is that make sense, let's forget about whether that's possible or not at this moment, If possible, is that as you expected.

Sorry, I didn't explore the splitwise approach before, so lots of questions. Hope you don't mind.

kerthcet:
Or API like:
```
type LeaderWorkerTemplate struct {
	Partitions []Partition
}

type Partition struct {
	Name    string
	Percent string
}
```

Then we'll have:
```
leaderWorkerTemplate:
       partitions:
       - name: prefill
         percent: 20%
       - name: decode
         percent: 80%
```

keps/115-Subgroup-support/README.md
kerthcet: So the subgroupsize here is 2? But why the worker index 4 has a subgroup-worker-index = 3, doesn't it be (workerIndex - 1) % subGroupSize = (4-1) % 2 = 1.

liurupeng:
The use case of the subgroup feature is to support disaggregated serving with multi-host inference (large model)

keps/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,227 @@
...
+motivation section can optionally provide links to [experience reports] to
+demonstrate the interest in a KEP within the wider Kubernetes community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+Disaggregated serving is an optimization made for LLM inference workloads. It takes advantage of the fact that the two phases of inference have different characteristics and thus it can be beneficial to run them on different machines. State of the art LLM serving frameworks such as [vLLM](https://github.com/vllm-project/vllm/issues/2472) are already adding support for this optimization based on the paper released by [Microsoft](https://www.microsoft.com/en-us/research/publication/splitwise-efficient-generative-llm-inference-using-phase-splitting/). 
+
+
+This KEP is to have LeaderWorkerSet to support running both prefill/decode servers in one pod group (aka one replica)
```
liurupeng, line 52:
```suggestion
This KEP is to have LeaderWorkerSet to support running both prefill/decode servers in one pod group (aka one replica) for LLM on multi-host
```

keps/115-Subgroup-support/README.md
Edwinhr716: made a mistake, fix it

kerthcet: You mentioned `meaning that one of the workers will have a worker index of four,` I think you mean the `Worker 4` in the pic, then `Because of the way the new labels are calculated, this worker will have a subgroup-index of two`, according to the formula, `leaderworkerset.sigs.k8s.io/subgroup-index = (workerIndex - 1) / subGroupSize = (4-1) / 2 =1`, but you mentioned the result is two, anything I misunderstood here?

kep/115-Subgroup-support/README.md
kerthcet: This is the last big concern as leaving the TPU logic to accelerators as we do today. see https://github.com/kubernetes-sigs/lws/blob/dde4a46b1db6ce431449ef19c0566000d878ba6e/pkg/utils/accelerators/tpu.go#L105

keps/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,227 @@
...
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+### LeaderWorkerSet API
+We extend the LeaderWorkerSet API to introduce a new field: subGroupSize to opt in and set the number of pods that each subgroup will contain. Current behavior is kept if not set. 
+
+```
+type LeaderWorkerTemplate struct {
+	 // Number of pods per subgroup. This value is immutable,
+   // and must not be greater than LeaderWorkerSet.Spec.Size.
```
ahg-g, line 111: We should mention what values are allowed: Size should be divisible by subGroupSize, in which case the sub groups will be of equal size, or Size-1 is divisible by subGroupSize, in which case the leader is considered as the extra pod and will be part of the first subgroup.

keps/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,227 @@
...
+The overall workflow will look like this 
+
+![kep-115](https://github.com/kubernetes-sigs/lws/assets/86417275/ff9fc93d-c738-4c09-abc8-50a7b16d49df)
+
+Suppose we have an LWS deployment with 2 replicas, subGroupSize 4, size 8, and the following annotations: 
+- `leaderworkerset.sigs.k8s.io/exclusive-topology: topology-1` 
+- `leaderworkerset.sigs.k8s.io/subgroup-exclusive-topology: topology-2`
+
+The leader pod will first be scheduled on a placement group. Once a worker pod is created, it will follow the leader to the placement group. 
+Afterward, it will be scheduled into a nodepool that has other pods with the same subgroup index. So, the placement will look something like this
```
ahg-g, line 128: s/node pool/topology-2

liurupeng:
Another reason that we can't use the Partition as the API is because the Partition we have for the prefill:decode can't be passed to the application layer that assigns prefill/decode slices.

liurupeng:
@ahg-g and @kerthcet want to take a final look before approving the KEP?

keps/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+LeaderWorkerSet supports the leader not requesting TPU resources. This raises a problem when determining the values of `subgroup-index` and `subgroup-worker-index`. Suppose we have two TPU slices with two vm hosts each. Since the leader doesn’t request TPU resources, it won't take full ownership of the TPU vm, allowing 
+a worker that does request TPU resources to run in the same vm. Therefore, there will be four workers + the leader, meaning that one of the workers will have a worker index of four, a
+as shown in the picture below. Because of the way the new labels are calculated, this worker will have a subgroup-index of two, creating three subgroup indices (0,1,2) even though there are only two TPU slices.
+
+![Leader doesn't request TPU resources](https://github.com/kubernetes-sigs/lws/assets/86417275/3a1292ac-7221-4324-acda-ff33702838f9)
+
+If the leader does not request TPU resources, then the labels will have the following values
+`leaderworkerset.sigs.k8s.io/subgroup-index = (workerIndex - 1) / subGroupSize`
+`leaderworkerset.sigs.k8s.io/subgroup-worker-index = (workerIndex - 1) % subGroupSize`
```
ahg-g, line 157: We can't change the semantics of subgroup-index and subgroup-worker-index based on tpu requests, what we change is the value we set to TPU_WORKER_ID.

keps/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+![kep-115 scheduling](https://github.com/kubernetes-sigs/lws/assets/86417275/33c49e9f-84da-4ac9-8e2b-973b532a0756)
+
+### Implementation
+- Two new annotations will be added
+  - `leaderworkerset.sigs.k8s.io/subgroup-exclusive-topology`
+  - `leaderworkerset.sigs.k8s.io/subgroup-size` 
+- Three new labels will be added,
+  - `leaderworkerset.sigs.k8s.io/subgroup-index = worker-index/subGroupSize`
+    - Tracks which subgroup the pod is part of
```
ahg-g, line 138: I don't think we need this for now, mostly because I am not sure about the semantics we want to have for it if the leader is not considered an actual "worker" (the case where the first subgroup has an extra pod, the leader).

keps/115-Subgroup-support/README.md
ahg-g: As suggested above, I don't think we need subgroup-worker-index for now.

kerthcet:
> prefill:decode:decode=1:1:1 since we would only need one decode right?

The decode means partition here not different decode process.

kerthcet:
Last big concern https://github.com/kubernetes-sigs/lws/pull/127#discussion_r1599341609 I think I'm aligned with @ahg-g

keps/115-Subgroup-support/README.md
Edwinhr716: removed subgroup-worker-index

keps/115-Subgroup-support/README.md
Edwinhr716: ditto

keps/115-Subgroup-support/README.md
Edwinhr716: Added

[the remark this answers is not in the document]
Edwinhr716: fixed

keps/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+How will security be reviewed, and by whom?
+
+How will UX be reviewed, and by whom?
+
+Consider including folks who also work outside the SIG or subproject.
+-->
+To avoid the changes from affecting initial functionality and to not set 
+unnecessary labels, there will be no default behavior if `SubGroupSize` is not 
+set
```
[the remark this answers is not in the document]
Edwinhr716: added

keps/115-Subgroup-support/README.md
Edwinhr716: The written example is to show the problem that arises if normal behavior is followed, while the diagram is the fixed version.

keps/115-Subgroup-support/README.md
```diff
@@ -0,0 +1,232 @@
...
+
+```
+type LeaderWorkerTemplate struct {
+	 // The number of pods per subgroup. This value is immutable,
+   // and must not be greater than LeaderWorkerSet.Spec.Size.
+   // Size must be divisible by subGroupSize in which case the 
+   // subgroups will be of equal size. Or size - 1 is divisible
+   // by subGroupSize, in which case the leader is considered as
+   // the extra pod, and will be part of the first subgroup.
+	SubGroupSize *int32 'json:"subGroupSize,omitempty'
```
ahg-g, line 116: I think we should put this inside a `SubgroupingPolicy` struct, there is a good chance we will have more parameters related to that in the future (how to deal with the leader, subgroup exclusive policy)

Edwinhr716: makes sense, added the struct

keps/115-Subgroup-support/README.md
ahg-g: pls address this

ahg-g:
one last nit, otherwise looks good to me

keps/115-Subgroup-support/README.md
Edwinhr716: fixed

PR #140, opened 2024-05-11
chore: upgrade e2e kind version

googs1025:
#### What type of PR is this?
/kind cleanup

- Currently we support k8s versions: v1.27, v1.28, v1.29, so our e2e test image should be upgraded to v1.29.1
- kind version 0.18 -> 0.22

Fixes #
None

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ ARTIFACTS ?= $(PROJECT_DIR)/bin
 
 INTEGRATION_TARGET ?= ./test/integration/...
 
-E2E_KIND_VERSION ?= kindest/node:v1.27.1
+E2E_KIND_VERSION ?= kindest/node:v1.29.1
 USE_EXISTING_CLUSTER ?= false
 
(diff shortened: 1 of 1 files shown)


kerthcet:
Can we support 1.30 now, it's released.

kerthcet:
That's mean you should add an e2e test of v1.30 to test-infra. I think you got what I mean as talked offline.

googs1025:
> Can we support 1.30 now, it's released.
@kerthcet 
Kind has not been updated yet, the latest version is 1.29. Maybe we should set it as todo and wait for kind to be updated before updating.
https://github.com/kubernetes-sigs/kind/releases

kerthcet (commented):
/lgtm
/approve

Make sense to me to catch the latest version as Kubernetes are backward compatible.

googs1025:
> That's mean you should add an e2e test of v1.30 to test-infra. I think you got what I mean as talked offline.

I will submit PR in the next few days. @kerthcet

PR #151, opened 2024-05-21
Remove resource limit for auth proxy

liurupeng:
#### What type of PR is this?
This PR removes the resource limit for auth proxy since we saw OOM issue in prod use case

Fixes #

```release-note

```

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -26,9 +26,6 @@ spec:
           protocol: TCP
           name: https
         resources:
-          limits:
-            cpu: 500m
-            memory: 128Mi
           requests:
             cpu: 5m
             memory: 64Mi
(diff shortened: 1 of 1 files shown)


config/default/manager_auth_proxy_patch.yaml
```diff
@@ -26,9 +26,6 @@ spec:
           protocol: TCP
           name: https
         resources:
-          limits:
-            cpu: 500m
-            memory: 128Mi
```
ahg-g, line 31: it is good practice to keep the memory limits, you can increase it, but lets not completely drop it

ahg-g: You can remove the cpu limit, it is a good practice actually, just the memory that we should set because it impacts the stability of the node of the container has a bug and tries to use all the memory.

liurupeng: updated

PR #152, opened 2024-05-24
Switch the order of env var for LWS_LEADER_ADDRESS

liurupeng:
#### What type of PR is this?
Quick change to make LWS_LEADER_ADDRESS to be the first container env var so we could reference that in the template from other env vars

Fixes #

```release-note

```

### DIFF
--- a/pkg/utils/pod/pod_utils.go
+++ b/pkg/utils/pod/pod_utils.go
@@ -97,7 +97,7 @@ func addEnvVarIfNotExists(c *corev1.Container, e corev1.EnvVar) {
 			return
 		}
 	}
-	c.Env = append(c.Env, e)
+	c.Env = append([]corev1.EnvVar{e}, c.Env...)
 }
 
 // AddLWSVariables adds LWS_LEADER_ADDRESS environment variable to every container.
--- a/test/integration/webhooks/pod_test.go
+++ b/test/integration/webhooks/pod_test.go
@@ -465,7 +465,7 @@ var _ = ginkgo.Describe("leaderworkerset pod defaulting, creation and update", f
 				return nil
 			},
 		}),
-		ginkgo.Entry("Pod has leader address env var populated", &testDefaultingCase{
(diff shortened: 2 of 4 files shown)


pkg/utils/pod/pod_utils.go
```diff
@@ -97,7 +97,7 @@ func addEnvVarIfNotExists(c *corev1.Container, e corev1.EnvVar) {
 			return
 		}
 	}
-	c.Env = append(c.Env, e)
+	c.Env = append([]corev1.EnvVar{e}, c.Env...)
```
kerthcet, line 100: Hostly this makes no sense to me as we can't guarantee the env orders, like once we add more envs, the `LWS_LEADER_ADDRESS` will not be the first one.

kerthcet (commented):
What's the motivation here, can't get it. If the c.Env containers the `LWS_LEADER_ADDRESS`, I think we can get it from the slice although we have to iterate all its elements.

pkg/utils/pod/pod_utils.go
liurupeng: yes, if there is another component modifying the env vars, it's very likely the first env var will not be LWS_LEADER_ADDRESS. The way that env var is resolved is sequential. So if I have an env var that want to use the value of leader address, then that env var need to be added after leader address env var

liurupeng:
In general, I'd agree with a better approach which is to have an annotation to store the leader address

kerthcet:
Talked offline, we want to put the internal env variables in prior to the use specified ones, then envs like `LWS_LEADER_ADDRESS` can be reused, then makes sense to me. Please add a comment to explain the reason. Thanks.

By the way, hope to use as less annotations as we can since we have a bunch of annotations already, my two cenets.

pkg/utils/pod/pod_utils.go
ahg-g: I think it is fine, we are not promising anything.

Issue #162, opened 2024-06-10
Enhance scheduling capabilities for a group of pods


vie-serendipity:
**What would you like to be added**:
Add a field like `ScheduleMode`. For scheduling a set of pods we may need more strict control. When deploying a distributed inference service, a set of pods, that is, a head plus several workers should be scheduled to neighboring nodes to reduce the communication cost between them.
**Why is this needed**:
There are many types of resources in a k8s cluster, and if the scheduling to nodes is constrained only by `requests` field, it is possible that the final distributed inference may not be too good.
An example is that a group of pods should be dispatched to nodes that have a special way of connecting to them, such as nvlink. Another example is that nodes in a cluster (a non-standard k8s cluster) may be across data centers, and a group of pods should be dispatched into the same data center.
**Completion requirements**:
I'm not sure about the eventual changes for the api, just an enhancement request. I'm also not sure if such a requirement is a reasonable enhancement request, and I'd be happy to contribute if it is.
This enhancement requires the following artifacts:

- [x] Design doc
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

vie-serendipity:
@googs1025 
> An example is that a group of pods should be dispatched to nodes that have a special way of connecting to them, such as nvlink. Another example is that nodes in a cluster (a non-standard k8s cluster) may be across data centers, and a group of pods should be dispatched into the same data center.

In this scenario, each set of pods uses the same template, so the entire lws corresponding pods will be dispatched to the same data center. However, due to the fact that gpu is distributed in different data centers, so different groups of pods under lws should be scheduled to different data centers. So I think affinity or selector can't fulfill such requirement.

googs1025:
> @googs1025 
> 
> > An example is that a group of pods should be dispatched to nodes that have a special way of connecting to them, such as nvlink. Another example is that nodes in a cluster (a non-standard k8s cluster) may be across data centers, and a group of pods should be dispatched into the same data center.
> 
> 
> 
> In this scenario, each set of pods uses the same template, so the entire lws corresponding pods will be dispatched to the same data center. However, due to the fact that gpu is distributed in different data centers, so different groups of pods under lws should be scheduled to different data centers. So I think affinity or selector can't fulfill such requirement.


[...]

Issue #167, opened 2024-06-28
LeaderWorkerSet doesn't support gang-scheduling


xgchena:
**What happened**:

It seems that the LeaderWorkerSet doesn't support gang-scheduling of a group of pods. If more replicas are scheduled at the same time, and there are not enough capacity to host them all, then the scheduler may prioritize scheduling of leader pods, and leave their worker pods pending forever.

**What you expected to happen**:

LeaderWorkerSet should support gang-scheduling, i.e. the pods of a group are either scheduled all together, or nothing.

**How to reproduce it (as minimally and precisely as possible)**:

I have tried with [the vllm example](https://github.com/kubernetes-sigs/lws/tree/9c231ddf09377887d7932bdb1e23e1b154c2d7e9/docs/examples/vllm) with an EKS cluster which has 4 nodes, each node has 1 GPU and enough resources to meet the requests of the pods. The example manifest uses size 2 and replicas 2, in total 4 pods.

* It works fine for the initial deployment, as each node can host one pod.

```
$ kubectl apply -f lws.yaml
$ kubectl get pods
NAME       READY   STATUS    RESTARTS   AGE
vllm-0     1/1     Running   0          10m
vllm-0-1   1/1     Running   0          10m
vllm-1     1/1     Running   0          3m43s
vllm-1-1   1/1     Running   0          3m43s
```

* But scheduling problem shows up if lws is scaled in and then scaled out to more replicas than the available nodes can host. See below, the first group was good, but the rest of the nodes were used to schedule two leader pods who workers were pending forever due to no capacity. Eventually only one group was working. 

```
$ kubectl scale --replicas=0 lws/vllm

$ kubectl get pods
No resources found in default namespace.

$ kubectl scale --replicas=4 lws/vllm
$ k get pods
NAME       READY   STATUS    RESTARTS   AGE
vllm-0     0/1     Running   0          7s
vllm-0-1   1/1     Running   0          7s
vllm-1     0/1     Running   0          7s
vllm-1-1   0/1     Pending   0          7s
vllm-2     0/1     Running   0          7s
vllm-2-1   0/1     Pending   0          7s
vllm-3     0/1     Pending   0          7s
vllm-3-1   0/1     Pending   0          7s
```

* The problem is more obvious in the following example when the scheduler prioritized scheduling the first 4 leader pods. Eventually no group was working.

```
$ kubectl scale --replicas=0 lws/vllm

$ kubectl get pods
No resources found in default namespace.

[...]

PR #168, opened 2024-07-01
add kep-165 Colocated Placement

vie-serendipity:
#### What type of PR is this?

/kind documentation
Support colocated placement of multiple pod groups

Fixes https://github.com/kubernetes-sigs/lws/issues/162

I don't know how to upload image to aws and I just push image to repo. This should be fixed later.

```release-note

```


vie-serendipity:
any suggestions?

kerthcet:
Will visit this API later for bandwidth, sorry for that.

liurupeng:
@vie-serendipity sry for the late reply, I was on vacation last week, checking it now

keps/162-colocated-placement/README.md
```diff
@@ -0,0 +1,268 @@
...
+This KEP is to have LeaderWorkerSet to support colocated topology placement of one pod group.
+
+### Goals
+
+<!--
+List the specific goals of the KEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+- Add a new way of topology scheduling and allow multiple pod groups land on the same domain
```
liurupeng, line 72: for one pod group (one multi-host inference replica), we need to colocate the pods, but for different pod groups, why we need to put them in the same domain?

keps/162-colocated-placement/README.md
```diff
@@ -0,0 +1,268 @@
...
+  // +kubebuilder:default=None
+  // +kubebuilder:validation=Enum={ExclusiveTopologyPlacementPolicyType,ColocatedTopologyPlacementPolicyType,NoneTopologyPlacementPolicyType}
+  type TopologyPlacementPolicyType `json:"type"`
+  topologyKey *string `json:"topologyKey",omitempty`
+}
+```
+
+### SubGroup Policy Support
+Colocated placement can support subgroup policy as well.
+Compared with [exclusive support for subgroup policy](https://github.com/kubernetes-sigs/lws/assets/86417275/ff9fc93d-c738-4c09-abc8-50a7b16d49df), the workflow is almost identical, as follows:
```
liurupeng, line 148: the subgroup feature will only have one leader pod for all the pods in the same topology

liurupeng:
overall lgtm, but one more part to consider is the existing lws workload that use exclusive placement. Do we want to keep two set of APIs to use different placement strategies?

liurupeng:
@ahg-g and @kerthcet any objection to this feature?

keps/162-colocated-placement/README.md
vie-serendipity: I might confuse you with my expression, I mean allowing all leaders to schedule freely across any topology, not restricting one pod group to a topology just  like exclusive placement.

keps/162-colocated-placement/README.md
vie-serendipity: For the subgroup policy, it seems like it just needs to ensure that the leader and workers are in the same topology, so I think colocated placement should meet its requirements.

liurupeng:
> @ahg-g and @kerthcet any objection to this feature?

Abdullah is on vacation, @kerthcet could you review this as well. In general, I think allowing configuring more customized scheduling semantics will be very helpful

liurupeng:
@ahg-g could you help check this one when got some time? thanks!

keps/162-colocated-placement/README.md
```diff
@@ -0,0 +1,268 @@
...
+Similar to implementation of exclusive placement, we only need to remove podAntiAffinity part in the pod webhook. Exclusive and colocated placement are contradictory. If we add this feature to annotations too, when we enable them both in the annotations, the meaning becomes seriously ambiguous. So we think they should be unified to one field and it is necessary to change API specs.
+
+### LeaderWorkerSet API
+
+```go
+type TopologyPlacementPolicyType string
+
+const (
+  ExclusiveTopologyPlacementPolicyType TopologyPlacementPolicyType = "Exclusive" 
+  ColocatedTopologyPlacementPolicyType TopologyPlacementPolicyType = "Colocated"
```
ahg-g, line 129: How will we implement this? 

If we implement this using pod-affinity (i.e., same as exclusive placement, but removing the anti-affinity term), we will run into race conditions: consider the case where two groups try to schedule on the same domain at the same time but there is capacity only for one, there is a possibility that both will partially schedule on the domain, and none will schedule fully.

vie-serendipity: That‘s a tricky case I didn't consider before. It seems like there‘s already an issue when scheduling multiple pod groups with limited resources. Refer to https://github.com/kubernetes-sigs/lws/issues/167
Co-scheduling plugin conflicts with `LeaderReady` startup policy.
In this situation, I think it has to rely on the autoscaler or adding more resources manully.

vie-serendipity: I understand that a deadlock with the plugin only happens when the `leaderready` startup policy is adopted. We can do gang scheduling with the pod group declared in the plugin. This can fix the problem you mentioned.
It's worth noting that colocated placement policy and leaderready startup policy can't be set together, and we have to validate in the webhook.
But I'm not sure whether relying on an out-of-tree scheduler plugin is appropriate.

keps/162-colocated-placement/README.md
```diff
@@ -0,0 +1,268 @@
...
+  NoneTopologyPlacementPolicyType TopologyPlacementPolicyType = "None"
+)
+
+type LeaderWorkerTemplate struct {
+  // +optional
+  TopologyPlacementPolicy TopologyPlacementPolicy `json:"topologyPlacementPolicy",omitempty`
+}
+
+type TopologyPlacementPolicy struct {
```
kerthcet, line 138: Maybe GroupPlacementPolicy?

keps/162-colocated-placement/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+const (
+  ExclusiveTopologyPlacementPolicyType TopologyPlacementPolicyType = "Exclusive" 
+  ColocatedTopologyPlacementPolicyType TopologyPlacementPolicyType = "Colocated"
+  NoneTopologyPlacementPolicyType TopologyPlacementPolicyType = "None"
+)
+
+type LeaderWorkerTemplate struct {
+  // +optional
+  TopologyPlacementPolicy TopologyPlacementPolicy `json:"topologyPlacementPolicy",omitempty`
```
kerthcet, line 135: Should this a slice in case we'll append more than one policies to the lws, and the policies should be ANDed?

keps/162-colocated-placement/README.md
kerthcet: I guess this is more related to scheduling rather than workload itself? This is an exist problem for all distributed ones. Unless we have a queueing system ourself just like `WaitForPodsReady` in kueue.

keps/162-colocated-placement/README.md
```diff
@@ -0,0 +1,268 @@
...
+}
+
+type TopologyPlacementPolicy struct {
+  // +kubebuilder:default=None
+  // +kubebuilder:validation=Enum={ExclusiveTopologyPlacementPolicyType,ColocatedTopologyPlacementPolicyType,NoneTopologyPlacementPolicyType}
+  type TopologyPlacementPolicyType `json:"type"`
+  topologyKey *string `json:"topologyKey",omitempty`
```
kerthcet, line 142: Why a pointer?

keps/162-colocated-placement/README.md
```diff
@@ -0,0 +1,268 @@
...
+you're proposing, but should not include things like API designs or
+implementation. What is the desired outcome and how do we measure success?.
+The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+### User Stories (Optional)
+
+#### Story 1
+Each pod group (leader and workers) is colocated in one domain(i.e., more than one group could land on the same domain)
```
kerthcet, line 96: Can you think of a real case, this one is somehow a more theoretical one.

keps/162-colocated-placement/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+type LeaderWorkerTemplate struct {
+  // +optional
+  TopologyPlacementPolicy TopologyPlacementPolicy `json:"topologyPlacementPolicy",omitempty`
+}
+
+type TopologyPlacementPolicy struct {
+  // +kubebuilder:default=None
+  // +kubebuilder:validation=Enum={ExclusiveTopologyPlacementPolicyType,ColocatedTopologyPlacementPolicyType,NoneTopologyPlacementPolicyType}
+  type TopologyPlacementPolicyType `json:"type"`
```
kerthcet, line 141: Should we add the constraint mode like prefer or required? But this is not that vital, we can add later.

kerthcet (commented):
I think we should have a build-in feature gate system as kubernetes or it will impact our api version a lot, like this one, I have no idea whether we're on the right path because we don't have much feedbacks, without feature gates, we'll have to bump the apiversion to avoid backwards compatibilities. WDYT? @ahg-g @liurupeng

keps/162-colocated-placement/README.md
vie-serendipity: If the policy is none, topologyKey should be nil, otherwise, it should be explicitly declared. So I think pointer is more suitable.

keps/162-colocated-placement/README.md
vie-serendipity: Sounds good.

keps/162-colocated-placement/README.md
vie-serendipity: Based on my understanding, assuming a nodepool is a topology domain, each nodepool contains many nodes with the same GPU(easy to scale). So, it makes sense to schedule more than one group to a nodepool naturally. 
And it's reasonable to schedule a group of pods to a single nodepool since they have the same GPU, ensuring equal inference speeds.

liurupeng:
@vie-serendipity we are planning the release v0.5.0, do you want to proceed with this KEP? thanks!

ahg-g:
> @vie-serendipity we are planning the release v0.5.0, do you want to proceed with this KEP? thanks!

The proposal must have a way to address https://github.com/kubernetes-sigs/lws/pull/168#discussion_r1704564752

vie-serendipity:
@liurupeng I'm glad to push this work forward.

vie-serendipity:
@ahg-g I agree with kerthcet. I think it's more about scheduling and it's not the scope lws should cover. How do you think?

ahg-g:
> @ahg-g I agree with kerthcet. I think it's more about scheduling and it's not the scope lws should cover. How do you think?

But we can't offer an API that is not backed by an implementation.

vie-serendipity:
I will close this pr for now until gang scheduling makes progress in the community, such as https://github.com/kubernetes/enhancements/issues/4671.

PR #174, opened 2024-07-20
Add example for llama.cpp

justinsb:
In the first commit we just bring up llama.cpp, not _really_ using LWS.

In the second commit we really use LeaderWorkerSet, leveraging llama.cpp's RPC support


liurupeng:
thanks @justinsb for the contribution! could you add a readme for this example like this? (https://github.com/kubernetes-sigs/lws/blob/main/docs/examples/vllm/README.md)

docs/examples/llamacpp/images/llamacpp-leader/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN mkdir /workspace
+WORKDIR /workspace
+RUN git clone https://github.com/ggerganov/llama.cpp.git
+
+WORKDIR /workspace/llama.cpp
+
+# GGML_RPC=ON: Builds RPC support
+# BUILD_SHARED_LIBS=OFF: Don't rely on shared libraries like libggml
+RUN cmake . -DGGML_RPC=ON -DBUILD_SHARED_LIBS=OFF -DGGML_DEBUG=1
+RUN cmake --build . --config Release --parallel 8
```
liurupeng, line 15: is the parallel here "tensor parallelism" or "pipeline parallelism"?

docs/examples/llamacpp/README.md
```diff
@@ -0,0 +1,76 @@
+# Deploy Distributed Inference Service with llama.cpp
+
+In this example, we will use LeaderWorkerSet to deploy a distributed
+inference service using [llama.cpp](https://github.com/ggerganov/llama.cpp).    
+llama.cpp began as a project to support CPU-only inference on a single node, but has
+since expanded to support accelerators and distributed inference.
+
+l.md)
```
liurupeng, line 8: is this added accidentally?

liurupeng:
thanks @justinsb! very nice example for using CPU for multi node inference!

docs/examples/llamacpp/images/llamacpp-leader/Dockerfile
justinsb: This is just running the cmake build in parallel.  Just a slightly faster docker build, no runtime effect :-)

docs/examples/llamacpp/README.md
justinsb: Yes, I'll remove ... I'm thinking I should add support for GPUs also next, so maybe I'll do that at the same time!

Issue #173, opened 2024-07-17
Create one headless service for one replica/podgroup


liurupeng:
**What would you like to be added**:
One headless service should be created for each replica/podgroup

**Why is this needed**:
For large scale workload, there is a limit of the how many pods can be included in the single headless service. And actually we don't need communication for cross replicas

**Completion requirements**:
Potentially provide an API for configuring the headless service

This enhancement requires the following artifacts:

- [x] Design doc
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

kannon92:
/assign

Would it be okay if I take this?

I'll write up a KEP first once I get a good understanding of the code.

liurupeng:
sure, sgtm @kannon92

kannon92:
Does the leader get its own headless service also?

liurupeng:
yes,  leaders will have headless service as well

kannon92:
So the goal of this proposal would be that each group of leader-worker would have their own headless service.

This is because we want leaders to community with workers but we don't need them to communicate with other replicas.

So we have as many headless services as we have replicas?

Sorry if this obvious. I'm trying to understand the difference between Replicas and size.

PR #177, opened 2024-07-26
kep update for headless service

kannon92:
#### What type of PR is this?
/kind feature

KEP for #173 

KEP for #173.

Agreement on design.

```release-note
NONE
```

### DIFF
--- a/keps/173-headless-service-per-group/README.md
+++ b/keps/173-headless-service-per-group/README.md
@@ -1,0 +1,297 @@
+# KEP-173: Headless Service Per Group
+
+<!--
+This is the title of your KEP. Keep it short, simple, and descriptive. A good
+title can help communicate what the KEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a KEP and for
+highlighting any additional information provided beyond the standard KEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1](#story-1)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API](#api)
+  - [Implementation](#implementation)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+LeaderWorkerSet creates a single headless service for the entire object.
+This KEP is about allowing one to create a headless service per each replica.
+
+## Motivation
+
+The number of pods per service is finite. At large scale, it is possible to exhaust this limit.
+See [DNS AT Scale](https://gist.github.com/aojea/32aeaa86aacebcdd93596ecb70fcba4f) for a more indepth explanation.
+
+With large number of replicas and size, we should allow one to create a service per replica.
+
+### Goals
+
+- Each group of leader-worker replicas can have a dedicated service
+<!--
+List the specific goals of the KEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+### Non-Goals
+
+- Providing an API to configure service per replicas.
+<!--
+What is out of scope for this KEP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+## Proposal
+
+<!--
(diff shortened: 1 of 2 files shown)


kannon92:
/hold

Still need to figure out the implementation.

Main issues are how do I create the services and what to do about the pod code that sets labels.

keps/173-headless-service-per-group/README.md
```diff
@@ -0,0 +1,297 @@
...
+This section should contain enough information that the specifics of your
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+### API
+
+```golang
+type LeaderWorkerSetSpec struct {
```
liurupeng, line 130: should we allow more configurable API here? 
type ServiceConfiguration struct {
   Type ServiceConfigurationType
}
type ServiceConfigurationType string
const (
 DefaultHeadlessService ServiceConfigurationType = "DefaultHeadlessService"
 HeadlessServicePerReplica ServiceConfigurationType = "ServiceConfigurationType"
)

liurupeng:
> /hold
> 
> Still need to figure out the implementation.
> 
> Main issues are how do I create the services and what to do about the pod code that sets labels.

could leaders use one headless service, and each worker sts use one headless service. It should be fine as long as the records are created following the format?

kannon92:
> > /hold
> > Still need to figure out the implementation.
> > Main issues are how do I create the services and what to do about the pod code that sets labels.
> 
> could leaders use one headless service, and each worker sts use one headless service. It should be fine as long as the records are created following the format?

But doesn’t a leader need to communicate with the workers?

keps/173-headless-service-per-group/README.md
kannon92: Not sure. I sorta took the approach that if there isn’t a concrete use case for this, we shouldn’t support it yet. YAGNI.

ahg-g: Two suggestions: 
1) To ensure that the API is future proof, I think we should encapsulate this in a `NetworkConfig` struct. We do the same in JobSet.
2) Since the headless service is an implementation detail of how we create dns host names of the pods, I think exposing that in the API is probably not the best approach. Instead, we can have the parameter represent the dns subdomain that will be created, shared vs unique per replica, and the consequence of that is shared vs one-per-replica headless service.

```
type NetworkConfig struct {
  SubdomainPolicy SubdomainPolicy
}
type SubdomainPolicy string
const (
  SubdomainShared SubdomainPolicy = "Shared"
  SubdomainUniquePerReplica SubdomainPolicy = "UniquePerReplica"
)
```

ahg-g:
I agree with @liurupeng, creating a headless service for each sts seems like the best approach and the easies to debug by users since we can simply match the names.

ahg-g:
> > > /hold
> > > Still need to figure out the implementation.
> > > Main issues are how do I create the services and what to do about the pod code that sets labels.
> > 
> > 
> > could leaders use one headless service, and each worker sts use one headless service. It should be fine as long as the records are created following the format?
> 
> But doesn’t a leader need to communicate with the workers?

yes, but that has nothing to do with how we create the services.

keps/173-headless-service-per-group/README.md
kannon92: Are you saying that you don't really want to have a single headless service anymore in general?

ahg-g: Yes, this is generally a more scalable approach.

kannon92: What would you suggest for a transition strategy?

This would be a breaking change I think. 

We set environment variables on Pods to match the single headless service. I think that new lws on upgrades would need to change to the new service.

ahg-g: > We set environment variables on Pods to match the single headless service.

I don't think this is how it works. Were do we do that?

kannon92: https://github.com/kubernetes-sigs/lws/blob/main/pkg/utils/pod/pod_utils.go#L115

ahg-g: ok, so it turns out that the service name is actually the subdomain, we can't set a subdomain that is different from the headless service name :(

because of this, the API suggestion above in https://github.com/kubernetes-sigs/lws/pull/177#discussion_r1696170639 is I think what we can go with

kannon92: Do you want to default to UniquePerReplica from 0.4.0 on?

ahg-g: Unfortunately we can't, the default must be what we have right now.

We need to think of whether we want to make this an immutable field, and if so how we will handle such an update.

liurupeng:
@kannon92 could you update the implementation part, thanks!

keps/173-headless-service-per-group/README.md
ahg-g: Can you please update the API as suggested in https://github.com/kubernetes-sigs/lws/pull/177/files#r1696170639? I think it better convoy what the user facing impact will be when it is used.

keps/173-headless-service-per-group/README.md
```diff
@@ -0,0 +1,297 @@
...
+}
+```
+
+### Implementation
+
+With HeadlessServicePerReplica true, we will create a headless service per replica.
+The name of the service will be `{lws.Name}-{replica-index}`.
+Creation of the services will be in a loop of replicas.
+
+Currently, LWS will add an environment variable to all pods that points to the headless service.
```
ahg-g, line 145: it is not to the headless service, it is to the dns record. Depending on how the headless service is created, the hostname will be different, and we should clarify that.

kannon92:
@liurupeng has asked to take this feature over. I got pulled into other things so I won't be able to focus on this anytime soon.

/close

PR #188, opened 2024-08-06
Headless Service Per Replica KEP

Edwinhr716:
#### What type of PR is this?
/kind feature

KEP for #173 . 

Based on original KEP #177. Added suggestions, and modified the way LWS_LEADER_ADDRESS is set.

Fixes #

```release-note

```


keps/173-headless-service-per-replica/README.md
```diff
@@ -0,0 +1,328 @@
...
+  SubdomainShared SubdomainPolicy = "Shared"
+  // SubdomainUniquePerReplica will create a headless service for each
+  // leader-worker group
+  SubdomainUniquePerReplica SubdomainPolicy = "UniquePerReplica"
+)
+```
+
+### Implementation
+
+With HeadlessServicePerReplica true, we will create a headless service per replica.
```
liurupeng, line 152: maybe we create a headless service per statefulset

keps/173-headless-service-per-replica/README.md
```diff
@@ -0,0 +1,328 @@
...
+The name of the service will be `{lws.Name}-{replicaIndex}`.
+Creation of the services will be in a loop of replicas. The selector of each
```
liurupeng, line 154: the last sentence hasn't been completed

keps/173-headless-service-per-replica/README.md
```diff
@@ -0,0 +1,328 @@
...
+    },
+    Spec: corev1.ServiceSpec{
+      ClusterIP: "None", // defines service as headless
+      Selector: map[string]string{
+        leaderworkerset.GroupIndexLabelKey: i,
+      },
+      PublishNotReadyAddresses: true,
+    },
+  }
+}
```
liurupeng, line 175: how to assign the generated pods to these headless service?

keps/173-headless-service-per-replica/README.md
```diff
@@ -0,0 +1,328 @@
...
+```
+
+With this feature enabled, we will need to set the headless service of the replica the pod belongs to.
+```golang
+ leaderAddressEnvVar := corev1.EnvVar{
+  Name:  leaderworkerset.LwsLeaderAddress,
+  Value: fmt.Sprintf("%s-%s.%s-%s.%s", lwsName, groupIndex, lwsName, groupIndex, pod.ObjectMeta.Namespace),
+ }
+```
+
```
liurupeng, line 197: we may need to add implementation details for validation/update webhook logic. Also do we allow mutating this field etc?

keps/173-headless-service-per-replica/README.md
Edwinhr716: So the leader will have one, while the workers will have a different one?

keps/173-headless-service-per-replica/README.md
Edwinhr716: removed sentence

keps/173-headless-service-per-replica/README.md
Edwinhr716: It uses the GroupIndexLabelKey as selector, so that way all pods that are part of the same replica are selected to the same headless service

keps/173-headless-service-per-replica/README.md
liurupeng: yes, ideally we only need the headless service to generate the dns record

keps/173-headless-service-per-replica/README.md
liurupeng: we need to add the details how pods will be selected by which headless service as well

keps/173-headless-service-per-replica/README.md
Edwinhr716: What are the benefits of creating a new one per statefulset instead? Creating a new one per sts creates more headless services

ahg-g: The headless service name will be part of the fully qualified host name of the pod, so how many and what name we give the service matters. It is also why I am suggesting that the API to be a knob that decides on the subdomain policy because this is the user facing side effect. 

The pod hostname is `<pod-name>.<svc-name>`

So if the new "policy" is `UniquePerReplica`, then each replica including the leader should be part of the same headless service so that the middle part above is common among the pods of a replica. Example:

Replica 0:
- my-lws-0.my-lws-0
- my-lws-0-1.my-lws-0

Replica 1:
- my-lws-1.my-lws-1
- my-lws-1-1.my-lws-1


If we create a headless service per sts including a special one for leaders, then we can't achieve the above since the leader will have a different subdomain name. Here is what it will look like:

Replica 0:
- my-lws-0.my-lws
- my-lws-0-1.my-lws-0

Replica 1:
- my-lws-1.my-lws
- my-lws-1-1.my-lws-1

I am fine with the above, but what would we call such a domain name "policy"? it wouldn't be `UniquePerReplica`.

keps/173-headless-service-per-replica/README.md
ahg-g: +1, if we want to support updates, that would be great, but we need to clearly discuss how we will do that without disruptions.

keps/173-headless-service-per-replica/README.md
Edwinhr716: I prefer the `uniquePerReplica` option. It creates less headless services, and it makes sense for the leader and worker to have the subdomain in common. 

Plus, implementing a headless service per sts will require adding a label or annotation in order for the headless service to only select the workers (we can use `worker-index=0` to only select the leaders)

Edwinhr716: Thinking about it a bit more, creating one exclusively for the leaders would allow for easier transition, since it doesn't require a change to the LWS_LEADER_ADDRESS environment variable. The leader address would still be `fmt.Sprintf("%s-%s.%s.%s", lwsName, groupIndex, lwsName, pod.ObjectMeta.Namespace)` https://github.com/kubernetes-sigs/lws/blob/main/pkg/utils/pod/pod_utils.go#L119

ahg-g: I am just struggling to suggest a name that clearly indicates the semantics, what would we call this policy?

Edwinhr716:
Changed implementation to creating a headless service per sts. Still have to come up with a name for the new subdomainPolicy.
 @ahg-g @liurupeng PTAL

keps/173-headless-service-per-replica/README.md
```diff
@@ -0,0 +1,328 @@
...
+ // SubdomainPolicy determines the policy that will be used when creating
+ // the headless service
+ SubdomainPolicy SubdomainPolicy `json:"subdomainPolicy,omitempty"`
+}
+
+type NetworkConfig struct {
+  SubdomainPolicy SubdomainPolicy
+}
+type SubdomainPolicy string
+const (
```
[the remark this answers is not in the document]
Edwinhr716: added

keps/173-headless-service-per-replica/README.md
Edwinhr716: Added details for update logic. After testing, I think it is doable to allow mutating the field.

keps/173-headless-service-per-replica/README.md
```diff
@@ -0,0 +1,359 @@
...
+  // SubdomainShared will create a single headless service that all replicas 
+  // will share. The host names look like:
+  // Replica 0: my-lws-0.my-lws, my-lws-0-1.my-lws
+  // Replica 1: my-lws-1.my-lws, my-lws-1-1.my-lws
+  SubdomainShared SubdomainPolicy = "Shared"
+  // SubdomainUniquePerReplica will create a headless service for each
+  // leader-worker group. 
+  // The leader host names will look like:
+  // Replica 0: my-lws-0.my-lws
```
kerthcet, line 145: IIUC, if `uniquePerReplica`, which means the leader pod and the worker Pods should have the same svc name, but here they're different. `my-lws-0.my-lws` vs `my-lws-0-1.my-lws-0`

keps/173-headless-service-per-replica/README.md
kerthcet: > I prefer the uniquePerReplica option. It makes sense for the leader and worker to have the subdomain in common.

 +1, or it's weird I think, unless we have other use cases like visiting the workers only.

keps/173-headless-service-per-replica/README.md
```diff
@@ -0,0 +1,359 @@
...
+)
+```
+
+### Implementation
+
+With SubdomainPolicy set to SubdomainUniquePerReplica, we will create a headless
+service for all of the leaders, and a headless service per replica for the workers. In order 
+to ensure backwards compatability, the default value if non is set will be Shared.
+
+A label will be added to determine whether a pod is a leader or a worker.
```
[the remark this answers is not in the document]
Edwinhr716: We can use `workerIndex = 0` to select the leaders, but cannot do `workerIndex != 0` to select the workers since AFAIK selector does not have that capability.

keps/173-headless-service-per-replica/README.md
Edwinhr716: The implementation is for headless service per sts, still figuring out a naming scheme so left the old name to discuss it further

keps/173-headless-service-per-replica/README.md
```diff
@@ -0,0 +1,359 @@
...
+proposal will be implemented, this is the place to discuss them.
+-->
+
+### API
+
+```golang
+type LeaderWorkerSetSpec struct {
+ // SubdomainPolicy determines the policy that will be used when creating
+ // the headless service
+ SubdomainPolicy SubdomainPolicy `json:"subdomainPolicy,omitempty"`
```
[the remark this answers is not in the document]
Edwinhr716: Pointer, fixed

keps/173-headless-service-per-replica/README.md
```diff
@@ -0,0 +1,359 @@
...
+}
+
+type NetworkConfig struct {
```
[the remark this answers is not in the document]
Edwinhr716: the `NetworkConfig` is the overarching type, fixed

keps/173-headless-service-per-replica/README.md
ahg-g: But again, what would you call that policy? it is certainly not `UniquePerReplica` if we do a headless service per sts.

Edwinhr716: What about `uniquePerStatefulSet`?

ahg-g: While statefulset is known as the implementation API, it is not a "concept" in the LWS API. The concepts we have are: replica/group, leader, worker etc.

Edwinhr716: What about `uniquePerLeaderAndWorkerReplicas`? It is quite long but I think gets the idea across.

ahg-g: `UniquePerReplicaWorkers` maybe?

kerthcet: What about `leaderSharedWorkerDedicated` vs `shared`.

Edwinhr716: Of the three I prefer `leaderSharedWorkerDedicated`

keps/173-headless-service-per-replica/README.md
```diff
@@ -356,4 +356,9 @@ Why should this KEP _not_ be implemented?
 What other approaches did you consider, and why did you rule them out? These do
 not need to be as detailed as the proposal, but should include enough
 information to express the idea and why it was not acceptable.
--->
\ No newline at end of file
+-->
+
+Instead of doing a headless service exclusively for the leaders, and a headless service per replica por the workers, we
```
ahg-g, line 361:
```suggestion
Instead of doing a headless service exclusively for the leaders, and a headless service per replica for the workers, we 
```

keps/173-headless-service-per-replica/README.md
ahg-g: `LeadersSharedWorkersDedicated`

keps/173-headless-service-per-replica/README.md
Edwinhr716: fixed

PR #197, opened 2024-08-16
Adding SubdomainPolicy to support a service per replica

### LINKED ISSUE #173: Create one headless service for one replica/podgroup

Edwinhr716:
#### What type of PR is this?
/kind feature

Details can be found in KEP #188 

Fixes #173 

```release-note
Adds NetworkConfig and SubdomainPolicy to the API
```


pkg/controllers/leaderworkerset_controller.go
```diff
@@ -122,27 +123,53 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	return ctrl.Result{}, nil
 }
 
-func (r *LeaderWorkerSetReconciler) createHeadlessServiceIfNotExists(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet) error {
+func (r *LeaderWorkerSetReconciler) createMultipleHeadlessServices(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet) error {
+	if lws.Spec.NetworkConfig.SubdomainPolicy == leaderworkerset.SubdomainShared {
```
kannon92, line 127: Maybe you should check the nil case also. On newly created replica sets you can assume it’s defaulted. But you may need to support LWS that are coming from a case where this is not yet defaulted. In that case I think you’d want to fall back to the original case?

Edwinhr716: Makes sense to me, I'll add the nil check. That does make me question if defaulting on the lws webhook is needed since the nil check does the same thing however.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -122,27 +123,53 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+		}
+		for i := 0; i < int(*lws.Spec.Replicas); i++ {
+			if err := r.deleteHeadlessServiceIfExists(ctx, lws, fmt.Sprintf("%s-%s", lws.Name, strconv.Itoa(i))); err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+
+	err := r.createHeadlessServiceIfNotExists(ctx, lws, lws.Name, map[string]string{leaderworkerset.PodRoleLabelKey: "leader"})
```
liurupeng, line 139: There is one case when you use "SubdomainShared" at first and then switched to "SubdomainLeadersSharedWorkersDedicated", the service of the leader will not be updated right? since the servicename didn't change. But actually we need to trigger an update to the service, since the selector of the leader's headless service has changed?

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -122,27 +123,53 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	}
+	return nil
+}
+
+func (r *LeaderWorkerSetReconciler) createHeadlessServiceIfNotExists(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, serviceName string, serviceSelector map[string]string) error {
 	log := ctrl.LoggerFrom(ctx)
 	// If the headless service does not exist in the namespace, create it.
 	var headlessService corev1.Service
-	if err := r.Get(ctx, types.NamespacedName{Name: lws.Name, Namespace: lws.Namespace}, &headlessService); err != nil {
+	if err := r.Get(ctx, types.NamespacedName{Name: serviceName, Namespace: lws.Namespace}, &headlessService); err != nil {
```
liurupeng, line 157: the approach here won't figure out that the leader headless service's selector could be changed?

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -153,6 +180,30 @@ func (r *LeaderWorkerSetReconciler) createHeadlessServiceIfNotExists(ctx context
 			return err
 		}
 	}
+
+	// updating the headless service as it transitions between Shared and LeadersSharedWorkersDedicated
+	if eq := reflect.DeepEqual(headlessService.Spec.Selector, serviceSelector); !eq {
+		headlessService.ObjectMeta.Name = serviceName
```
liurupeng, line 186: why we set the headless service name, should it already be serviceName?

test/testutils/validators.go
```diff
@@ -55,31 +55,67 @@ func ExpectLeaderSetExist(ctx context.Context, lws *leaderworkerset.LeaderWorker
...
+			}
 		}
 		return true, nil
 	}, Timeout, Interval).Should(gomega.Equal(true))
 }
 
+func validateService(headlessService corev1.Service, lws *leaderworkerset.LeaderWorkerSet, serviceName string, wantSelector map[string]string) (bool, error) {
+	if headlessService.ObjectMeta.Name != serviceName {
+		return false, errors.New("service name mismatch")
+	}
```
liurupeng, line 99: consider use this approach to print out the diff for easier debugging? https://github.com/kubernetes-sigs/lws/blob/93bfa66aa718c05ba4a4dcd2082df9e05b480b2e/test/testutils/validators.go#L138C3-L143C19

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -176,6 +183,31 @@ type SubGroupPolicy struct {
...
+	// Replica 0: my-lws-0.my-lws, my-lws-0-1.my-lws
+	// Replica 1: my-lws-1.my-lws, my-lws-1-1.my-lws
+	SubdomainShared SubdomainPolicy = "Shared"
+	// SubdomainLeadersSharedWorkersDedicated will create a headless service for the leaders,
+	// and a headless service per replica for the workers.
+	// The leader host names will look like:
+	// Replica 0: my-lws-0.my-lws
+	// Replica 1: my-lws-1.my-lws
+	// The worker host names will look like:
+	// Replica 0: my-lws-0-1.my-lws-0, my-lws-0-2.my-lws-0
```
liurupeng, line 206: if we switched from "Shared" to "LeadersSharedWorkersDedicated", the worker pods' hostnames will change  right? for tpu case, do we need to update the container env vars as well? and do we need to trigger a rolling update?

liurupeng:
Would be great if you could 
1. Add one test to verify that after headless service mode update, for tpu case, we could see container env vars being updated in integration test
2. Add a e2e test to verify the creation of each headless service mode and an update 
3. Manually test the multi-host TPU solution for not configuration the headless service mode and switch to the "LeaderSharedWorkerDedicated" case

api/leaderworkerset/v1/leaderworkerset_types.go
Edwinhr716: That makes sense, they will need to be updated. In the case where there is an update from "Shared" to "LeadersSharedWorkersDedicated", and one of the TPUs disconnects but doesn't cause a restart, when the TPU attempts to connect again the `TPU_WORKER_HOSTNAMES` env variable will have the wrong value, causing it to fail. We can trigger a rolling update, which in the default case has `MaxSurge=0` and `MaxUnavailable=1`, so updating it won't cause much disruption.

Edwinhr716:
1) I think to test that scenario, an e2e is better since the env variables change due to the `Pod.Spec.Subdomain` value changing, which in a integration-test for the pod-webhook we'll have to manually change that value
2) I can do an e2e test that covers both 1 & 2

test/e2e/e2e_test.go
```diff
@@ -201,6 +202,51 @@ var _ = ginkgo.Describe("leaderWorkerSet e2e tests", func() {
...
+
+		// Rolling update completes.
+		testing.ExpectValidLeaderStatefulSet(ctx, k8sClient, lws, 4)
+		testing.ExpectValidWorkerStatefulSets(ctx, lws, k8sClient, true)
+		testing.ExpectValidServices(ctx, k8sClient, lws, 4)
+		testing.ExpectValidPods(ctx, k8sClient, lws, &corev1.PodList{})
+		// Wait for leaderWorkerSet to be ready again.
+		testing.ExpectLeaderWorkerSetAvailable(ctx, k8sClient, lws, "All replicas are ready")
+	})
+
```
liurupeng, line 249: could we add an e2e test from updating UniquePerReplica to Shared and verify the container env is updated?

test/testutils/validators.go
```diff
@@ -52,34 +52,65 @@ func ExpectLeaderSetExist(ctx context.Context, lws *leaderworkerset.LeaderWorker
...
+
+		var headlessService corev1.Service
+		var headlessServiceList corev1.ServiceList
+		if err := k8sClient.List(ctx, &headlessServiceList, client.InNamespace(lws.Namespace)); err != nil {
+			return false, err
 		}
-		if headlessService.Spec.ClusterIP != "None" {
-			return false, errors.New("service type mismatch")
+
+		if len(headlessServiceList.Items) != (numHeadlessServices) {
```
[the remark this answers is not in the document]
Edwinhr716: I don't think so, or at least not consistently. It can work after updates are done (check 1 if shared or check lws.spec.replicas if not) but during rollingUpdate with MaxSurge the number of services won't have either value and there's no way to identify that unless we do if rollingUpdateInProgress then lws.spec.replicas+maxSurge+1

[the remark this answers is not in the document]
liurupeng: let's keep it as this then

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -122,27 +122,60 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	return ctrl.Result{}, nil
 }
 
-func (r *LeaderWorkerSetReconciler) createHeadlessServiceIfNotExists(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet) error {
+func (r *LeaderWorkerSetReconciler) createMultipleHeadlessServices(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, replicas int32) error {
+	if lws.Spec.NetworkConfig.SubdomainPolicy == leaderworkerset.SubdomainShared || lws.Spec.NetworkConfig == nil {
```
liurupeng, line 126: switch the order here?

Edwinhr716: Fixed

test/e2e/e2e_test.go
Edwinhr716: Yep, added from UniquePerReplica to Shared

liurupeng:
overall lgtm, @ahg-g could help check as well? thanks!

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -122,27 +122,61 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+
+	if lws.Spec.NetworkConfig == nil || lws.Spec.NetworkConfig.SubdomainPolicy == leaderworkerset.SubdomainShared {
+		if err := r.createHeadlessServiceIfNotExists(ctx, lws, lws.Name, map[string]string{leaderworkerset.SetNameLabelKey: lws.Name}); err != nil {
+			return err
+		}
+		return nil
+	}
+	// using replicas instead of lws.spec.replicas as needed to create services for burst replicas during maxSurge
+	for i := 0; i < int(replicas); i++ {
+		err := r.createHeadlessServiceIfNotExists(ctx, lws, fmt.Sprintf("%s-%s", lws.Name, strconv.Itoa(i)), map[string]string{leaderworkerset.GroupIndexLabelKey: strconv.Itoa(i)})
```
ahg-g, line 135: This is not enough, we need to also have the lws.Name as well (an AND); otherwise the headless service will target every replica of this index across all LWSs in the cluster.

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -67,6 +67,11 @@ func (r *LeaderWorkerSetWebhook) Default(ctx context.Context, obj runtime.Object
 			MaxSurge:       intstr.FromInt32(0),
 		}
 	}
+
+	if lws.Spec.NetworkConfig == nil {
```
ahg-g, line 71: we need to default `lws.Spec.NetworkConfig.SubdomainPolicy` in case networkconfig was set but not the policy. `SubdomainPolicy` should be a pointer as mentioned in the api file.

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -180,6 +188,27 @@ type SubGroupPolicy struct {
 	SubGroupSize *int32 `json:"subGroupSize,omitempty"`
 }
 
+type NetworkConfig struct {
+	// SubdomainPolicy determines the policy that will be used when creating
+	// the headless service
+	SubdomainPolicy SubdomainPolicy `json:"subdomainPolicy"`
```
ahg-g, line 194: should be a pointer as well and mention that the default is Shared

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -97,6 +102,7 @@ func (r *LeaderWorkerSetWebhook) ValidateUpdate(ctx context.Context, oldObj, new
 	if newLws.Spec.LeaderWorkerTemplate.SubGroupPolicy == nil && oldLws.Spec.LeaderWorkerTemplate.SubGroupPolicy != nil {
 		allErrs = append(allErrs, field.Invalid(specPath.Child("leaderWorkerTemplate", "SubGroupPolicy", "subGroupSize"), oldLws.Spec.LeaderWorkerTemplate.SubGroupPolicy.SubGroupSize, "cannot remove subGroupSize after enabled"))
 	}
+
```
ahg-g, line 105: we should validate that lws.Spec.NetworkConfig.SubdomainPolicy can not be updated to nil

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -122,27 +122,61 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
-					leaderworkerset.SetNameLabelKey: lws.Name,
-				},
+				ClusterIP:                "None", // defines service as headless
+				Selector:                 serviceSelector,
 				PublishNotReadyAddresses: true,
 			},
 		}
+
 		// Set the controller owner reference for garbage collection and reconciliation.
 		if err := ctrl.SetControllerReference(lws, &headlessService, r.Scheme); err != nil {
```
ahg-g, line 181: To manage deletions of headless services in an easier why, set the owner reference to the worker sts for the uniquePerReplica case. With this, you don't need to delete them explicitly because they will automatically be deleted when the worker sts is deleted. 

I would also be ok to not delete the headless service in the case we are switching from shared to uniquePerReplica, just leave a comment to that effect. This is because it makes the transition safer, and this headless service will be deleted eventually when the lws gets deleted anyway, so there is no resource leak.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -122,27 +122,61 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	}
+	// using replicas instead of lws.spec.replicas as needed to create services for burst replicas during maxSurge
+	for i := 0; i < int(replicas); i++ {
+		err := r.createHeadlessServiceIfNotExists(ctx, lws, fmt.Sprintf("%s-%s", lws.Name, strconv.Itoa(i)), map[string]string{leaderworkerset.GroupIndexLabelKey: strconv.Itoa(i)})
+		if err != nil {
+			return err
+		}
+	}
+
+	var headlessServiceList corev1.ServiceList
```
ahg-g, line 141: we don't need to explicitly delete, see the comment below.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -262,6 +340,9 @@ func (r *LeaderWorkerSetReconciler) rollingUpdateParameters(ctx context.Context,
 	// Case 3:
 	// In normal cases, return the values directly.
 	if rollingUpdateCompleted {
+		if err := r.deleteMultipleHeadlessSevices(ctx, lws); err != nil {
```
ahg-g, line 343: I don't think we will need this if we set the ownerReference on the headless services correctly

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -582,6 +663,10 @@ func constructLeaderStatefulSetApplyConfiguration(lws *leaderworkerset.LeaderWor
 			podAnnotations[leaderworkerset.SubGroupExclusiveKeyAnnotationKey] = lws.Annotations[leaderworkerset.SubGroupExclusiveKeyAnnotationKey]
 		}
 	}
+
+	if lws.Spec.NetworkConfig != nil && lws.Spec.NetworkConfig.SubdomainPolicy == leaderworkerset.SubdomainUniquePerReplica {
```
ahg-g, line 667: you will also not need this annotation if we do deletions via ownerReference

ahg-g (commented):
Didn't review the tests yet.

I think we can simplify the implementation, see comments.

pkg/controllers/leaderworkerset_controller.go
liurupeng: oh, right, this is a good point

pkg/controllers/leaderworkerset_controller.go
Edwinhr716: The annotation is there to determine what domain to inject on the leader. Still need it even setting the worker stateful set as the owner.

pkg/webhooks/leaderworkerset_webhook.go
Edwinhr716: If we default `subdomainPolicy` in the case where `networkConfig` is set but not `subdomainPolicy`, doesn't that cover the case where it is updated to nil? So it is valid to update it to nil, it would just be equivalent to setting it to `Shared`.

pkg/controllers/leaderworkerset_controller.go
Edwinhr716: Done

pkg/controllers/leaderworkerset_controller.go
Edwinhr716: fixed

pkg/controllers/leaderworkerset_controller.go
Edwinhr716: Changed

api/leaderworkerset/v1/leaderworkerset_types.go
Edwinhr716: Changed

pkg/controllers/leaderworkerset_controller.go
Edwinhr716: fixed

pkg/webhooks/leaderworkerset_webhook.go
Edwinhr716: fixed

test/testutils/validators.go
```diff
@@ -52,34 +52,68 @@ func ExpectLeaderSetExist(ctx context.Context, lws *leaderworkerset.LeaderWorker
 	}, Timeout, Interval).Should(gomega.Equal(true))
 }
 
-func ExpectValidServices(ctx context.Context, k8sClient client.Client, lws *leaderworkerset.LeaderWorkerSet) {
+func ExpectValidServices(ctx context.Context, k8sClient client.Client, leaderWorkerSet *leaderworkerset.LeaderWorkerSet, numHeadlessServices int) {
 	gomega.Eventually(func() (bool, error) {
-		var headlessService corev1.Service
-		if err := k8sClient.Get(ctx, types.NamespacedName{Name: lws.Name, Namespace: lws.Namespace}, &headlessService); err != nil {
+
+		// Always got the latest lws.
```
ahg-g, line 58:
```suggestion
		// Always get the latest lws. 
```

test/testutils/validators.go
```diff
@@ -52,34 +52,68 @@ func ExpectLeaderSetExist(ctx context.Context, lws *leaderworkerset.LeaderWorker
...
 	}, Timeout, Interval).Should(gomega.Equal(true))
 }
 
+func validateService(headlessService corev1.Service, lws *leaderworkerset.LeaderWorkerSet, serviceName string, wantSelector map[string]string) (bool, error) {
+	if headlessService.ObjectMeta.Name != serviceName {
+		return false, errors.New("service name mismatch")
+	}
+	if headlessService.ObjectMeta.Namespace != lws.Namespace {
+		return false, errors.New("service namespace mismatch")
+	}
```
ahg-g, line 99: Checking the name and namespace is redundent, we already used them to get the object, can we please remove them.

test/e2e/e2e_test.go
```diff
@@ -201,6 +202,59 @@ var _ = ginkgo.Describe("leaderWorkerSet e2e tests", func() {
...
+		lws := testing.BuildLeaderWorkerSet(ns.Name).Replica(1).Size(2).LeaderTemplateSpec(leaderPodSpec).WorkerTemplateSpec(workerPodSpec).Obj()
+		testing.MustCreateLws(ctx, k8sClient, lws)
+		testing.ExpectValidPods(ctx, k8sClient, lws, &corev1.PodList{})
+		testing.UpdateSubdomainPolicy(ctx, k8sClient, lws, leaderworkerset.SubdomainUniquePerReplica)
+		lwsPods := &corev1.PodList{}
+		testing.ExpectValidPods(ctx, k8sClient, lws, lwsPods)
+
+		for _, pod := range lwsPods.Items {
+			gomega.Expect(testing.HasTPUEnvVarsPopulated(pod)).To(gomega.BeTrue())
+			if pod.Labels[leaderworkerset.GroupIndexLabelKey] == "0" {
```
ahg-g, line 217: The GroupIndex refers to replica index, not the pod index within a replica. Since you have one replica, this will be 0 for both pods, and so I am confused how come the test is passing because the check in the else should fail, what am I missing?

Edwinhr716: It never reaches the else statement, so it never fails that portion of the test. I'll update the number of replicas to two to verify that the subdomain name is different on both replicas

test/testutils/validators.go
Edwinhr716: removed

pkg/controllers/pod_controller.go
```diff
@@ -231,6 +237,42 @@ func (r *PodReconciler) topologyValueFromPod(ctx context.Context, pod *corev1.Po
 	return topology, nil
 }
 
+func (r *PodReconciler) createHeadlessServiceIfNotExists(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, pod *corev1.Pod) error {
```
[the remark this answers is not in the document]
Edwinhr716: Combined into one

PR #207, opened 2024-09-03
Bump kubernetes dependencies to v0.31.x.

mbobrovskyi:
#### What type of PR is this?
/kind cleanup

Bump kubernetes dependencies to v0.31.x.

Fixes #

```release-note
NONE
```


.golangci.yaml
```diff
@@ -1,10 +1,10 @@
...
 # see: https://golangci-lint.run/usage/configuration/
 
-# Options for analysis running
-run:
-  # Which dirs to skip: they won't be analyzed;
-  skip-dirs:
+# Settings related to issues
+issues:
+  # Which dirs to exclude: issues from them won't be reported
+  exclude-dirs:
```
mbobrovskyi, line 7: This changes fix warning:

```
WARN [config_reader] The configuration option `run.skip-dirs` is deprecated, please use `issues.exclude-dirs`.
```

Makefile-deps.mk
```diff
@@ -21,6 +21,9 @@ else
 endif
 GO_CMD ?= go
 
+GOLANGCI_LINT = $(PROJECT_DIR)/bin/golangci-lint
+golangci-lint:
+	@GOBIN=$(PROJECT_DIR)/bin GO111MODULE=on $(GO_CMD) install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.60.3
```
mbobrovskyi, line 26: These changes are necessary because I encountered an error with the previous version:

```
ERRO [linters_context/goanalysis] SA5012: panic during analysis: interface conversion: interface {} is nil, not *buildir.IR, goroutine 8413 [running]:
runtime/debug.Stack()
```

kannon92:
Thank you for adding this. We probably should make sure we have a 1.31 test also. Its not quite related but I don't think we have one for this yet.

kannon92:
Added https://github.com/kubernetes/test-infra/pull/33413

PR #209, opened 2024-09-06
Injecting TPU_NAME environment variable

Edwinhr716:
#### What type of PR is this?
/kind feature

When requesting TPU resources, it injects the environment variable TPU_NAMES, which contains the name of the pod the TPU is in. This is needed as popular distributed runtime libraries such as Ray expect this variable to be set https://github.com/ray-project/ray/blob/ab27fe805c3fef4c8dc66bf5804f32cd522d1aad/python/ray/_private/accelerators/tpu.py#L247

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/utils/accelerators/tpu.go
+++ b/pkg/utils/accelerators/tpu.go
@@ -154,6 +154,10 @@ func addTPUVariablesSubGroup(pod *corev1.Pod) error {
 			Name:  TpuWorkerId,
 			Value: fmt.Sprint(tpuWorkerId),
 		},
+		corev1.EnvVar{
+			Name:  "TPU_NAME",
+			Value: fmt.Sprint(pod.Name),
+		},
 	)
 	return nil
 
--- a/pkg/utils/accelerators/tpu.go
+++ b/pkg/utils/accelerators/tpu.go
@@ -32,6 +32,7 @@ const (
 	TpuResourceName                 corev1.ResourceName = corev1.ResourceName("google.com/tpu")
 	TpuWorkerHostNames              string              = "TPU_WORKER_HOSTNAMES"
 	TpuWorkerId                     string              = "TPU_WORKER_ID"
+	TpuName                         string              = "TPU_NAME"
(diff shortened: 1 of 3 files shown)


pkg/utils/accelerators/tpu.go
```diff
@@ -211,6 +216,10 @@ func AddTPUVariables(pod *corev1.Pod, size int) error {
 			Name:  TpuWorkerId,
 			Value: fmt.Sprint(tpuWorkerId),
 		},
+		corev1.EnvVar{
+			Name:  TpuName,
+			Value: fmt.Sprint(pod.Name),
```
liurupeng, line 221: I think this value is used by Ray to know which TPU host belongs to which slice? So if it's a multi-host slice, the value that be the same.  So one replica has one name, let's use leader pod's name here

liurupeng, line 221: the tpu_name here should be the identifier for a tpu slice. So for multi-host tpu slice, all hosts share the same tpu_name value. Let's use leader pod's name here.

liurupeng: actually this may cause the ray only to use one host to load the model which OOM happens on that host?

BenTheElder:
/override pull-lws-test-e2e-main-1-27

see: https://kubernetes.slack.com/archives/C7J9RP96G/p1725908686722759 / https://kubernetes.slack.com/archives/C09QZ4DQB/p1725891002485159

BenTheElder:
er, that command has to be  done by a github admin of this repo

ahg-g:
btw, we can do this today as follows: https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/

pkg/utils/accelerators/tpu.go
Edwinhr716: Updated to use leader pod's name

BenTheElder:
maybe try closing and re-opening, that forces github to clear the statuses

Issue #213, opened 2024-09-12
feature: add scalePolicy in lws


googs1025:
**What would you like to be added**:
I want to implement a function similar to scalePolicy, the api is as follows. 
```yaml
apiVersion: leaderworkerset.x-k8s.io/v1
kind: LeaderWorkerSet
metadata:
  name: leaderworkerset-sample
spec:
  scalePolicy:
    minReplicas: 1
    maxReplicas: 3
    metrics:
      - type: Resource
        resource:
          name: cpu
          target:
            type: Utilization
            averageUtilization: 80
  replicas: 3
  leaderWorkerTemplate:
    size: 4
    workerTemplate:
      spec:
        containers:
        - name: nginx
          image: nginx:1.14.2
          resources:
            limits:
              cpu: "100m"
            requests:
              cpu: "50m"
          ports:
          - containerPort: 8080
```
We can implement the scale subresource (it seems to have been implemented now) and embed HPA, so that users can use scalePolicy directly without using HPA separately.

This would involve an api-change, I'll contribute a KEP if needed.

**Why is this needed**:
- users can use scalePolicy directly without using HPA separately.

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

googs1025:
@ahg-g @liurupeng @kerthcet Any ideas or suggestions?

kerthcet:
But how to scale, aren't we also depend on the HPA implementation. If so, I don't think this is fit with lws as an plain workload. Some higher orchestration tools may need this for easy configuration, if you're interest, you can take a look at https://github.com/InftyAI/llmaz/issues/3, we're aiming for the simplified configuration.

ahg-g:
Right, why do we need this if we have HPA?

googs1025:
thanks！

My thought is to integrate HPA through the API in LWS (just like we use a headless service), so that users wouldn't need to configure HPA separately and could manage it directly from within LWS.

googs1025:
Thank you all for your suggestions. I will close this issue.：）

PR #219, opened 2024-09-14
docs: add kep-213 scale policy

googs1025:
#### What type of PR is this?
/kind documentation

- add scale policy api change kep docs

Fixes # part of https://github.com/kubernetes-sigs/lws/issues/213

```release-note
None
```

### DIFF
--- a/keps/213-scale-policy/README.md
+++ b/keps/213-scale-policy/README.md
@@ -1,0 +1,386 @@
+# KEP-213: Scale policy API
+
+<!--
+This is the title of your KEP. Keep it short, simple, and descriptive. A good
+title can help communicate what the KEP is and should be considered as part of
+any review.
+-->
+
+<!--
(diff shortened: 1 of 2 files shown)


googs1025:
I had a chat with @kerthcet  offline, and we disagreed on whether we really need to roll out this feature. Still, I'm happy to put together a simple KEP (please excuse my ignorance) for everyone to discuss the design. Any suggestions are more than welcome from all. If the consensus later is that this feature isn't necessary, I'll close this PR.

liurupeng:
This API is similar to HPA's. The user could configure this in an HPA object. So the proposal is mostly for simplifying the user experience right? We could add this API if this could greatly simplified the user workflow, otherwise we need to be cautious when adding new API. Also, in the long term, would we consider allowing the number of pods in each replica to be autoscaled? (lws.spec.leaderworkertemplate.size). Although we need to think about how to create two HPA objects for size&replica and a scale endpoint for each.

PR #250, opened 2024-11-07
change to not use NetworkConfig in hash calculation if it has default value

### LINKED ISSUE #240: When rollout new controller version, it should not trigger a rolling update
**What happened**:
In the rolling update triggering logic, we should store the template somewhere else, and only compare the fields with the original one, and check if these fields are updated, then decides to trigger a rolling update. This will mitigate the issue when we added a API field, it will not trigger rolling update when rolling out new controller version

**Environment**:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Edwinhr716:
#### What type of PR is this?
/kind bug

When calculating the template hash, if the NetworkConfig has the defaulted value, it will not be used. This is to prevent a rolling update to be triggered when upgrading from a previous version.

Fixes #240 

```release-note

```

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -41,7 +41,7 @@ func NonZeroValue(value int32) int32 {
 }
 
 func LeaderWorkerTemplateHash(lws *leaderworkerset.LeaderWorkerSet) string {
-	if lws.Spec.NetworkConfig == nil {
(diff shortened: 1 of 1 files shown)


ahg-g:
Did you test that this actually addresses the issue?

Edwinhr716:
Yes. Tested upgrading from 0.2.0 to 0.4.1, no rolling update. Afterward I made a change that doesn't trigger rolling update, NetworkConfig did get defaulted, but it doesn't trigger rolling update either.

ahg-g:
/lgtm
/approve

Thanks, can you please submit a patch?

Issue #238, opened 2024-10-23
Replicas restart during rolling upgrade will not use the original worker pod spec


liurupeng:
**What happened**:
When lws is in rolling update, any replicas that are not updated yet, but restarted due to failure handling, the worker statefulset will be recreated with updated lws worker template. The expected behavior should be recreated with the previous worker template
**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:

- Kubernetes version (use `kubectl version`):
- LWS version (use `git describe --tags --dirty --always`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

ahg-g:
The solution is https://github.com/kubernetes-sigs/lws/issues/239

PR #261, opened 2024-11-20
add KEP 238, to add controller revision

Edwinhr716:
#### What type of PR is this?
/kind documentation

Kep for the fix for #238 #240 #281

Fixes #

```release-note

```


Edwinhr716:
@kerthcet PTAL as well

keps/238-controller-revision/README.md
```diff
@@ -0,0 +1,394 @@
...
+		lws.Status.CurrentRevision = lws.Status.UpdateRevision
+	}
+}
+```
+
+### Pod Controller
+
+In order to determine what worker pod spec to use to create the worker pods, we can compare 
+the value of the template hash generated by the LWS object, with the template hash that the leader pod 
+hash. Because the leader pod spec is determined by the statefulset controller, there is a guarantee that
```
ahg-g, line 211: s/hash/has?

keps/238-controller-revision/README.md
```diff
@@ -0,0 +1,394 @@
...
+	lws.Status.CurrentRevision = currentRevision.Name
+	lws.Status.UpdateRevision = updateRevision.Name
+	lws.Status.CollisionCount = new(int32)
+	lws.Status.CollisionCount = &collisionCount
+
+	partition, replicas, err := r.rollingUpdateParameters(ctx, lws)
+}
+```
+
+Once the update has been determined to be done, `currentRevision` will be set to be the value of `updateRevision`
```
ahg-g, line 196: can you clarify when updateRevision gets created/set?

Edwinhr716: `updateRevision` gets created when there isn't an equal revisions. So the pseudocode is

```golang
 func GetLeaderWorkerSetRevisions(ctx, r.Client, lws) {
    revisions := ListControllerRevisions()
    updateRevision := NewRevision(lws)
    equalRevisions := FindEqualRevisions(revisions, updateRevision)
    if len(equalRevisions) == 0 {
        CreateControllerRevision(lws, updateRevision)
    }
}
```

CurrentRevision doesn't get updated/changed unless `lws.CurrentRevision` is nil. So during an update, `updateRevision` and `currentRevision` will be different until the update is complete.

keps/238-controller-revision/README.md
```diff
@@ -0,0 +1,394 @@
...
+}
+```
+
+### Pod Controller
+
+In order to determine what worker pod spec to use to create the worker pods, we can compare 
+the value of the template hash generated by the LWS object, with the template hash that the leader pod 
+hash. Because the leader pod spec is determined by the statefulset controller, there is a guarantee that 
+it will always have the right pod spec. So, if the template hashes don't match, it means that the leader pod was created 
+using the old pod spec, meaning the old worker pod spec needs to be used.
```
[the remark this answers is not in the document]
Edwinhr716: We don't block updates if an update is in progress. 

It is a good point, there's a scenario where another update causes the difference. If an update is started from 1 to 2, and then is updated to 3, if the update to 3 happens when the leader pod has been updated to 2, but the worker pod hasn't, then the worker pod will be created with 1, causing the discrepancy again. 

To address this, we could do something like this

```golang
func constructWorkerStatefulSetApplyConfiguration(currentRevision) {
   if updatedTemplateHash != leaderPod.Labels[templateHash] {
        originalLws, err := ApplyRevision(&lws, currentRevision)
        podTemplateSpec = *originalLws.WorkerTemplate.DeepCopy()
        if !TemplateHashMatches(leaderPod.Labels[templateHash], originalLws) {
              originalLws, err := GetRevision(leaderPod.Labels[templateHash])
         }
     }
}

// iterates through the list of revisions and returns the lws object with the matching template hash
func GetRevision(string templateHash) {
      revisions := ListRevisions()
      for _, revision := range revisions {
             lws, err := ApplyRevision(&lws, revision)
             if lws.Labels[templateHash] == templateHash {
                   return lws
            }
      }
}
```

Wdyt?

keps/238-controller-revision/README.md
Edwinhr716: fixed

keps/238-controller-revision/README.md
```diff
@@ -0,0 +1,394 @@
...
+-->
+
+## Design Details
+The following LWS API fields need to be added to support controller revision.
+
+```golang
+// LeaderWorkerSetStatus defines the observed state of LeaderWorkerSet
+type LeaderWorkerSetStatus struct {
+	// currentRevision, if not empty, indicates the version of lws
+	// used to generate the worker pods in sequence [0,currentReplicas)
```
ahg-g, line 160: can you define what currentReplicas, replicas and updatedReplicas?

keps/238-controller-revision/README.md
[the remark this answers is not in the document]
ahg-g: sounds good, so we try to find the correct revision by finding the one with a matching hash, makes sense.

ahg-g (commented):
I think we can simplify this as follows:

1) Create a ControllerRevision each time we set the leader sts with a different template hash ([here](https://github.com/kubernetes-sigs/lws/blob/4cf14aca0bd6bd071083e5f5f0d5d9f33b9d50bb/pkg/controllers/leaderworkerset_controller.go#L545)); 
2) In the pod reconciler, we simply lookup the revision with a matching template hash
3) Truncate history each time an update is done, leaving only the controllerRevision with a matching hash of the current template

With this, no need for maintain anything in status.

Edwinhr716:
Makes sense, I'll update the KEP with those changes, and add changes to fix #281

ahg-g (commented):
Discussed offline:

- Nothing will be tracked in lws status
- We will use the template hash as a label on each new controller revision that gets created. This key is stored as a label on the leaders sts.

On each lws reconcile:
- Use the hash key on the leader sts to lookup the latest controller revision
- If one exists, compare the existing controller revision with the current lws spec: 
  - If different, trigger an update
  - If the same, don't trigger an update
- If no one exists, create one and don't trigger an update
- Each time an update is done, trim the controller revisions and only keep the one with the same hash key as the current leaders sts

One each leader pod reconcile:
- Use the hash key to lookup the controller revision 
- Use the worker template in the controller revision to create workers sts

keps/238-controller-revision/README.md
```diff
@@ -0,0 +1,393 @@
...
+
+### LWS controller
+In order to fix #281, we will use the hash key in the leaderSts to make a controller revision lookup, and get the LWS object that was used to create it. To determine if an update has happened, we'll compare the fields that are used to generate the template hash.
+
+
+```golang
+func templateUpdated(sts, lws) bool {
+	controllerRevision := GetLeaderWorkerSetRevisionFromTemplateHash(sts.Labels[templateHash])
+	baselineLws:= controllerutils.ApplyRevision(lws, controllerRevision)
+	return !utils.EqualLeaderWorkerTemplates(baselineLws, lws)
```
ahg-g, line 178: we also need to check some parameters of the LWS spec itself, specifically the subdomain policy

Tacorico787: Yes, `equalLeaderWorkerTemplates` would check both `lws.spec.leaderWorkerTemplate` and checking `subdomainPolicy`, also treating `networkConfig == nil` as `subdomainShared` just like it is done for `templateHash`

Edwinhr716: Yes, equalLeaderWorkerTemplates would check both lws.spec.leaderWorkerTemplate and checking subdomainPolicy, also treating networkConfig == nil as subdomainShared just like it is done for templateHash

PR #266, opened 2024-11-24
Add helm charts for lws.

### LINKED ISSUE #260: Adding Helm Chart for lws
Helm chart is now a normal artifact, widely used in CI/CD systems or displayed in application market, it seems charming if we can support this besides kustomize.  I am interested in contributing a Helm chart.

**Why is this needed**:

- A standard artifact to end-users.

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

wqlparallel:
#### What type of PR is this?
/kind feature

Added helm charts for lws.

Fixes #260

```release-note

```


charts/lws/README.md
```diff
@@ -0,0 +1,67 @@
...
+#### Prerequisites
+
+- [Helm](https://helm.sh/docs/intro/quickstart/#install-helm)
+
+#### Installing the chart
+
+##### Install chart using Helm v3.0+
+
+```bash
+$ git clone git@github.com:kubernetes-sigs/kueue.git
```
kannon92, line 27:
```suggestion
$ git clone git@github.com:kubernetes-sigs/lws.git
```

charts/lws/README.md
```diff
@@ -0,0 +1,67 @@
...
+$ cd charts
+$ helm install lws lws --create-namespace --namespace lws-system
+```
+
+##### Verify that controller pods are running properly.
+
+```bash
+$ kubectl get deploy -n kueue-system
```
kannon92, line 35:
```suggestion
$ kubectl get deploy lws-system
```

wqlparallel: sorry，I referenced the queue project during development because the helm charts for queue and lws are quite similar.  I've corrected the submission now. The helm chart has been tested and appears to be functioning correctly.

kerthcet:
/hold

Put my concern here: https://github.com/kubernetes-sigs/lws/issues/260#issuecomment-2497316210

charts/lws/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "v0.4.2"
```
ahg-g, line 24: is this the only place that we need to update every time we make a release?

charts/lws/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+#### Installing the chart
+
+##### Install chart using Helm v3.0+
+
+```bash
+$ git clone git@github.com:kubernetes-sigs/lws.git
+$ cd charts
+$ helm install lws lws --create-namespace --namespace lws-system
+```
```
ahg-g, line 30: how do we install this from a release package? can we add instructions to making the helm charts an artifact of the release?

ahg-g (commented):
can we add a link to the helm-based installation instructions in https://github.com/kubernetes-sigs/lws/blob/main/docs/setup/install.md

charts/lws/Chart.yaml
kerthcet: I think this refers to the lws version, but we need to update the release version as well.

charts/lws/README.md
kerthcet: Maybe we can store the helm packages in github with gh-pages. But that's the next step.

kerthcet (commented):
Hi @wqlparallel could you solve the comments? And does this support install lws in a different namespace rather than lws-system?

charts/lws/Chart.yaml
wqlparallel: In each release, we should update at least two fields: `appVersion` and `version` . We typically update `appVersion` when we're releasing a new version of LWS. But, always update the `version` field with any change to the Helm chart. If there are no other changes to the Helm chart-liske adding more parameters or configuration-then you'll only need to focus on those two.

charts/lws/README.md
wqlparallel: Installing lws from a release package sounds good, and I agree that it's a step we consider next.

charts/lws/README.md
```diff
@@ -0,0 +1,67 @@
...
+    - [Configuration](#configuration)
+<!-- /toc -->
+
+### Installation
+
+Quick start instructions for the setup and configuration of lws using Helm.
+
+#### Prerequisites
+
+- [Helm](https://helm.sh/docs/intro/quickstart/#install-helm)
```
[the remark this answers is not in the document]
wqlparallel: Thank you for your comment! I believe that Cert-manager should not be listed as a prerequisite for the Helm chart since it can be installed optionally if the user sets `enableCertManager: true` in `values.yaml`. In this context, prerequisites should only include Helm itself, while Cert-manager is treated as an application dependency that users can choose to enable.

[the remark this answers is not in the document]
kerthcet: That's why it's optional

charts/lws/Chart.yaml
ahg-g: please add release instructions, and I will make them part of the release template when this gets merged

wqlparallel:
> Hi @wqlparallel could you solve the comments? And does this support install lws in a different namespace rather than lws-system?

I have committed the code to address the comments.  And this Helm chart supports installing the lws in a different namespace when referencing the main image. However, it may not work as expected in the latest release (v0.4.2) because the commit that [deployed in the specified namespace](https://github.com/kubernetes-sigs/lws/commit/1fe278e6470862113c743c67b53568cc7ae98d09) hasn’t been merged yet.

charts/lws/README.md
kerthcet: Install in a different namespace is another thing I'm concerned, but that can be follow up together.

kerthcet:
Please squash, I'm ok with the PR. Leave to @ahg-g for final approval.

charts/lws/README.md
ahg-g: can you please open issues to track the follow up items?

kerthcet: I created here https://github.com/kubernetes-sigs/lws/issues/289

PR #282, opened 2024-12-13
Use the namespace of the service account

kannon92:
#### What type of PR is this?
/kind feature

Users should not need to specify namespace in the deployment of the LWS operator. 
We want to default the namespace to where the serviceaccount of LWS lives. This allows us to avoid configuration problems where a user specifies the wrong namespace for the CLI and the deployment went to other namespace. In this case, the pod is non functional.

 
Fixes #

```release-note
If the namespace of leaderworkerset deployment is different from default, we will take the service account for the namespace.
```

### DIFF
--- a/docs/setup/install.md
+++ b/docs/setup/install.md
@@ -67,36 +67,6 @@ To uninstall LeaderWorkerSet, run the following command:
 make undeploy
 ```
 
-# Install in a different namespace
-
-To install the leaderWorkerSet controller in a different namespace rather than `lws-system`, you should first:
-
-```sh
-git clone https://github.com/kubernetes-sigs/lws.git
-cd lws
-```
-
-Then change the [kustomization.yaml](../../config/default/kustomization.yaml) _namespace_ field as:
-
-```yaml
-namespace: <your-namespace>
-```
-
-You should change the [manager_auth_proxy_patch.yaml](../../config/default/manager_auth_proxy_patch.yaml) as well:
-
-```yaml
-- name: manager
-  args:
-  - "--namespace=<your-namespace>"
-```
-
-Finally run:
-
-```
-IMAGE_REGISTRY=<registry>/<project> make image-push deploy
-```
-
-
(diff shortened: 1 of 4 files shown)


pkg/utils/utils.go
```diff
@@ -70,3 +76,13 @@ func SortByIndex[T appsv1.StatefulSet | corev1.Pod | int](indexFunc func(T) (int
 
 	return result
 }
+
+// GetOperatorNamespace will pick the namespace based on the serviceaccount
+func GetOperatorNamespace() string {
```
ahg-g, line 81: Is doing this a common pattern? what other controllers does this?

kannon92: https://github.com/kubernetes-sigs/kueue/blob/8f51174d9d018f85fc09c3ab72769f5a7e7ff097/apis/config/v1beta1/defaults.go#L55

Kueue and Jobset both use this for the default namespace on the component config.

ahg-g:
/lgtm
/approve

Thanks

docs/setup/install.md
```diff
@@ -67,36 +67,6 @@ To uninstall LeaderWorkerSet, run the following command:
...
-# Install in a different namespace
-
-To install the leaderWorkerSet controller in a different namespace rather than `lws-system`, you should first:
-
-```sh
-git clone https://github.com/kubernetes-sigs/lws.git
-cd lws
-```
-
-Then change the [kustomization.yaml](../../config/default/kustomization.yaml) _namespace_ field as:
```
kerthcet, line 79: If we want to install in another namespace, I guess this is still needed.

kerthcet (commented):
Have you tried with the new commit? Seems failed to me @kannon92

docs/setup/install.md
kannon92: https://github.com/kubernetes-sigs/lws/pull/285

Added that back. I did test by changing this before I opened this up for review.

PR #297, opened 2024-12-30
Tests for Controller Revision

### LINKED ISSUE #294: Add integration tests for the controller revision enhancement

Edwinhr716:
#### What type of PR is this?
/kind cleanup

Adds tests for Controller Revision

Fixes #294

```release-note

```

### DIFF
--- a/test/integration/controllers/leaderworkerset_test.go
+++ b/test/integration/controllers/leaderworkerset_test.go
@@ -1592,7 +1592,7 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
 						testing.DeleteLeaderPod(ctx, k8sClient, lws, 0, 1)
 						var leaderSts appsv1.StatefulSet
 						gomega.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: lws.Name, Namespace: lws.Namespace}, &leaderSts)).To(gomega.Succeed())
-						testing.CreateNotUpdatedLeaderPods(ctx, leaderSts, k8sClient, lws, 0, 1)
+						testing.CreateLeaderPodsFromRevisionNumber(ctx, leaderSts, k8sClient, lws, 0, 1, 1)
 					},
 					checkLWSState: func(lws *leaderworkerset.LeaderWorkerSet) {
 						testing.ExpectValidLeaderStatefulSet(ctx, k8sClient, lws, 4)
@@ -1675,7 +1675,7 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
 					checkLWSState: func(lws *leaderworkerset.LeaderWorkerSet) {
 						var leaderPod corev1.Pod
 						gomega.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: lws.Name + "-3", Namespace: lws.Namespace}, &leaderPod)).To(gomega.Succeed())
-						gomega.Expect(leaderPod.DeletionTimestamp == nil).To(gomega.BeTrue())
+						gomega.Consistently(leaderPod.DeletionTimestamp == nil, testing.Timeout, testing.Interval).Should(gomega.BeTrue())
 					},
 				},
 				{
--- a/test/integration/controllers/leaderworkerset_test.go
+++ b/test/integration/controllers/leaderworkerset_test.go
@@ -1531,6 +1531,158 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
 				},
 			},
 		}),
+		ginkgo.Entry("Not updated worker gets recreated with old worker spec if restarted during update", &testCase{
+			makeLeaderWorkerSet: func(nsName string) *testing.LeaderWorkerSetWrapper {
+				return testing.BuildLeaderWorkerSet(nsName).Replica(4)
+			},
+			updates: []*update{
+				{
+					// Set lws to available condition.
+					lwsUpdateFn: func(lws *leaderworkerset.LeaderWorkerSet) {
+						testing.SetPodGroupsToReady(ctx, k8sClient, lws, 4)
+					},
+					checkLWSState: func(lws *leaderworkerset.LeaderWorkerSet) {
+						testing.ExpectLeaderWorkerSetAvailable(ctx, k8sClient, lws, "All replicas are ready")
+						testing.ExpectValidLeaderStatefulSet(ctx, k8sClient, lws, 4)
+						testing.ExpectValidWorkerStatefulSets(ctx, lws, k8sClient, true)
+						testing.ExpectLeaderWorkerSetStatusReplicas(ctx, k8sClient, lws, 4, 4)
+					},
+				},
+				{
+					// Check the rolling update initial state.
+					lwsUpdateFn: func(lws *leaderworkerset.LeaderWorkerSet) {
+						testing.UpdateWorkerTemplate(ctx, k8sClient, lws)
+					},
+					checkLWSState: func(lws *leaderworkerset.LeaderWorkerSet) {
+						testing.ExpectValidLeaderStatefulSet(ctx, k8sClient, lws, 4)
+						testing.ExpectLeaderWorkerSetUnavailable(ctx, k8sClient, lws, "All replicas are ready")
+						testing.ExpectLeaderWorkerSetProgressing(ctx, k8sClient, lws, "Replicas are progressing")
+						testing.ExpectLeaderWorkerSetUpgradeInProgress(ctx, k8sClient, lws, "Rolling Upgrade is in progress")
+						testing.ExpectLeaderWorkerSetStatusReplicas(ctx, k8sClient, lws, 4, 0)
+					},
+				},
+				{
+					// Rolling update 1 replica.
+					lwsUpdateFn: func(lws *leaderworkerset.LeaderWorkerSet) {
+						testing.SetPodGroupToReady(ctx, k8sClient, lws.Name+"-3", lws)
+					},
+					checkLWSState: func(lws *leaderworkerset.LeaderWorkerSet) {
+						testing.ExpectValidLeaderStatefulSet(ctx, k8sClient, lws, 4)
+						testing.ExpectUpdatedWorkerStatefulSet(ctx, k8sClient, lws, lws.Name+"-3")
+						testing.ExpectNotUpdatedWorkerStatefulSet(ctx, k8sClient, lws, lws.Name+"-2")
+						testing.ExpectNotUpdatedWorkerStatefulSet(ctx, k8sClient, lws, lws.Name+"-1")
+						testing.ExpectNotUpdatedWorkerStatefulSet(ctx, k8sClient, lws, lws.Name+"-0")
+						testing.ExpectLeaderWorkerSetUnavailable(ctx, k8sClient, lws, "All replicas are ready")
+						testing.ExpectLeaderWorkerSetProgressing(ctx, k8sClient, lws, "Replicas are progressing")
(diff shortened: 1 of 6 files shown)


Edwinhr716:
Added integration tests that were previously e2e tests. 
Added unit tests for revision_util functions
Still to add:
- Check for amount of revisions in history before, during, and after update
- e2e test that tries to replicate updating from a previous LWS version

test/integration/controllers/leaderworkerset_test.go
```diff
@@ -1527,6 +1531,164 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
...
+					},
+				},
+				{ // The workerPods are deleted to update them. This should not delete the leaderPod
+					lwsUpdateFn: func(lws *leaderworkerset.LeaderWorkerSet) {
+						testing.DeleteWorkerPods(ctx, k8sClient, lws)
+					},
+					checkLWSState: func(lws *leaderworkerset.LeaderWorkerSet) {
+						var leaderPod corev1.Pod
+						gomega.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: lws.Name + "-3", Namespace: lws.Namespace}, &leaderPod)).To(gomega.Succeed())
+						gomega.Expect(leaderPod.DeletionTimestamp == nil).To(gomega.BeTrue())
```
ahg-g, line 1678: This doesn't guarantee that the leader was not updated because the pod reconciler may not have been invoked yet by the time we get here. To have a higher probability that the pod reconciler was invoked, wrap those three lines in a Consistently clause with a timeout of 3 seconds for example (i.e., that the leader pod consistently exists and not deleted)

test/integration/controllers/leaderworkerset_test.go
```diff
@@ -1527,6 +1531,164 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
...
+						testing.SetPodGroupsToReady(ctx, k8sClient, lws, 4)
+					},
+					checkLWSState: func(lws *leaderworkerset.LeaderWorkerSet) {
+						testing.ExpectValidLeaderStatefulSet(ctx, k8sClient, lws, 4)
+						testing.ExpectValidWorkerStatefulSets(ctx, lws, k8sClient, true)
+						testing.ExpectLeaderWorkerSetNotProgressing(ctx, k8sClient, lws, "Replicas are progressing")
+						testing.ExpectLeaderWorkerSetNoUpgradeInProgress(ctx, k8sClient, lws, "Rolling Upgrade is in progress")
+						testing.ExpectLeaderWorkerSetAvailable(ctx, k8sClient, lws, "All replicas are ready")
 					},
 				},
```
ahg-g, line 1693: Can we add another step where the workers are created from the same revision and a delete for one of the workers did trigger deleting the leader?

Edwinhr716: Isn't that covered in this test https://github.com/kubernetes-sigs/lws/blob/main/test/integration/controllers/leaderworkerset_test.go#L423? Or are we checking that it still works after an update is done?

test/testutils/util.go
```diff
@@ -160,6 +175,74 @@ func CreateLeaderPods(ctx context.Context, leaderSts appsv1.StatefulSet, k8sClie
...
+		}
+		var podTemplateSpec corev1.PodTemplateSpec
+		// if leader template is nil, use worker template
+		if lws.Spec.LeaderWorkerTemplate.LeaderTemplate != nil {
+			podTemplateSpec = *oldLws.Spec.LeaderWorkerTemplate.LeaderTemplate.DeepCopy()
+		} else {
+			podTemplateSpec = *oldLws.Spec.LeaderWorkerTemplate.WorkerTemplate.DeepCopy()
+		}
+		for i := start; i < end; i++ {
+			pod := corev1.Pod{
```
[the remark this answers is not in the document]
Edwinhr716: Do you mean use with 
https://github.com/kubernetes-sigs/lws/blob/e69c972f20946f577888678853e60400c7d72e42/test/testutils/util.go#L119
 WorkerPods require different labels, which would either extend the amount of arguments or add if statements to differentiate between worker and leader pods.

test/testutils/util.go
```diff
@@ -160,6 +175,74 @@ func CreateLeaderPods(ctx context.Context, leaderSts appsv1.StatefulSet, k8sClie
 	return nil
 }
 
+func CreateNotUpdatedLeaderPods(ctx context.Context, leaderSts appsv1.StatefulSet, k8sClient client.Client, lws *leaderworkerset.LeaderWorkerSet, start int, end int) {
```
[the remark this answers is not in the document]
Edwinhr716: Changed name to `CreateLeaderPodsFromRevisionNumber` which I think is a better name after changing the logic

test/testutils/util.go
```diff
@@ -160,6 +175,74 @@ func CreateLeaderPods(ctx context.Context, leaderSts appsv1.StatefulSet, k8sClie
...
+	gomega.Eventually(func() error {
```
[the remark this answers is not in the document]
Edwinhr716: We are not waiting on a particular event to become true. Switched to `gomega.Expect()` for error handling

test/testutils/util.go
```diff
@@ -128,6 +139,36 @@ func CreateLeaderPods(ctx context.Context, leaderSts appsv1.StatefulSet, k8sClie
...
+}
+
+func CreateLeaderPodsFromRevisionNumber(ctx context.Context, leaderSts appsv1.StatefulSet, k8sClient client.Client, lws *leaderworkerset.LeaderWorkerSet, start, end, revisionNumber int) {
+	selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: map[string]string{
+		leaderworkerset.SetNameLabelKey: lws.Name,
+	}})
+	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+	revisions, err := revisionutils.ListRevisions(ctx, k8sClient, lws, selector)
+	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+	var oldRevision *appsv1.ControllerRevision
```
ahg-g, line 153: nit, I would call it targetRevision instead

test/testutils/util.go
```diff
@@ -128,6 +139,36 @@ func CreateLeaderPods(ctx context.Context, leaderSts appsv1.StatefulSet, k8sClie
 	if err != nil {
 		return err
 	}
+
+	return createLeaderPods(ctx, k8sClient, leaderSts, lws, podTemplateSpec, revisionutils.GetRevisionKey(cr), start, end)
+}
+
+func CreateLeaderPodsFromRevisionNumber(ctx context.Context, leaderSts appsv1.StatefulSet, k8sClient client.Client, lws *leaderworkerset.LeaderWorkerSet, start, end, revisionNumber int) {
```
ahg-g, line 146: I am afraid that the parameters at the end can easily be mixed up since they are all integers. Do we really need the start/end? If not yet, lets remove them and assume 0 to `leaderSts.Replicas`.

test/testutils/util.go
```diff
@@ -128,6 +139,36 @@ func CreateLeaderPods(ctx context.Context, leaderSts appsv1.StatefulSet, k8sClie
...
+	}
+	oldLws, err := revisionutils.ApplyRevision(lws, oldRevision)
+	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+	var podTemplateSpec corev1.PodTemplateSpec
+	// if leader template is nil, use worker template
+	if lws.Spec.LeaderWorkerTemplate.LeaderTemplate != nil {
+		podTemplateSpec = *oldLws.Spec.LeaderWorkerTemplate.LeaderTemplate.DeepCopy()
+	} else {
+		podTemplateSpec = *oldLws.Spec.LeaderWorkerTemplate.WorkerTemplate.DeepCopy()
+	}
```
ahg-g, line 167: can we move this inside createLeaderPods? this will help remove one parameter from `createLeaderPods` func and dedup some logic that we also have in `CreateLeaderPods`

test/testutils/util.go
Edwinhr716: We do need start/end, since there are cases where we don't want to recreate all of them (like in the worker spec mismatch integration test)

test/testutils/util.go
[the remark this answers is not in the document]
Edwinhr716: refactored to use with `CreateLeaderPods`, not with `CreateWorkerPodsForLeaderPod`

test/integration/controllers/leaderworkerset_test.go
```diff
@@ -1691,6 +1696,20 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
 						testing.ExpectLeaderWorkerSetAvailable(ctx, k8sClient, lws, "All replicas are ready")
 					},
 				},
+				{ // Validate that RecreateGroupOnPodRestart works as intended after update
+					lwsUpdateFn: func(lws *leaderworkerset.LeaderWorkerSet) {
+						var workers corev1.PodList
+						gomega.Expect(k8sClient.List(ctx, &workers, client.InNamespace(lws.Namespace), &client.MatchingLabels{"worker.pod": "workers"})).To(gomega.Succeed())
+						gomega.Expect(k8sClient.Delete(ctx, &workers.Items[0])).To(gomega.Succeed())
```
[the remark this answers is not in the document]
Edwinhr716: Because no other worker pods exist. We only create the workerPods for group 3 https://github.com/kubernetes-sigs/lws/pull/297/files#diff-7ab7c49611a2824f44a7f0e343828980195762c5c577196485ca086d12d46b3cR1687-R1688. The only way that worker Pods exist is if we create them manually.

test/testutils/util.go
ahg-g: when I search for CreateLeaderPodsFromRevisionNumber, I see a single call to it

Edwinhr716: Yes, in that call we only want to create a single leaderPod with the old revision, not all of them. So we can't do `leaderSts.Replicas` for the end, otherwise we will attempt to create 4, which will fail since we only delete one of them.

test/integration/controllers/leaderworkerset_test.go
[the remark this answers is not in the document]
ahg-g: This is a hidden assumption that can be a cause for confusion. I would still ensure that this is the case in the selector used for the list (we can use the group index label).

test/testutils/util.go
ahg-g: ok, I thought it is 2, sg.

test/integration/controllers/leaderworkerset_test.go
```diff
@@ -1714,6 +1714,27 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
...
+					lwsUpdateFn: func(lws *leaderworkerset.LeaderWorkerSet) {
+						testing.UpdateLeaderStatefulSetRevisionKey(ctx, k8sClient, lws, "template-hash")
+						testing.SetPodGroupsToReady(ctx, k8sClient, lws, 2)
+					},
+					checkLWSState: func(lws *leaderworkerset.LeaderWorkerSet) {
+						testing.ExpectLeaderWorkerSetAvailable(ctx, k8sClient, lws, "All replicas are ready")
+						testing.ExpectRevisions(ctx, k8sClient, lws, 1)
+						testing.ExpectValidLeaderStatefulSet(ctx, k8sClient, lws, 2)
+						testing.ExpectValidWorkerStatefulSets(ctx, lws, k8sClient, true)
+						testing.ExpectLeaderWorkerSetStatusReplicas(ctx, k8sClient, lws, 2, 2)
```
ahg-g, line 1733: we need to add another expectation that there is a CR with revisionKey matching the one we used to update the leaders sts

PR #298, opened 2025-01-03
Minimize number of unnecessary logs

Edwinhr716:
Minimize the number of unnecessary logs that get emitted

Fixes #

```release-note

```

### DIFF
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -376,8 +376,11 @@ func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *l
 		var sts appsv1.StatefulSet
 		if !noWorkerSts {
 			if err := r.Get(ctx, client.ObjectKey{Namespace: lws.Namespace, Name: pod.Name}, &sts); err != nil {
-				log.Error(err, "Fetching worker statefulSet")
-				return false, false, err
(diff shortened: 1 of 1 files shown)


pkg/controllers/leaderworkerset_controller.go
```diff
@@ -476,8 +479,16 @@ func (r *LeaderWorkerSetReconciler) updateStatus(ctx context.Context, lws *leade
 	if err != nil {
 		return false, err
 	}
+	// Get the most up to date lws object to merge the status with
```
ahg-g, line 482: This is forcing an update, I think this is dangerous to do.

A better option is to check if the error is a conflict, in which case we requeue the key and return without printing an error, we should do that here: https://github.com/kubernetes-sigs/lws/blob/7a3cd29a7f0e6ecb5717108edf387e8ba05f6993/pkg/controllers/leaderworkerset_controller.go#L151

Edwinhr716: So switch it back to `status().Update()`?

ahg-g: yes, and `return ctrl.Result{Requeue: true}, nil` when the error is a conflict

Edwinhr716: Makes sense. Will also change it so that it doesn't log the error if there is a conflict

PR #304, opened 2025-01-07
Ignore not found error when handling ReacreateGroupOnPodRestart

Edwinhr716:
#### What type of PR is this?
/kind cleanup

Returns gracefully if the leaderPod is not found when handling a worker pod. 
This happens when a rolling update is occurring. The leaderPod gets deleted, which triggers the worker pods to be deleted. Since the leader has already been deleted, the Get Operation will fail, but this is intended behavior.

Fixes #

```release-note

```

### DIFF
--- a/pkg/controllers/pod_controller.go
+++ b/pkg/controllers/pod_controller.go
@@ -188,6 +188,8 @@ func (r *PodReconciler) handleRestartPolicy(ctx context.Context, pod corev1.Pod,
 			return false, fmt.Errorf("parsing pod name for pod %s", pod.Name)
 		}
 		if err := r.Get(ctx, types.NamespacedName{Name: leaderPodName, Namespace: pod.Namespace}, &leader); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controllers/pod_controller.go
```diff
@@ -188,7 +188,7 @@ func (r *PodReconciler) handleRestartPolicy(ctx context.Context, pod corev1.Pod,
 			return false, fmt.Errorf("parsing pod name for pod %s", pod.Name)
 		}
 		if err := r.Get(ctx, types.NamespacedName{Name: leaderPodName, Namespace: pod.Namespace}, &leader); err != nil {
-			return false, err
+			return false, client.IgnoreNotFound(err)
```
ahg-g, line 191: add a comment explaining why we are ignoring the error here: it is because in this case the leader was likely deleted and the deletion of the worker sts didn't yet propagate all the way to delete the worker pods.

PR #284, opened 2024-12-16
Replace kube-rbac-proxy

anshuman-agarwala:
#### What type of PR is this?
/kind deprecation

#### Which issue(s) this PR fixes

Fixes #274

#### Does this PR introduce a user-facing change?

```release-note

```

PR #307, opened 2025-01-08
Fixed metrics service port, added log level flag back

Edwinhr716:
#### What type of PR is this?
/kind bug

Addressed open comments left in #284 

Fixes #

```release-note

```

### DIFF
--- a/config/default/manager_metrics_service.yaml
+++ b/config/default/manager_metrics_service.yaml
@@ -10,8 +10,8 @@ metadata:
 spec:
   ports:
   - name: https
-    port: 8080
+    port: 8443
     protocol: TCP
-    targetPort: 8080
+    targetPort: 8443
   selector:
     control-plane: controller-manager
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -71,7 +71,7 @@ func main() {
 		leaderElectionID         string
 	)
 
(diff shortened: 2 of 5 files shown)


ahg-g:
@Edwinhr716 need to fix the port in the e2e test: https://github.com/kubernetes-sigs/lws/blob/main/test/e2e/e2e_test.go#L350 and https://github.com/kubernetes-sigs/lws/blob/main/test/e2e/e2e_test.go#L369

Edwinhr716:
I think the changes are more complicated than that. Switched it from 8080 to 8443, got this error 

```
  The function passed to Eventually failed at lws/test/e2e/e2e_test.go:381 with:
  curl pod in wrong status
  Expected
      <string>: Failed
  to equal
      <string>: Succeeded
```

config/default/manager_metrics_service.yaml
```diff
@@ -10,8 +10,8 @@ metadata:
 spec:
   ports:
   - name: https
-    port: 8080
+    port: 8443
     protocol: TCP
-    targetPort: 8080
+    targetPort: 8443
```
ahg-g, line 15: The target port is different: https://github.com/kubernetes-sigs/lws/blob/main/cmd/main.go#L74

ahg-g: Can you change the default in main?

PR #310, opened 2025-01-08
Add public vLLM on TPU image

Edwinhr716:
#### What type of PR is this?
/kind documentation

Fixes #

```release-note

```

### DIFF
--- a/docs/examples/vllm/build/Dockerfile.TPU
+++ b/docs/examples/vllm/build/Dockerfile.TPU
@@ -1,4 +1,2 @@
-# The TPU vllm image can be built using the dockerfile found in 
-# in the vllm repo https://github.com/vllm-project/vllm/blob/main/Dockerfile.tpu
-FROM tpu_vllm_image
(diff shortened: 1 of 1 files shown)


docs/examples/vllm/build/Dockerfile.TPU
```diff
@@ -1,4 +1,2 @@
-# The TPU vllm image can be built using the dockerfile found in 
-# in the vllm repo https://github.com/vllm-project/vllm/blob/main/Dockerfile.tpu
-FROM tpu_vllm_image
+FROM docker.io/vllm/vllm-tpu:2e33fe419186c65a18da6668972d61d7bbc31564
```
ahg-g, line 1: hmm, can we use a tagged version, like latest?

Edwinhr716: There exist a `nightly` tag, but there isn't much test coverage for the TPU image yet, so there is no guarantee that it will work. The tag that I added has been validated, which is why I added that one.

PR #311, opened 2025-01-09
Removed dependency to controller-runtime from v1 package

### LINKED ISSUE #283: Use separate module names for the client-go and api pkgs from the operator implementation

Allow developers to import the LWS api and client-go pkgs without the operator implementation.

**Why is this needed**:

In some cases, the dependencies that the operator brings with it are not compatible or not desirable with applications that would like to import lws api and client-go library only

See https://go.dev/doc/modules/managing-source

Edwinhr716:
#### What type of PR is this?
/kind documentation

Removes the dependency to controller-runtime from the v1 package

Fixes #283

```release-note

```

### DIFF
--- a/api/leaderworkerset/scheme/scheme.go
+++ b/api/leaderworkerset/scheme/scheme.go
@@ -1,0 +1,96 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


api/leaderworkerset/scheme/scheme.go
```diff
@@ -0,0 +1,96 @@
+/*
```
ahg-g, line 1: please update the year and add a comment that this is adapted from controller-runtime and that we are doing this to cut the api pkg dependency on controller-runtime

api/leaderworkerset/scheme/scheme.go
```diff
@@ -0,0 +1,96 @@
...
+//	}
+//
+//	func main() {
+//		mgr := controllers.NewManager(context.Background(), controllers.GetConfigOrDie(), manager.Options{
+//			Scheme: scheme,
+//		})
+//		// ...
+//	}
+
+// This package is taken from https://github.com/kubernetes-sigs/controller-runtime/blob/c80ea332ca131e1b79caab274a3413b93dc97e62/pkg/scheme/scheme.go
```
ahg-g, line 55: oh, i see it here

PR #313, opened 2025-01-09
Refactored wrappers into its own package

### LINKED ISSUE #296: Split test wrappers into a different package from testutils
We should create a new package for the wrapper functions https://github.com/kubernetes-sigs/lws/blob/main/test/testutils/wrappers.go

**Why is this needed**: 
testutils wrappers are used outside of the testutils package. This can cause a dependency cycle. For instance, validators.go https://github.com/kubernetes-sigs/lws/blob/main/test/testutils/validators.go#L153 and util.go https://github.com/kubernetes-sigs/lws/blob/main/test/testutils/util.go#L60 use revision functions. This means that the revision package cannot import the testutils package to use the wrapper functions since it creates a dependency cycle (testutils -> revision -> testutils), forcing us to have duplicate functions in unit tests https://github.com/kubernetes-sigs/lws/blob/main/pkg/utils/revision/revision_utils_test.go#L80. 

Splitting it into its own package allows us to avoid this cycle.

Further context: https://github.com/kubernetes-sigs/lws/pull/277/files#r1898155596

Edwinhr716:
#### What type of PR is this?
/kind cleanup

Fixes #296

```release-note

```

### DIFF
--- a/pkg/controllers/leaderworkerset_controller_test.go
+++ b/pkg/controllers/leaderworkerset_controller_test.go
@@ -34,22 +34,22 @@ import (
 
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	revisionutils "sigs.k8s.io/lws/pkg/utils/revision"
-	testutils "sigs.k8s.io/lws/test/testutils"
+	"sigs.k8s.io/lws/test/wrappers"
 )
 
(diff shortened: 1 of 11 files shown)


ahg-g:
/lgtm

Thanks!

PR #314, opened 2025-01-09
Update KEP #238 to reflect implementation

Edwinhr716:
#### What type of PR is this?
/kind documentation

Fixes #

```release-note

```

### DIFF
--- a/keps/238-controller-revision/README.md
+++ b/keps/238-controller-revision/README.md
@@ -166,10 +166,10 @@ Taking those into account, we'll combine controller revisions and template hashe
 
 
 ### LWS controller
-In order to fix #281, we will use the hash key in the leaderSts to make a controller revision lookup, and get the LWS object that was used to create it. To determine if an update has happened, we'll compare the fields that are used to generate the template hash.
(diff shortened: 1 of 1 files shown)


keps/238-controller-revision/README.md
```diff
@@ -170,57 +168,72 @@ Taking those into account, we'll combine controller revisions and template hashe
 ### LWS controller
 In order to fix #281, we will use the hash key in the leaderSts to make a controller revision lookup, and get the LWS object that was used to create it. To determine if an update has happened, we'll compare the fields that are used to generate the template hash.
 
+There are three different cases where we need to create a new revision
+- Case 1: Upgrading from the LWS controller from a version that doesn't support controller revision
```
ahg-g, line 172:
```suggestion
- Case 1: Upgrading the LWS controller from a version that doesn't support controller revision
```

keps/238-controller-revision/README.md
```diff
@@ -170,57 +168,72 @@ Taking those into account, we'll combine controller revisions and template hashe
...
-```
 
-There are three cases where we need to create a new controllerRevision: 
-- There are no existing controller revisions in the history
-- A new LWS object is created
-- An update has been made to the LWS object
+#### Lifecycle of Revisions
+To cover Case 1 & 2, we create a controller revision if it doesn't exist. 
+
+For Case 1, we use the revisionKey that the leader statefulset has, so that the revisionKey label in the leader statefulset isn't updated,
```
ahg-g, line 180: clarify that the templateHash label is being repurposed as the revisionKey

PR #323, opened 2025-01-16
Remove pods create permission

### LINKED ISSUE #320: Is Pod Create Permission Really Necessary
I've been thinking about whether we really need the pod create permission in our RBAC setup for this project. It's a pretty sensitive permission, and here's why I'm bringing it up: we're trying to get lws into the app marketplace, and this permission popped up as a risk.

From what I understand, the pod creation reference to lws is usually handled by the statefulset controller in kcm. So, does lws itself actually need this permission?

wqlparallel:
#### What type of PR is this?
/kind feature

Remove pods create permission since it's sensitive and unnecessary.

Fixes #320

```release-note

```

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -28,7 +28,6 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - pods
   - services
   verbs:
   - create
@@ -38,6 +37,17 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+    - ""
+  resources:
+    - pods
+  verbs:
+    - delete
+    - get
+    - list
+    - patch
+    - update
+    - watch
 - apiGroups:
   - ""
   resources:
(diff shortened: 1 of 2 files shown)


config/rbac/role.yaml
```diff
@@ -38,6 +37,17 @@ rules:
...
+- apiGroups:
+    - ""
+  resources:
+    - pods
+  verbs:
+    - delete
+    - get
+    - list
+    - patch
+    - update
```
ahg-g, line 49: do we need the patch and update? are they required for the mutating webhook? if not, we can remove those too.

wqlparallel: Based on my understanding, we do need the patch and update permissions. According to this function [link](https://github.com/kubernetes-sigs/lws/blob/57ed41a65518811a50980026e06e40de10292f77/pkg/webhooks/pod_webhook.go#L82), LWS requires patching the pod's labels and annotations.

ahg-g: yes, but I wasn't sure if the mutating webhook requires update/patch permissions since the webhook doesn't explicitly call them

PR #339, opened 2025-01-25
copy the golang code rather than add folders into docker file

kannon92:
#### What type of PR is this?
/kind cleanup

Copy the files with general copy rather than adding them into Dockerfile.

Fixes #

https://github.com/kubernetes-sigs/lws/pull/325#issuecomment-2596927125

We hit this quite a bit in JobSet repo and it is always a source of confusion.

I don't see why we need to make this more complicated than it needs to be.

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,22 +7,10 @@ ARG TARGETOS
 ARG TARGETARCH
 
 WORKDIR /workspace
-# Copy the Go Modules manifests
-COPY go.mod go.mod
-COPY go.sum go.sum
-
-# cache deps before building and copying source so that we don't need to re-download as much
+# cache deps before building so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -7,21 +7,7 @@ ARG TARGETOS
 ARG TARGETARCH
 
 WORKDIR /workspace
-# Copy the Go Modules manifests
-COPY go.mod go.mod
-COPY go.sum go.sum
-
-# cache deps before building and copying source so that we don't need to re-download as much
-# and so that source changes don't invalidate our downloaded layer
-RUN go mod download
```
ahg-g, line 16: this is not needed?

kannon92: I think it is. I’ll add this back and bring that to Jobset.

kannon92: updated.

PR #342, opened 2025-01-27
Added handling of nil revision in pod controller

### LINKED ISSUE #338: possible nil dereference
GetRevision may return nil, however in 

https://github.com/kubernetes-sigs/lws/blob/78998a045cc582a21a6fb75600c90d24178d86d4/pkg/controllers/pod_controller.go#L125

we don't check for that. If the revision was nil, we should requeue after a few seconds.

Edwinhr716:
#### What type of PR is this?
/kind bug

In the case where the lws controller is upgraded when a rolling update is happening, it is possible for the revision that matches the RevisionKey in the leader pod doesn't exist yet, triggering a nil dereference error, and causing a panic. By requeuing, we wait for the revision to be created to avoid the error from happening.

Fixes #338 

```release-note

```

### DIFF
--- a/pkg/controllers/pod_controller.go
+++ b/pkg/controllers/pod_controller.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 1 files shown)


ahg-g:
Can we patch this back as well?

PR #347, opened 2025-01-28
Removed update operation from pod webhook

### LINKED ISSUE #346: Pod webhook intercepts update events, potentially breaking garbage collector
**What happened**:
When upgrading to a new LWS controller version while a rolling update is happening, the leader pod isn't deleted. 

For the leader pod to be deleted.

The leader pod has `foregroundDeletion` so that it is deleted after the worker pods in the pod group. This `foregroundDeletion` is removed by the garbage collector via a pod.patch operation, which then makes it possible for the leader pod to be deleted. 

When upgrading to a new version of LWS, extra environment variables are added. So the pod webhook intercepts the pod.patch operation and looks to add the new environment variables, causing this error

```
Pod <leader pod name> is invalid: spec: Forbidden: pod updates may not change fields other than `spec.containers[*].image`,`spec.initContainers[*].image`,`spec.activeDeadlineSeconds`,`spec.tolerations` (only additions to existing tolerations),`spec.terminationGracePeriodSeconds` (allow it to be set to 1 if it was previously negative)
```

This can be fixed by removing the UPDATE operation from the webhook.

Deploy LWS controller v0.2.0 and trigger a rolling update. While the rolling update is happening, upgrade the LWS controller version to v0.5.0.

**Environment**:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Edwinhr716:
#### What type of PR is this?
/kind bug

Removes the update operation from the pod webhook as it is not needed, and can cause issues with the garbage collector.

Fixes #346 

```release-note

```

### DIFF
--- a/charts/lws/templates/webhook/webhook.yaml
+++ b/charts/lws/templates/webhook/webhook.yaml
@@ -43,7 +43,6 @@ webhooks:
           - v1
         operations:
           - CREATE
-          - UPDATE
         resources:
           - pods
     sideEffects: None
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -40,7 +40,6 @@ webhooks:
     - v1
     operations:
     - CREATE
-    - UPDATE
     resources:
     - pods
   sideEffects: None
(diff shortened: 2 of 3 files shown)


ahg-g:
lets update that in the helm charts too

PR #354, opened 2025-01-29
Remove duplicated label in helm chart: app.kubernetes.io/instance

### LINKED ISSUE #353: Duplicated label in helm chart: app.kubernetes.io/instance
**What happened**: app.kubernetes.io/instance is duplicated on the following resources: 
* ServiceAccount.release-name-lws-controller-manager
* ClusterRole.metrics-reader
* ClusterRole.proxy-role
* ClusterRoleBinding.manager-rolebinding
* ClusterRoleBinding.proxy-rolebinding
* Role.leader-election-role
* RoleBinding.leader-election-rolebinding
* Service.controller-manager-metrics-service
* Service.webhook-service

Helm output shoundn't have duplicated labels, when applying with helm the latest label, the one set by `lws.labels(lws.selectorLabels)` will assume the value of the label, however, if you're using a HelmRelease, it will break the Helmrelease because of the duplicated labels.

```
git clone git@github.com:kubernetes-sigs/lws.git
cd lws/charts
helm template lws
```

It seems like the `Deployment.controller-manager` is the only one that needs the instance label that is added 
The `app.kubernetes.io/instance` label set in the resources is more descriptive than the one that comes from the function lws.labels(lws.selectorLabels), so I assume this is probably the one that you wanna keep it.

```
Client Version: v1.32.1
Kustomize Version: v5.5.0
Server Version: v1.31.4+rke2r1
- LWS version (use `git describe --tags --dirty --always`): `v0.5.0-39-g06f2219`
- Cloud provider or hardware configuration: aws
- OS (e.g: `cat /etc/os-release`): `Ubuntu 22.04.5 LTS`
- Kernel (e.g. `uname -a`): `6.8.0-51-generic #52~22.04.1-Ubuntu`
- Install tools: 
```
version.BuildInfo{Version:"v3.17.0", GitCommit:"v3.17.0", GitTreeState:"", GoVersion:"go1.23.4 /0.1s
``` 
- Others:

### PR DESCRIPTION
**Pull request by DiegoLelis**
#### What type of PR is this?
/kind bug

Changes the behaviour of a couple of functions and introduces 2 other functions to make sure the label `app.kubernetes.io/instance` is not set twice.
Also fixed the typo on the certificate file name.

Fixes #353 

The previous behaviour added the label `app.kubernetes.io/instance` twice on a few resources.
I've tried to keep changes minimal, and to respect the previous design to the best of my ability.
Kept all the previous labels, only removed the duplicated ones.
When choosing to remove between the more generic label, and the more specific one, I choose to deprecate the more generic one.
Tested installing on my cluster with default values and it works.

```release-note
Removes duplicated label `app.kubernetes.io/instance` from helm chart.
```


Edwinhr716:
/lgtm 
makes sense to me to keep the label set in the resources instead

PR #361, opened 2025-02-07
Deprecate flags as the recommended path is to pass configuration file

ardaguclu:
#### What type of PR is this?
/kind cleanup

It was suggested in here https://github.com/kubernetes-sigs/lws/pull/325#discussion_r1923006469 by @kerthcet to deprecate the flags, because passing them in the configuration file via `--config` flag is cleaner.

This PR
* Deprecates the flags as continuation of https://github.com/kubernetes-sigs/lws/pull/325
* Migrates from deprecated function `GetValueFromIntOrPercent` to `GetScaledValueFromIntOrPercent`
* Moves metric server initialization logic in `apply` function as suggested in https://github.com/kubernetes-sigs/lws/pull/325#discussion_r1926762256

Fixes #

```release-note
Deprecating the flags whose can be cleanly passed through configuration file
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -76,28 +76,28 @@ func main() {
 		configFile               string
 	)
 
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8443", "The address the metric endpoint binds to.")
-	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
-	flag.Float64Var(&qps, "kube-api-qps", 500, "Maximum QPS to use while talking with Kubernetes API")
-	flag.IntVar(&burst, "kube-api-burst", 500, "Maximum burst for throttle while talking with Kubernetes API")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8443", "DEPRECATED(please pass configuration file via --config flag): The address the metric endpoint binds to.")
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "DEPRECATED(please pass configuration file via --config flag): The address the probe endpoint binds to.")
+	flag.Float64Var(&qps, "kube-api-qps", 500, "DEPRECATED(please pass configuration file via --config flag): Maximum QPS to use while talking with Kubernetes API")
+	flag.IntVar(&burst, "kube-api-burst", 500, "DEPRECATED(please pass configuration file via --config flag): Maximum burst for throttle while talking with Kubernetes API")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", true,
-		"Enable leader election for controller manager. "+
+		"DEPRECATED(please pass configuration file via --config flag): Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.DurationVar(&leaderElectLeaseDuration, "leader-elect-lease-duration", 15*time.Second,
-		"The duration that non-leader candidates will wait after observing a leadership renewal until attempting to acquire "+
+		"DEPRECATED(please pass configuration file via --config flag): The duration that non-leader candidates will wait after observing a leadership renewal until attempting to acquire "+
 			"leadership of a led but unrenewed leader slot. This is effectively the maximum duration that a leader can be stopped"+
 			" before it is replaced by another candidate. This is only applicable if leader election is enabled.")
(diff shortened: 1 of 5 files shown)


kerthcet:
Thanks @ardaguclu 
/ok-to-test

cmd/main_test.go
```diff
@@ -101,10 +101,13 @@ internalCertManagement:
 				LeaderElection:             false,
 				LeaderElectionResourceLock: "changed",
 				LeaderElectionID:           "changed",
+				LeaderElectionNamespace:    "test-ns",
 				LeaseDuration:              ptr.To(1 * time.Minute),
 				RenewDeadline:              ptr.To(1 * time.Minute),
 				RetryPeriod:                ptr.To(1 * time.Minute),
 				HealthProbeBindAddress:     ":9443",
+				LivenessEndpointName:       "/healthz",
```
ardaguclu, line 109: This is independent from the changes in this PR. Actually, without this assignment, unit tests should be failing on main branch (it fails on my local).  I'm not sure there is an issue in testing side?

ahg-g: Because `cmd` is not included in the test, do you mean adding it please? https://github.com/kubernetes-sigs/lws/blob/main/Makefile#L113

cmd/main.go
```diff
@@ -76,28 +75,28 @@ func main() {
 		configFile               string
 	)
 
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8443", "The address the metric endpoint binds to.")
-	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
-	flag.Float64Var(&qps, "kube-api-qps", 500, "Maximum QPS to use while talking with Kubernetes API")
-	flag.IntVar(&burst, "kube-api-burst", 500, "Maximum burst for throttle while talking with Kubernetes API")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8443", "DEPRECATED(please pass configuration file via --config flag): The address the metric endpoint binds to.")
```
kerthcet, line 78: I'm thinking whether we're too aggressive to deprecate all the flags, should we leave some basic flags for easy configuration, like `metricsAddr` `probeAddr` `qps` `burst` `enableLeaderElection`. I need some feedbacks.

kerthcet (commented):
Thanks, left some comments. Next, we need to add the configurations to the helm chart template, then people can update the values.yaml to inject the flags.

cmd/main_test.go
ardaguclu: @ahg-g  Thank you for your guidance. Unit tests under cmd directory will be executed in `make test` from now on.

cmd/main.go
```diff
@@ -115,7 +114,8 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
-	options, cfg, err := apply(configFile, probeAddr, enableLeaderElection, leaderElectLeaseDuration, leaderElectRenewDeadline, leaderElectRetryPeriod, leaderElectResourceLock, leaderElectionID)
+	namespace := utils.GetOperatorNamespace()
```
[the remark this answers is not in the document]
ardaguclu: Updated.

cmd/main.go
ardaguclu: @kerthcet thank you for your review. I think it makes sense. 

* `kube-api-qps`
* `kube-api-burst`
* `leader-elect`
will  also be supported as flag. I agree with you that if someone wants to explicitly enable leader election, it would be overkill to force using a configuration file and it would be easier by simply passing `leader-elect` flag.

Deprecated ones by this PR;

* `metrics-bind-address`
* `health-probe-bind-address`
* `leader-elect-lease-duration`
* `leader-elect-renew-deadline`
* `leader-elect-retry-period`
* `leader-elect-resource-lock`
* `leader-elect-resource-name`
These configurations seem to fall in to fine-grained tuning category and better to be only represented in configuration file rather than a flag?.

ardaguclu:
> Thanks, left some comments. Next, we need to add the configurations to the helm chart template, then people can update the values.yaml to inject the flags.

I'm planning to update helm chart template in a followup PR.

kerthcet (commented):
Thanks for your patience @ardaguclu 
/lgtm
/approve

Issue #324, opened 2025-01-16
Add adoption page


ahg-g:
We need to reach out to providers to get consent, candidates based on public posts:

- https://docs.nvidia.com/nim/large-language-models/1.1.0/deploy-helm.html#leaderworkerset
- https://aws.amazon.com/blogs/hpc/scaling-your-llm-inference-workloads-multi-node-deployment-with-tensorrt-llm-and-triton-on-amazon-eks/
- https://cloud.google.com/blog/products/ai-machine-learning/deploy-and-serve-open-models-over-google-kubernetes-engine?e=48754805 
- https://docs.nvidia.com/deeplearning/triton-inference-server/user-guide/docs/tutorials/Deployment/Kubernetes/EKS_Multinode_Triton_TRTLLM/README.html
- https://github.com/kubernetes-sigs/lws/issues/320 
- DaoCloud?

Also list talks:
- https://www.youtube.com/watch?v=Al51wafTrRE

yankay:
DaoCloud is OK to appear on the adoption page. :-)

kerthcet:
/assign

I'm working on this.

PR #367, opened 2025-02-08
Add adopters, integrations, talks

kerthcet:
#### What type of PR is this?

/kind cleanup

Fixes https://github.com/kubernetes-sigs/lws/issues/324

```release-note
None
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ Read the [installation guide](/docs/setup/install.md) to learn more.
 
 Read the [examples](/docs/examples/sample/README.md) to learn more.
 
+Also discover adopters, integrations, and talks [here](/docs/showcase.md).
+
 ## Community, discussion, contribution, and support
 
 Learn how to engage with the Kubernetes community on the [community page](http://kubernetes.io/community/).
--- a/docs/showcase.md
+++ b/docs/showcase.md
@@ -1,0 +1,25 @@
+# Adoption, Integration, and Presentation
+
+## Adopters
+
+*This is based on public documentations, please open an issue if you would like to be added or removed the list.*
+
+**AWS**: Amazon EKS supports to run *superpod* with LeaderWorkerSet (LWS) to server large LLMs, see blog [here](https://aws.amazon.com/blogs/hpc/scaling-your-llm-inference-workloads-multi-node-deployment-with-tensorrt-llm-and-triton-on-amazon-eks/).
+
+**DaoCloud**: LeaderWorkerSet (LWS) is the default deployment method to run large models crossing multiple nodes on Kubernetes.
+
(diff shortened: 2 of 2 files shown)


yankay:
There is a link in the LWS: https://docs.vllm.ai/en/latest/deployment/frameworks/lws.html
Does it mean adoption or integration?

kerthcet:
> There is a link in the LWS: https://docs.vllm.ai/en/latest/deployment/frameworks/lws.html Does it mean adoption or integration?

Thanks, I forgot to mention this, actually we're an inference engine agnostic framework, like we support llama.cpp, tensorrt-llm as well, so I just skipped the integration on purpose, but I can add them back if you all think worth to mention it.

We at least initialized the Dockerfile and shell script to integrate with these inference engines, from this point of view, we can add them as part of the `integrations`.

ahg-g:
I think we should be open to mentioning any related adoption or integration tutorials

README.md
```diff
@@ -33,6 +33,8 @@ Read the [installation guide](/docs/setup/install.md) to learn more.
 
 Read the [examples](/docs/examples/sample/README.md) to learn more.
 
+Also discover real-world adopters, integrations, and talks [here](/docs/showcase.md).
```
ahg-g, line 36:
```suggestion
Also discover adopters, integrations, and talks [here](/docs/showcase.md).
```

docs/showcase.md
```diff
@@ -0,0 +1,23 @@
+# Adoption, Integration, and Presentation
+
+## Adopters
+
+*This is based on the public documentations, contact us if you want to be added or removed.*
```
ahg-g, line 5:
```suggestion
*This is based on public documentations, please open an issue if you would like to be added or removed from the list.*
```

docs/showcase.md
```diff
@@ -0,0 +1,23 @@
...
+
+**Google Cloud**: GKE leverages LeaderWorkerSet (LWS) to deploy and serve multi-host gen AI large open models, see blog [here](https://cloud.google.com/blog/products/ai-machine-learning/deploy-and-serve-open-models-over-google-kubernetes-engine?e=48754805).
+
+**Nvidia**: LeaderWorkerSet (LWS) deployments are the recommended method for deploying Multi-Node models with NIM, see document [here](https://docs.nvidia.com/nim/large-language-models/1.5.0/deploy-helm.html#multi-node-models).
+
+## Integrations
+
+*Feel free to submit a PR if you use LeaderWorkerSet (LWS) in your project and want to be added here.*
+
+[**llmaz**](https://github.com/InftyAI/llmaz): llmaz, serving as an easy to use and advanced inference platform, uses LeaderWorkerSet (LWS) as the underlying workload to support both single-host and multi-host inference scenarios.
```
ahg-g, line 19: can we add the vllm one please?

ahg-g (commented):
Thanks!

docs/showcase.md
kerthcet: done

kerthcet:
Addressed the comments and squashed. PTAL.

PR #374, opened 2025-02-10
Add missing external types to apply configurations

astefanutti:
#### What type of PR is this?

/kind bug

/kind api-change

This PR adds the external types the LWS API depends on to the code-generator command so the generated client apply configurations are complete and do not rely on "non-apply" types.

```release-note
Add missing external types to apply configurations
```


ardaguclu:
That is a nice catch and I agree that generated codes for apply configuration needs this change (I saw similar PRs in Kueue and JobSet as well).

kerthcet (commented):
/lgtm
/approve

Thanks @astefanutti

Issue #223, opened 2024-09-14
Unique node selector and toleration per replica


ahg-g:
**What would you like to be added**:

Allow injecting a unique nodeSelector and toleration for each LWS replica to trigger cluster autoscaler to create a dedicated placement group for each replica. 

In the api, the user sets the key they would like to use, and the value would be the name of the replica (the leader pod name)

```
ReplicaSpecificNodeSelector: 
   - compact-placement-group
```

The result is a nodeSelector injected as follows:
```compact-placement-group: <lws-leader-name>```

Similarly for tolerations:
```
ReplicaSpecificToleration
      - key: compact-placement-group
        effect: NoSchedule
```

The result is a toleration injected on the pods of a group as follows:
```
      - key: group
        operator: Equal
        value: <lws-leader-name>
        effect: NoSchedule
```

**Why is this needed**:
To force cluster autoscaler to create a node group per replica, which can be necessary to create compactly placed nodes (on the same rack for example) for better network performance, and can improve multi-host GPU inference.

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

**Comment by googs1025** [comment]
I'm willing to give it a try.：）
/assign

Would it be better to provide a Google Docs first?

**Comment by googs1025** [comment]
Sorry, I don't quite understand how compact-placement-group is defined.

```yaml
apiVersion: leaderworkerset.x-k8s.io/v1
kind: LeaderWorkerSet
metadata:
  name: leaderworkerset-multi-template
spec:
  replicas: 3
  leaderWorkerTemplate:
    ReplicaSpecificNodeSelector: compact-placement-group
    ReplicaSpecificToleration:
      - key: compact-placement-group
    leaderTemplate:
      spec:
        containers:
          - name: nginx2
...
```

ahg-g:
`compact-placement-group` is a string that the user sets, and we use it as the key to a nodeSelector with a value equal to the leader pod name.

so the snippet you have for the api is correct, the outcome is that we inject a nodeSelector for each group as follows:

```
nodeSelector:
 - compact-placement-group: <leader-pod-name>
 ```

Edwinhr716:
@googs1025 Are you still looking to work on this item?

googs1025:
[...]

PR #399, opened 2025-02-25
Add KEP#223: Unique Node Selector and Toleration Per Replica

Edwinhr716:
#### What type of PR is this?
/kind documentation

Adds KEP for #223

Fixes #

```release-note

```

### DIFF
--- a/keps/223-unique-selector-per-replica/README.md
+++ b/keps/223-unique-selector-per-replica/README.md
@@ -1,0 +1,327 @@
+# KEP-223: Unique Node Selector and Toleration Per Replica
+
+<!--
+This is the title of your KEP. Keep it short, simple, and descriptive. A good
+title can help communicate what the KEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a KEP and for
+highlighting any additional information provided beyond the standard KEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1](#story-1)
+- [Design Details](#design-details)
+  - [Implementation](#implementation)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+<!--
+This section is incredibly important for producing high-quality, user-focused
+documentation such as release notes or a development roadmap. It should be
+possible to collect this information before implementation begins, in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself. KEP editors and SIG Docs
+should help to ensure that the tone and content of the `Summary` section is
+useful for a wide audience.
+
+A good summary is probably at least a paragraph in length.
+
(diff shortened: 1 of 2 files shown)


googs1025:
/cc 

I'll be able to give some of my thoughts later this week :)

keps/223-unique-selector-per-replica/README.md
```diff
@@ -0,0 +1,327 @@
...
+We propose adding two new API fields to LWS:
+
+```golang
+type LeaderWorkerSetSpec {
+    // Defines the label that will be targeted by the node selector
+    // The key will be the name of the leader pod
+	ReplicaUniqueNodeSelector string `json:"replicaUniqueNodeSelector,omitempty"`
+
+    // Defines the toleration configuration that will be unique per replica
+	ReplicaUniqueToleration *ReplicaUniqueToleration `json:"replicaUniqueToleration,omitempty"`
```
ahg-g, line 146: I would put this in the `LeaderWorkerTemplate`, and encapsulate it in a struct named `SchedulingPolicy`

keps/223-unique-selector-per-replica/README.md
```diff
@@ -0,0 +1,327 @@
...
+The leader will be scheduled on the existing nodeGroup, however, the workers won't be able to be scheduled on it because of the unique node selector.
+Therefore, to ensure compatability with the exclusive placement feature, both the leader and worker pods will have the unique node selector and toleration.
+
+We can only add the node selector and toleration after the leader pod has been created, so injecting them will be handled by the pod webhook. The pod webhook does
+not have access to the LWS object, so three new annotations, corresponding to the new configurable fields, will be added:
+
+* `leaderworkerset.sigs.k8s.io/replicaUniqueNodeSelector` 
+* `leaderworkerset.sigs.k8s.io/replicaUniqueTolerationKey`
+* `leaderworkerset.sigs.k8s.io/replicaUniqueTolerationEffect`
+
```
ahg-g, line 172: clarify that we will also add them to the workers sts

keps/223-unique-selector-per-replica/README.md
Edwinhr716: updated

keps/223-unique-selector-per-replica/README.md
ahg-g: correct, the pod webhook should ideally only deal with the leaders

kerthcet:
Since we'll release this week and I think the implementation hasn't been ready yet, can we merge this later, I'll take another look. Does that make sense to you?

Edwinhr716:
That makes sense, we can include it in the next release. Not an urgent feature either way

keps/223-unique-selector-per-replica/README.md
```diff
@@ -0,0 +1,334 @@
...
+## Design Details
+
+We propose adding two new API fields to LWS:
+
+```golang
+
+type type LeaderWorkerTemplate struct {
+  // Defines additional scheduling constraints for the group
+  SchedulingPolicy *SchedulingPolicy `json:"schedulingPolicy,omitempty"`
```
ahg-g, line 143: I feel we need to make this more automated, the user only sets a topology-key and we just inject the node-selector and toleration. It feels like we can piggyback on exclusive-placement annotation, but we need a knob to select between pod affintity/anti-affinity and nodeSelector/taint, wdyt?

We could start with another annotation, and then we migrate this into a proper exclusive placement api.

Edwinhr716: >  It feels like we can piggyback on exclusive-placement annotation, but we need a knob to select between pod affintity/anti-affinity and nodeSelector/taint

So use `leaderworkerset.sigs.k8s.io/exclusive-topology` and have in the Scheduling Policy whether it is a pod affinitity or a nodeSelector/taint? That sounds good to me, I can't think of a case where we want both exclusive placement and a create a node group per replica. 

> We could start with another annotation, and then we migrate this into a proper exclusive placement api

I would prefer to have a proper exclusive placement API right off the bat. I don't think it would add much complexity implementation wise, and we are targeting July (0.7.0) anyway. Is there a benefit to adding another annotation instead?

ahg-g: Lets propose one then, we need to continue to support the annotation though until users migrate off of it

Edwinhr716:
Added `ExclusivePlacement` to the `SchedulingPolicy` part of the API. PTAL @ahg-g

Edwinhr716:
@ahg-g plan to close this since it will be addressed by the gang-scheduling feature in upstream. Any concerns?

ahg-g:
make sense

Issue #257, opened 2024-11-13
support leader being in its own subgroup


avrittrohwer:
**What would you like to be added**:

Add the ability for the leader Pod to be in its own affinity group when using the subgroup feature.  For example, when deploying a leader Pod that should be scheduled on a CPU-only VM and worker Pods that should be scheduled on multiple TPU slices:

```
apiVersion: leaderworkerset.x-k8s.io/v1
kind: LeaderWorkerSet
metadata:
  name: my-lws
  annotations:
    leaderworkerset.sigs.k8s.io/subgroup-exclusive-topology: cloud.google.com/gke-nodepool
spec:
  replicas: 1
  leaderWorkerTemplate:
    subGroupPolicy:
      subGroupSize: 2
    size: 5
    leaderTemplate:
      spec:
        nodeSelector:
          cloud.google.com/machine-familty: n2
          node.kubernetes.io/instance-type: n2-standard-8
        containers:
        - name: leader
        ...
    workerTemplate:
      spec:
        nodeSelector:
          cloud.google.com/gke-tpu-accelerator: tpu-v5p-slice
          cloud.google.com/gke-tpu-topology: 2x2x2
        containers:
        - name: worker
          ...
          resources:
            limits:
              google.com/tpu: "4"
```

Currently the leader Pod is put in subgroup 0 which causes it to have the same affinity key as the workers in subgroup 0: https://github.com/kubernetes-sigs/lws/blob/main/pkg/webhooks/pod_webhook.go#L132.  This causes the leader Pod in my example to be unscheduable because of the CPU instance type node selectors.

**Why is this needed**:

To support deploying leader-worker architectures where the leader should be scheduled in separate topologies from the worker groups.

**Completion requirements**:

An option in `subGroupPolicy` that causes the leader to have its own affinity key.

This enhancement requires the following artifacts:

- [ ] Design doc
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

avrittrohwer:
@ahg-g

ahg-g:
@Edwinhr716

Edwinhr716:
@avrittrohwer do we still want to support subgroups within the workers? So for example, if `subGroupPolicy=leaderOnly`, `size=5` and `subgroupSize=2`, make it so that the resulting groups would be (`leader`), (`worker-1,worker-2`), (`worker-3, worker-4`)

kerthcet:
[...]

PR #402, opened 2025-02-27
Add KEP #257: Leader Only SubGroup

Edwinhr716:
#### What type of PR is this?
/kind documentation

Part of #257 

```release-note

```


kerthcet:
I may review this tomorrow because of limited bandwidth, sorry for that. 
/assign

kerthcet:
reviewing now.

keps/257-Subgroup-leader-only/README.md
```diff
@@ -0,0 +1,353 @@
...
+
+	SubGroupPolicyLeaderOnly SubGroupPolicyType = "LeaderOnly"
+)
+```
+
+A new annotation will be created to determine whether or not the subgroup type is LeaderOnly
+
+* `leaderworkerset.sigs.k8s.io/subgroup-policy-type`
+
+In order to keep backwards compatability, it will only be added if the type is `LeaderOnly`.
```
kerthcet, line 185: I think even we add the annotation for both types is still backward compatibility. No behavior change.

keps/257-Subgroup-leader-only/README.md
```diff
@@ -0,0 +1,353 @@
...
+
+```golang
+start := subGroupSize*subGroupIndex + 1
+end := subGroupSize * (subGroupIndex + 1)
+for i := start; i <= end; i++ {
+    hostnames = append(hostnames, fmt.Sprintf("%s-%d.%s", leaderName, i, pod.Spec.Subdomain))
+}
+```
+
+Because the leader now occupies a subgroup, and `SubGroupSize` will not always be one, the way we calculate the SubGroupIndex needs to be modified. When generating
```
kerthcet, line 207: One argue here is should we mark the leader a sub group? What's the Cons if we don't do this.

kerthcet (commented):
several comments.

keps/257-Subgroup-leader-only/README.md
Edwinhr716: The scenario that I see is that there is an existing deployment using `SubGroupPolicy` which then upgrades the LWS controller. After the upgrade, it will inject the annotation on the PodSpec, which will then trigger a rolling update since the Pod has changed.

keps/257-Subgroup-leader-only/README.md
```diff
@@ -0,0 +1,353 @@
...
+List the specific goals of the KEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+Add an option to create a LeaderOnly subgroup, while still being able to create subgroups on the other workers. Therefore, it should be possible to do the following split:
+(leader), (worker-1, worker-2), (worker-3, worker-4). If the desired effect is to have one subgroup for the leader, and one subgroup for the worker, subgroupSize should be set 
+to the number of workers (so size - 1). 
+
+### Non-Goals
+This KEP assumes that the leader will not request TPU resources, and thus, subGroupSize will always be assumed to be an odd number. Moreover, no TPU environment variables will be
```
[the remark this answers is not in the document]
Edwinhr716: Both TPU and GPU are supported for exclusive placement. I'll rephrase to be accelerator agnostic

keps/257-Subgroup-leader-only/README.md
```diff
@@ -0,0 +1,353 @@
...
+
+    SubGroupSize *int32 `json:"subGroupSize,omitempty"`
+}
+
+type SubGroupPolicyType string
+
+const (
+	SubGroupPolicyLeaderWorker SubGroupPolicyType = "LeaderWorker"
+
+	SubGroupPolicyLeaderOnly SubGroupPolicyType = "LeaderOnly"
```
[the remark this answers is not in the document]
Edwinhr716: I'm fine with `LeaderAlone`, but `All` sounds a bit ambiguous to me. `LeaderAlone` vs `LeaderWorker` maybe? Or something that specifies that the leader and a worker pod will be part of the same subgroup

keps/257-Subgroup-leader-only/README.md
Edwinhr716: That's a good point, hadn't considered not making the leader part of a subgroup at all. Doing it this way simplifies the rest of the changes (though the TPU environment injection will still need some adjustment). Will test it out and add to the `Alternatives` which ever version we don't implement

Edwinhr716:
Changed implementation to the leader not actually being part of any subgroup since it is simpler, added original implementation to Alternatives. Also changed LeaderOnly to LeaderAlone. Please take another look @kerthcet

keps/257-Subgroup-leader-only/README.md
```diff
@@ -0,0 +1,356 @@
...
+List the specific goals of the KEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+Add an option to create a LeaderOnly subgroup, while still being able to create subgroups on the other workers. Therefore, it should be possible to do the following split:
+(leader), (worker-1, worker-2), (worker-3, worker-4). If the desired effect is to have one subgroup for the leader, and one subgroup for the worker, subgroupSize should be set 
+to the number of workers (so size - 1). 
+
+### Non-Goals
+This KEP assumes that the leader will not request TPU resources, and thus, subGroupSize will always be assumed to be an odd number. Moreover, no TPU environment variables will be
```
ahg-g, line 94: pls rephrase to be accelerator agnostic, this applies to both

keps/257-Subgroup-leader-only/README.md
```diff
@@ -0,0 +1,356 @@
...
+
+<!--
+Detail the things that people will be able to do if this KEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
+As a user, I should be able to create a subGroup that only includes the leader in order to be able to schedule the leader on a different topology while
```
ahg-g, line 123:
```suggestion
As a user, I should be able to create a subGroup that only includes the leader in order to be able to schedule the leader with different resource requirements
```

keps/257-Subgroup-leader-only/README.md
```diff
@@ -0,0 +1,356 @@
...
+A new API field will be added under the SubGroupPolicy type. This API field will determine what type 
+of subgroup is desired. The default is LeaderWorker, which is the existing implementation of subgroup. `SubGroupSize`
+will continue to be a required field, while `Type` will be optional. 
+
+```golang
+type SubGroupPolicy struct {
+
+	// +kubebuilder:validation:Enum={LeaderWorker,LeaderAlone}
+	// +kubebuilder:default=LeaderWorker
+	Type SubGroupPolicyType `json:"subGroupPolicyType,omitempty"`
```
ahg-g, line 168: fix indentation pls

keps/257-Subgroup-leader-only/README.md
```diff
@@ -0,0 +1,356 @@
...
+A new annotation will be created to determine whether or not the subgroup type is LeaderAlone
+
+* `leaderworkerset.sigs.k8s.io/subgroup-policy-type`
+
+This annotation will only be injected in the leader pod.
+
+In order to keep backwards compatability, it will only be added if the type is `LeaderAlone`.
+
+### Subgroup Creation 
+Implementation wise, the only change needed is to not add the SubGroup labels on the leader if the SubGroupType is LeaderAlone. Effectively, this means
```
ahg-g, line 191: The indexing within the group will be different now. Currently we don't define an index within the group, we do it only for tpuWorkerId, and this should be calculated differently now to exclude the leader: https://github.com/kubernetes-sigs/lws/blob/main/pkg/utils/accelerators/tpu.go#L113, right?

keps/257-Subgroup-leader-only/README.md
```diff
@@ -0,0 +1,356 @@
...
+
+* `leaderworkerset.sigs.k8s.io/subgroup-policy-type`
+
+This annotation will only be injected in the leader pod.
+
+In order to keep backwards compatability, it will only be added if the type is `LeaderAlone`.
+
+### Subgroup Creation 
+Implementation wise, the only change needed is to not add the SubGroup labels on the leader if the SubGroupType is LeaderAlone. Effectively, this means 
+that the leader is not part of a subgroup at all. The only point of a pod being in a subgroup is to guarantee exclusive placement with the other pods
```
kerthcet, line 192: > the leader is not part of a subgroup at all

LeaderAlone is based on the original assumption that leader pod is a separate group, if not, maybe we should rename it to like `leaderExcluded`, in the future, leaderAlone can still be reuse if leader pod should be the 0-index subgroup.

kerthcet (commented):
I think we can leave more implementation details to the PR.

keps/257-Subgroup-leader-only/README.md
Edwinhr716: We already do this when the leader doesn't request TPU resources https://github.com/kubernetes-sigs/lws/blob/main/pkg/utils/accelerators/tpu.go#L115-L117

keps/257-Subgroup-leader-only/README.md
Edwinhr716: Makes sense to me, will change it to `leaderExcluded`. Does `leaderWorker` then sound good to you for the default setup?

keps/257-Subgroup-leader-only/README.md
```diff
@@ -0,0 +1,355 @@
...
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+This KEP aims to extend the SubGroup feature by adding an API field that defines the type of Subgroup that will be created. The existing version
+will become "LeaderWorker", and a new type will be added to create a SubGroup that only includes the leader.
```
kerthcet, line 64: There's no subgroup for leader pod now.

keps/257-Subgroup-leader-only/README.md
```diff
@@ -0,0 +1,355 @@
...
+This section is for explicitly listing the motivation, goals, and non-goals of
+this KEP.  Describe why the change is important and the benefits to users. The
+motivation section can optionally provide links to [experience reports] to
+demonstrate the interest in a KEP within the wider Kubernetes community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+There are cases where the resource requirements for the leader are different from the ones by the worker, while still being able to guarantee exclusive placement
+on the workers. Having a leaderOnly subgroup means that the placement of the leader will be independent of the workers, and by adding the `subgroup-exclusive-topology`, the
```
kerthcet, line 79: There's no leaderOnly type as well.

keps/257-Subgroup-leader-only/README.md
```diff
@@ -0,0 +1,355 @@
...
+workers can be guaranteed to be scheduled on the same topology.
+
+### Goals
+
+<!--
+List the specific goals of the KEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+Add an option to create a LeaderOnly subgroup, while still being able to create subgroups on the other workers. Therefore, it should be possible to do the following split:
```
kerthcet, line 89: ditto

keps/257-Subgroup-leader-only/README.md
kerthcet: LGTM

kerthcet (commented):
I think the calculation of worker subGroup index still needs to change, for example, we set the subgroupSize=2,
- before with group size = 6, it looks like (0, 0), (1, 0), (2, 1), (3, 1), (4, 2), (5 ,2)
- after with group size = 7, it looks like (0, x), (1, 0), (2, 0), (3, 1), (4, 1), (5, 2), (6, 2) 

(x, y) represents the (workerIndex, subGroupIndex), for the workerIndex=2, before the subgroupIndex is 1, after is 0. Not familiar with the TPU implementation, maybe the same? But this is just code implementation, we can leave the detailed review to the PR, just mention this fact in the proposal briefly.

keps/257-Subgroup-leader-only/README.md
Edwinhr716: Right, updated this and all other references to leader being its own subgroup

kerthcet:
Yes, the result is as we expected, I just mean the method may need to change: https://github.com/kubernetes-sigs/lws/blob/ab40b6285254c4fae69e605578d95f456b64e09f/pkg/webhooks/pod_webhook.go#L245-L251

For workIndex=2, before it's (2, 1), now it's (2, 0), the subGroupIndex changes. Do I miss anything? I think we'll call the method, right?

Edwinhr716:
Yeah we call this method. I guess I'm a little bit confused on what scenario you are talking about. The current behavior when using this feature has workerIndex=2 at subgroup 0. If using the default configuration, but still having an odd number for size, it will still be workerIndex=2 at subgroup 0. The only case where workerIndex=2 gives subgroup 1, is if using default behavior and size is even

PR #406, opened 2025-03-04
Emphasize the version requirement

kerthcet:
#### What type of PR is this?

/kind cleanup

We got several feedbacks about version lower than v1.26 and lws behaves unexpected. Let's emphasize the version requirement in installation documentation.

Fixes #

```release-note
None
```

### DIFF
--- a/docs/setup/install.md
+++ b/docs/setup/install.md
@@ -4,8 +4,8 @@
 
 Make sure the following conditions are met:
 
-- A Kubernetes cluster with version 1.26 or newer is running. Learn how to [install the Kubernetes tools](https://kubernetes.io/docs/tasks/tools/).
-    - For any cluster with version lower than 1.27, you need to enable the [feature gate][feature_gate] for [Start Ordinal][start_ordinal]
+- A Kubernetes cluster with version >= 1.26 is **Required**, or it will behave unexpected. Learn how to [install the Kubernetes tools](https://kubernetes.io/docs/tasks/tools/).
+    - For any cluster with version 1.26, you need to enable the [feature gate][feature_gate] for [Start Ordinal][start_ordinal] manually. For version greater than 1.26, it's enabled by default.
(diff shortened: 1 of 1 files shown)


docs/setup/install.md
```diff
@@ -4,8 +4,8 @@
 
 Make sure the following conditions are met:
 
-- A Kubernetes cluster with version 1.26 or newer is running. Learn how to [install the Kubernetes tools](https://kubernetes.io/docs/tasks/tools/).
-    - For any cluster with version lower than 1.27, you need to enable the [feature gate][feature_gate] for [Start Ordinal][start_ordinal]
+- A Kubernetes cluster with version 1.26 or newer is **Required**. Learn how to [install the Kubernetes tools](https://kubernetes.io/docs/tasks/tools/).
```
kannon92, line 7: Anyway to fail lws on k8s < 1.26?

kerthcet: Kevin, is that a suggestion or a question? FYI: startOrdinal was introduced in v1.26.

kannon92: It is a question.

Edwinhr716: Current behavior is that LWS fails to create the worker sts if version < 1.26. as seen here https://github.com/kubernetes-sigs/lws/issues/401#issuecomment-2694989196.

kannon92: Yea. If we wanted to "emphasize" it and make sure it is not supported.

kerthcet: Yes, panic the controller could be one possible way, but I requires additional logic like judging the version as well as the feature gate enabled or not, I prefer the light way. I think it's aligned with other sig projects.

kerthcet: And event will be emitted here once version < 1.26. https://github.com/kubernetes-sigs/lws/blob/86c206f11b11fb5f532d9d4197ef2ae76310879b/pkg/controllers/pod_controller.go#L109

kerthcet:
If good to you, anyone take a LGTM. @Edwinhr716 @kannon92

kannon92 (commented):
/lgtm

PR #417, opened 2025-03-10
Run controller-gen only for api and pkg directories

ardaguclu:
#### What type of PR is this?
/kind cleanup

There are go.mods under `docs/examples/llamacpp/tools/` path (e.g. `docs/examples/llamacpp/tools/blobserver/go.mod`) which are not directly related to the actual code block and they present only as an examples. However, controller-gen executes also for them unnecessarily. 

This PR modifies controller-gen's path to only execute `api` and `pkg` to align with jobset https://github.com/kubernetes-sigs/jobset/blob/e8702b95e33c38d8a856ab63b14ffec5abd5e22a/Makefile#L90

Additionally, this PR updates to use branch name instead of commit hash in setupenv-test binary for better readability.

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,7 +99,7 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 		rbac:roleName=manager-role output:rbac:artifacts:config=config/rbac \
 		crd:generateEmbeddedObjectMeta=true output:crd:artifacts:config=config/crd/bases \
 		webhook output:webhook:artifacts:config=config/webhook \
-		paths="./..."
+		paths="{./api/..., ./pkg/...}"
 
 .PHONY: generate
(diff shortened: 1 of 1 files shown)


Edwinhr716:
I see that JobSet also changed the `generate` command to only use `./api` https://github.com/kubernetes-sigs/jobset/pull/470. Can we do the same?

ardaguclu:
> I see that JobSet also changed the `generate` command to only use `./api` [kubernetes-sigs/jobset#470](https://github.com/kubernetes-sigs/jobset/pull/470). Can we do the same?

@Edwinhr716 thank you for your review. Updated based on your advise.

ardaguclu:
These jobs failed in another PR too. May not be related to this PR.
/retest

kerthcet (commented):
/lgtm
/approve

Thanks!

PR #422, opened 2025-03-10
Adding dependencies for push-chart.sh to Makefile

Edwinhr716:
#### What type of PR is this?
/kind bug

https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-lws-push-images/1899147188785647616 current image push fails due to `yq` not in directory. Added dependencies to match JobSet 

Fixes #

```release-note

```


Edwinhr716:
@kannon92 Sorry for all the requests. Tested it out locally, only step failing on my end was 403 error, so should work in the pipeline now

PR #428, opened 2025-03-13
Implementation KEP 257: LeaderExcluded SubGroup

### LINKED ISSUE #257: support leader being in its own subgroup

Edwinhr716:
#### What type of PR is this?
/kind feature

Adds a new type of subgroup to exclude the leader. Can provide exclusive placement guarantees in the workers even if leader has different hardware requirements

Fixes #257 

```release-note

```


pkg/webhooks/pod_webhook_test.go
```diff
@@ -268,3 +268,38 @@ func TestExclusiveAffinityApplied(t *testing.T) {
...
 }
+
+func TestGetSubGroupIndex(t *testing.T) {
+	tests := []struct {
+		name          string
+		podCount      int
+		subGroupSize  int
+		workerIndex   int
+		leaderOnly    bool
+		expectedIndex string
```
kerthcet, line 279: better rename to expectedSubGroupIndex, we have a lot of index here.

kerthcet (commented):
Does the TPU part good to you @ahg-g otherwise, LGTM.

pkg/webhooks/pod_webhook.go
```diff
@@ -126,7 +126,8 @@ func (p *PodWebhook) Default(ctx context.Context, obj runtime.Object) error {
 			SetExclusiveAffinities(pod, groupUniqueKey, epKey, leaderworkerset.GroupUniqueHashLabelKey)
 		}
 		_, foundSubGroupSize := pod.Annotations[leaderworkerset.SubGroupSizeAnnotationKey]
-		if foundSubGroupSize && pod.Labels[leaderworkerset.SubGroupIndexLabelKey] == "" {
+		_, leaderExcludedSubGroup := pod.Annotations[leaderworkerset.SubGroupPolicyTypeAnnotationKey]
```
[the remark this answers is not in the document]
Edwinhr716: We don't have an annotation for `subGroupType` if it is `LeaderWorker`. Although if we are breaking backwards compatibility with #434 anyway, I think we can reconsidering adding a label even if it is `LeaderWorker`

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -188,6 +201,25 @@ type SubGroupPolicy struct {
...
+	// LeaderWorker will include the leader in the first subgroup.
+	// If (LeaderWorkerSet.Spec.LeaderWorkerTemplate.Size-1) is divisible
+	// by LeaderWorkerSet.Spec.SubGroupPolicy.Size the groups will look like:
+	// (0, 1, ... subGroupSize), (subGroupSize + 1, ... 2 * subGroupSize), ...
+	// If it is even, the groups will look like:
+	// (0, 1, ... subGroupSize-1), (subGroupSize, ... 2*subGroupSize - 1), ...
+	SubGroupPolicyTypeLeaderWorker SubGroupPolicyType = "LeaderWorker"
+
+	// LeaderExcluded excludes the leader from any subgroup. Only
+	// supported if If (LeaderWorkerSet.Spec.LeaderWorkerTemplate.Size-1) is divisible
```
kerthcet, line 216:
```suggestion
	// LeaderExcluded excludes the leader from any subgroup. 
	// Only supported when (LeaderWorkerSet.Spec.LeaderWorkerTemplate.Size-1) is divisible
```

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -82,6 +82,13 @@ func (r *LeaderWorkerSetWebhook) Default(ctx context.Context, obj runtime.Object
 		subdomainPolicy := v1.SubdomainShared
 		lws.Spec.NetworkConfig.SubdomainPolicy = &subdomainPolicy
 	}
+
+	if lws.Spec.LeaderWorkerTemplate.SubGroupPolicy != nil {
```
kerthcet, line 86: I think we can totally remove this now.

kerthcet (commented):
just nits.

pkg/webhooks/pod_webhook.go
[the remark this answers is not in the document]
kerthcet: ok, it's all about subgroup.

pkg/webhooks/leaderworkerset_webhook.go
kerthcet: I mean with the kubebuilder default tag, we'll set the value by default once subGroupPolicy != nil.

kerthcet:
Seems the test is crashing ...

pkg/webhooks/leaderworkerset_webhook.go
Edwinhr716: Removed, same with the integration test

kerthcet:
/lgtm

I think we need to update the https://lws.sigs.k8s.io/docs/reference/labels-annotations-and-environment-variables/ as well, but could be a follow-up.

Thanks for this.

PR #431, opened 2025-03-14
Add Labels, Annotations and Environment Variables to the docs

yankay:
…n page.

#### What type of PR is this?

/kind documentation

#### What this PR does / why we need it

Add Labels, Annotations and Environment Variables to the docs.
Fix some comments in code that are incorrect.

#### Which issue(s) this PR fixes

Fixes #412

#### Does this PR introduce a user-facing change?

```release-note

```

PR #434, opened 2025-03-14
Change annotation `leaderworkerset.gke.io/subgroup-size` to `leaderworkerset.sigs.k8s.io/subgroup-size`

### LINKED ISSUE #430: The label named `leaderworkerset.gke.io/subgroup-size` is suffix by the `gke.io` not `sigs.k8s.io`

The label named `leaderworkerset.gke.io/subgroup-size` is suffixed by the `gke.io` not `sigs.k8s.io`, ref to https://github.com/kubernetes-sigs/lws/blob/main/api/leaderworkerset/v1/leaderworkerset_types.go#L82
Are there any special reasons ?

**Why is this needed**:

All the annotations are suffixed by `sigs.k8s.io` so it's so strange.

yankay:
…sigs.k8s.io/subgroup-size`

#### What type of PR is this?

/kind cleanup

From the KEP: https://github.com/kubernetes-sigs/lws/blob/main/keps/115-Subgroup-support/README.md?plain=1#L144,
the annotation needs to be `leaderworkerset.sigs.k8s.io/subgroup-size`, 
So change annotation `leaderworkerset.gke.io/subgroup-size` to `leaderworkerset.sigs.k8s.io/subgroup-size`.

And needs to change the docs if https://github.com/kubernetes-sigs/lws/pull/431 is merged.

The label should be more consistent with others.

Fixes #430

```release-note
Change annotation  `leaderworkerset.gke.io/subgroup-size` to `leaderworkerset.sigs.k8s.io/subgroup-size`
```

### DIFF
--- a/api/leaderworkerset/v1/leaderworkerset_types.go
+++ b/api/leaderworkerset/v1/leaderworkerset_types.go
@@ -79,7 +79,7 @@ const (
 
 	// SubGroupSize will be added to pods as an annotation which corresponds to
 	// LeaderWorkerSet.Spec.SubGroupPolicy.SubGroupSize
-	SubGroupSizeAnnotationKey string = "leaderworkerset.gke.io/subgroup-size"
+	SubGroupSizeAnnotationKey string = "leaderworkerset.sigs.k8s.io/subgroup-size"
 
 	// Pods that are part of the same subgroup will have the same unique hash value.
 	SubGroupUniqueHashLabelKey string = "leaderworkerset.sigs.k8s.io/subgroup-key"


Edwinhr716:
We need to keep backwards compatibility here, keeping the `gke.io` on deployments that already have that label on them. Otherwise, it triggers a rolling update at the leader pod level as soon as the controller version is updated, since the label on the pods will be changed. 

@ahg-g @kerthcet thoughts?

ahg-g:
oh, what a mistake! This will impact deployments that uses subgroup policy, which I don't think there are any right now in production. I am ok with this change.

Edwinhr716:
Sounds good from me then, but I think we should add documentation somewhere that it might be a potential issue

ahg-g:
There is a release note

Issue #437, opened 2025-03-17
[flaky] pull-lws-test-e2e-main-1-30 timeout


kerthcet:
see https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_lws/431/pull-lws-test-e2e-main-1-30/1901511741909504000

ardaguclu:
I think, there is permanent failure. I'm looking too.

ardaguclu:
Latest tag of this gcr.io/k8s-staging-lws/lws:latest does not work.

Edwinhr716:
Jobset has the same issue https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_jobset/838/pull-jobset-test-e2e-main-1-32/1901628451081687040, could be something on the Kubernetes Infra side

kannon92:
@Edwinhr716 I don't think this is related.

My failure is consistent on a dependency update for 0.32.3 for 1.32..

Edwinhr716:
1.31 is failing as well due to time out https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_lws/445/pull-lws-test-e2e-main-1-31/1902041411415445504

ardaguclu:
https://github.com/kubernetes-sigs/lws/pull/451 have merged. If you see any failures, please paste the link in here and we can investigate the logs.

ardaguclu:
I checked the latest CI runs and couldn't find any flaky one. I think, we can close this issue and investigate in a new issue.
/close

Edwinhr716:
]https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_lws/509/pull-lws-test-e2e-main-1-31/1916327525068312576 e2e tests seem to still be flaky

PR #451, opened 2025-03-19
Export kind and cluster related logs under artifacts directory for e2e

ardaguclu:
#### What type of PR is this?
/kind cleanup

While investigating https://github.com/kubernetes-sigs/lws/issues/437, I realized that there is almost no logs making the troubleshooting easier (e.g. https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_lws/445/pull-lws-test-e2e-main-1-31/1902041411415445504/artifacts/). 

This PR exporting kind and cluster related logs under artifacts directory to troubleshoot any issues including the flaky one that I'm looking right now.

This PR mimics from jobset https://github.com/kubernetes-sigs/jobset/blob/15adf20ce9e0a10f01e7bcd8ea585a011814a2aa/hack/e2e-test.sh#L14-L17

```release-note
None
```

### DIFF
--- a/hack/e2e-test.sh
+++ b/hack/e2e-test.sh
@@ -22,6 +22,12 @@ export CWD=$(pwd)
 function cleanup {
     if [ $USE_EXISTING_CLUSTER == 'false' ]
     then
+        if [ ! -d "$ARTIFACTS" ]; then
+            mkdir -p "$ARTIFACTS"
+        fi
+        $KUBECTL logs -n lws-system deployment/lws-controller-manager > "$ARTIFACTS"/lws-controller-manager.log || true
+        $KUBECTL describe pods -n lws-system > "$ARTIFACTS"/lws-system-pods.log || true
+        $KIND export logs "$ARTIFACTS" || true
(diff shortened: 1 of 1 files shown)


ardaguclu:
Logs https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_lws/451/pull-lws-test-e2e-main-1-30/1902247411548426240/artifacts/ are collected correctly.

This PR is ready for review.

kerthcet (commented):
I'm thinking whether we should make it a flag rather than exposing logs all the time, but let's find the root cause first.
/lgtm
/approve

Thanks!

ardaguclu:
> I'm thinking whether we should make it a flag rather than exposing logs all the time, but let's find the root cause first. /lgtm /approve
> 
> Thanks!

That is a good point. What about we collect the logs if the exit code of ginkgo execution is non-zero?. That would be better in my opinion and I can open a followup PR.

PR #457, opened 2025-03-20
Set readOnlyRootFilesystem explicitly to true

ardaguclu:
#### What type of PR is this?
/kind cleanup

This PR sets `readOnlyRootFilesystem` to true. This is important aspect to prevent any writes in root file system. Besides, LWS does not need to write root file system at all.

```release-note
Setting the pods readOnlyRootFilesystem to true
```

### DIFF
--- a/charts/lws/values.yaml
+++ b/charts/lws/values.yaml
@@ -23,6 +23,7 @@ podSecurityContext:
   # fsGroup: 2000
 securityContext:
   allowPrivilegeEscalation: false
+  readOnlyRootFilesystem: true
   capabilities:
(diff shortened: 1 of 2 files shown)


yankay:
I'm not sure whether we should change the Helm Chart https://github.com/kubernetes-sigs/lws/blob/main/charts/lws/templates/manager/deployment.yaml#L41, together.

ardaguclu:
Nice catch and you are right. Updated.

kerthcet (commented):
/lgtm
/approve

Thanks!

PR #469, opened 2025-03-25
Remove node.js version in `netlify.toml` for site

yankay:
#### What type of PR is this?

/kind cleanup

Because Node.JS 16 is endoflife : https://endoflife.date/nodejs , and https://github.com/kubernetes-sigs/lws/pull/465 needs node.js >= 18 .
Remove node.js version in `netlify.toml` for the site. So it can be automatically up to date by netlify.

Fixes #

```release-note
Remove node.js version in `netlify.toml` for the site.
```

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -5,13 +5,10 @@
 
 [context.deploy-preview.environment]
   HUGO_VERSION = "0.92.0"
-  NODE_VERSION = "16"
 
 [context.production.environment]
   HUGO_VERSION = "0.92.0"
   HUGO_ENV = "production"
-  NODE_VERSION = "16"
 
 [context.branch-deploy.environment]
   HUGO_VERSION = "0.92.0"
-  NODE_VERSION = "16"


Edwinhr716:
Could you test what happens if we remove `NODE_VERSION`? Kueue doesn't have it https://github.com/kubernetes-sigs/kueue/blob/main/netlify.toml. Might be better for the long run so that we don't have to keep upgrading it

yankay:
> Could you test what happens if we remove `NODE_VERSION`? Kueue doesn't have it https://github.com/kubernetes-sigs/kueue/blob/main/netlify.toml. Might be better for the long run so that we don't have to keep upgrading it

Thanks @Edwinhr716 
Greate idea, it works.

Edwinhr716:
Does it use the latest Node version? Or what does it default to?

Edwinhr716:
Found the version in the logs

```
Now using node v22.14.0 (npm v10.9.2)
```

So it does use the latest. 

/lgtm
/approve

Edwinhr716:
@kannon92 FYI, JobSet also has `NODE_VERSION=16`

Issue #476, opened 2025-03-28
Golang version selection


rphillips:
**What happened**:
https://github.com/kubernetes-sigs/lws/pull/452/files merged to bump golang to 1.24. However, the project should likely still build with golang 1.22. The go.mod go stanza is the minimally viable golang version supported to build the project. Since golang 1.22 still builds the project, then the only bump should be in the Dockerfile and not in the go.mod files.

Thoughts?

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:

- Kubernetes version (use `kubectl version`):
- LWS version (use `git describe --tags --dirty --always`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

yankay:
> **What happened**: https://github.com/kubernetes-sigs/lws/pull/452/files merged to bump golang to 1.24. However, the project should likely still build with golang 1.22. The go.mod go stanza is the minimally viable golang version supported to build the project. Since golang 1.22 still builds the project, then the only bump should be in the Dockerfile and not in the go.mod files.
> 
> Thoughts?
> 
> **What you expected to happen**:
> 
> **How to reproduce it (as minimally and precisely as possible)**:
> 
> **Anything else we need to know?**:
> 
> **Environment**:
> 
> * Kubernetes version (use `kubectl version`):
> * LWS version (use `git describe --tags --dirty --always`):
> * Cloud provider or hardware configuration:
> * OS (e.g: `cat /etc/os-release`):
> * Kernel (e.g. `uname -a`):
> * Install tools:
> * Others:

We're upgrading Go to stay aligned with upstream requirements: https://github.com/kubernetes/release/issues/3772 .
Would this cause any issues for you? Could you please share more details about your use case? :-)

rphillips:
Upgrading Go's builder image is perfectly fine. If you note in https://github.com/kubernetes/release/issues/3772, I don't see anywhere where go.mod files are being updated. go.mod's `go` stanza is stating what is the minimum viable version to use to build the project. Golang is usually backwards compatible when bumping versions, like going from 1.22 to 1.24. In other words, bumping the go.mod version of golang to 1.24 is saying the project's source code is using features within golang 1.24.

[...]

PR #477, opened 2025-03-28
set golang to the minimum viable version of go

rphillips:
Related to https://github.com/kubernetes-sigs/lws/issues/476

go.mod's go stanza is tracking the minimum viable version of go to build the project. Since we are not using any Go 1.24 language features yet, the go.mod declarations should not be bumped.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/lws
 
-go 1.24.0
+go 1.23.0
 
 require (
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/lws
 
-go 1.24.0
+go 1.23.0
```
ardaguclu, line 3: I agree with you. I think it is better to bump this to when the k8s dependencies are bumped to 1.33. But I'll defer the decision to owners.

ahg-g: I am ok with aligning with the upstreaming k8s if it doesn't break anything

PR #478, opened 2025-03-28
Inject certs for metrics endpoint when cert-manager is enabled

ardaguclu:
#### What type of PR is this?
/kind feature

ServiceMonitor resource is configured to serve TLS with `insecureSkipVerify: true` . However, this is not recommended for production use cases. This PR injects the appropriate certificates to the ServiceMonitor to serve from the trusted certificates within the cluster, when cert-manager (or another external certificate management tool) is enabled.

This PR configures kustomize and helm charts according to this. 

There is no functional change, when internal cert management is enabled (which is the default setting).

This PR is inspired by https://github.com/kubernetes-sigs/kueue/pull/4385

```release-note
Injecting certificates to ServiceMonitor in order to serve securely, when cert-manager is enabled
```

### DIFF
--- a/charts/lws/templates/certmanager/certificatemetrics.yaml
+++ b/charts/lws/templates/certmanager/certificatemetrics.yaml
@@ -1,0 +1,20 @@
+{{- if .Values.enableCertManager }}
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  labels:
+    app.kubernetes.io/instance: serving-cert
+    {{- include "lws.labels" . | nindent 4 }}
+  name: lws-metrics-cert
+  namespace: {{ .Release.Namespace }}
+spec:
+  commonName: {{ include "lws.fullname" . }}-metrics
+  dnsNames:
+    - {{ include "lws.fullname" . }}-controller-manager-metrics-service.{{ .Release.Namespace }}.svc
+    - {{ include "lws.fullname" . }}-controller-manager-metrics-service.{{ .Release.Namespace }}.svc.cluster.local
+  issuerRef:
+    group: cert-manager.io
+    kind: Issuer
+    name: lws-selfsigned-issuer
+  secretName: lws-metrics-server-cert
+{{- end }}
--- a/charts/lws/templates/manager/deployment.yaml
+++ b/charts/lws/templates/manager/deployment.yaml
@@ -71,6 +71,11 @@ spec:
             - mountPath: /tmp/k8s-webhook-server/serving-certs
               name: cert
               readOnly: true
+            {{- if .Values.enableCertManager }}
+            - mountPath: /tmp/k8s-metrics-server/serving-certs
+              name: metrics-cert
+              readOnly: true
+            {{- end }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
@@ -88,3 +93,16 @@ spec:
           secret:
             defaultMode: 420
             secretName: lws-webhook-server-cert
+        {{- if .Values.enableCertManager }}
+        - name: metrics-cert
+          secret:
+            secretName: lws-metrics-server-cert
+            optional: false
+            items:
+              - key: ca.crt
+                path: ca.crt
+              - key: tls.crt
+                path: tls.crt
+              - key: tls.key
+                path: tls.key
+        {{- end }}
--- a/charts/lws/templates/prometheus/monitor.yaml
+++ b/charts/lws/templates/prometheus/monitor.yaml
@@ -15,7 +15,23 @@ spec:
       port: https
       scheme: https
       tlsConfig:
-        insecureSkipVerify: true
+        {{- if .Values.enableCertManager }}
+        serverName: {{ include "lws.fullname" . }}-controller-manager-metrics-service.{{ .Release.Namespace }}.svc
+        insecureSkipVerify: false
+        ca:
+          secret:
+            name: lws-metrics-server-cert
+            key: ca.crt
+        cert:
+          secret:
+            name: lws-metrics-server-cert
+            key: tls.crt
+        keySecret:
+          name: lws-metrics-server-cert
+          key: tls.key
+        {{- else }}
+          insecureSkipVerify: true
+        {{- end }}
   selector:
     matchLabels:
       control-plane: controller-manager
--- a/config/certmanager/certificate-metrics.yaml
+++ b/config/certmanager/certificate-metrics.yaml
@@ -1,0 +1,19 @@
+# The following manifests contain a metrics certificate CR.
+# More document can be found at https://docs.cert-manager.io
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: metrics-cert  # this name should match the one appeared in kustomizeconfig.yaml
+  namespace: system
+spec:
+  commonName: lws-metrics
+  dnsNames:
+    # SERVICE_NAME and SERVICE_NAMESPACE will be substituted by kustomize
+    # replacements in the config/default/kustomization.yaml file.
+    - SERVICE_NAME.SERVICE_NAMESPACE.svc
+    - SERVICE_NAME.SERVICE_NAMESPACE.svc.cluster.local
+  issuerRef:
+    group: cert-manager.io
+    kind: Issuer
+    name: selfsigned-issuer
+  secretName: metrics-server-cert
--- a/config/certmanager/kustomization.yaml
+++ b/config/certmanager/kustomization.yaml
@@ -1,5 +1,6 @@
 resources:
 - certificate.yaml
+- certificate-metrics.yaml
 
 configurations:
 - kustomizeconfig.yaml
--- a/config/components/prometheus/monitor_tls_patch.yaml
+++ b/config/components/prometheus/monitor_tls_patch.yaml
@@ -1,0 +1,19 @@
+# Patch for Prometheus ServiceMonitor to enable secure TLS configuration
+# using certificates managed by cert-manager
+- op: replace
+  path: /spec/endpoints/0/tlsConfig
+  value:
+    # SERVICE_NAME and SERVICE_NAMESPACE will be substituted by kustomize
+    serverName: SERVICE_NAME.SERVICE_NAMESPACE.svc
+    insecureSkipVerify: false
+    ca:
+      secret:
+        name: metrics-server-cert
+        key: ca.crt
+    cert:
+      secret:
+        name: metrics-server-cert
+        key: tls.crt
+    keySecret:
+      name: metrics-server-cert
+      key: tls.key
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -23,6 +23,7 @@ resources:
 - ../webhook
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'. 'WEBHOOK' components are required.
 - ../internalcert
+#- ../certmanager
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
 #- ../prometheus
 # [METRICS] Expose the controller manager metrics service.
@@ -42,6 +43,13 @@ patches:
 # Uncomment 'CERTMANAGER' sections in crd/kustomization.yaml to enable the CA injection in the admission webhooks.
 # 'CERTMANAGER' needs to be enabled to use ca injection
 #- path: webhookcainjection_patch.yaml
+
+# Uncomment the patches line if you enable Metrics and CertManager
+# [CERTMANAGER] To enable metrics protected with certManager, uncomment the following line.
+# This patch will protect the metrics with certManager self-signed certs.
+# - path: cert_metrics_manager_patch.yaml
+#  target:
+#    kind: Deployment
 
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER' prefix.
 # Uncomment the following replacements to add the cert-manager CA injection annotations
@@ -118,6 +126,7 @@ patches:
 #          kind: Certificate
 #          group: cert-manager.io
 #          version: v1
+#          name: serving-cert
 #        fieldPaths:
 #          - .spec.dnsNames.0
 #          - .spec.dnsNames.1
@@ -135,6 +144,7 @@ patches:
 #          kind: Certificate
 #          group: cert-manager.io
 #          version: v1
+#          name: serving-cert
 #        fieldPaths:
 #          - .spec.dnsNames.0
 #          - .spec.dnsNames.1
@@ -142,3 +152,62 @@ patches:
 #          delimiter: '.'
 #          index: 1
 #          create: true
+#  - source: # Add cert-manager annotation to the webhook Service
+#      kind: Service
+#      version: v1
+#      name: controller-manager-metrics-service
+#      fieldPath: .metadata.name # name of the service
+#    targets:
+#      - select:
+#          kind: Certificate
+#          group: cert-manager.io
+#          version: v1
+#          name: metrics-cert
+#        fieldPaths:
+#          - .spec.dnsNames.0
+#          - .spec.dnsNames.1
+#        options:
+#          delimiter: '.'
+#          index: 0
+#          create: true
+#      - select:
+#         kind: ServiceMonitor
+#         group: monitoring.coreos.com
+#         version: v1
+#         name: controller-manager-metrics-monitor
+#        fieldPaths:
+#         - spec.endpoints.0.tlsConfig.serverName
+#        options:
+#         delimiter: '.'
+#         index: 0
+#         create: true
+#  - source:
+#      kind: Service
+#      version: v1
+#      name: controller-manager-metrics-service
+#      fieldPath: .metadata.namespace # namespace of the service
+#    targets:
+#      - select:
+#          kind: Certificate
+#          group: cert-manager.io
+#          version: v1
+#          name: metrics-cert
+#        fieldPaths:
+#          - .spec.dnsNames.0
+#          - .spec.dnsNames.1
+#        options:
+#          delimiter: '.'
+#          index: 1
+#          create: true
+#      - select:
+#          kind: ServiceMonitor
+#          group: monitoring.coreos.com
+#          version: v1
+#          name: controller-manager-metrics-monitor
+#        fieldPaths:
+#          - spec.endpoints.0.tlsConfig.serverName
+#        options:
+#          delimiter: '.'
+#          index: 1
+#          create: true
+
--- a/charts/lws/templates/certmanager/certificate.yaml
+++ b/charts/lws/templates/certmanager/certificate.yaml
@@ -12,6 +12,7 @@ spec:
     - lws-webhook-service.{{ .Release.Namespace }}.svc
     - lws-webhook-service.{{ .Release.Namespace }}.svc.cluster.local
   issuerRef:
+    group: cert-manager.io
     kind: Issuer
     name: lws-selfsigned-issuer
   secretName: lws-webhook-server-cert
--- a/config/certmanager/certificate.yaml
+++ b/config/certmanager/certificate.yaml
@@ -34,6 +34,7 @@ spec:
   - SERVICE_NAME.SERVICE_NAMESPACE.svc
   - SERVICE_NAME.SERVICE_NAMESPACE.svc.cluster.local
   issuerRef:
+    group: cert-manager.io
     kind: Issuer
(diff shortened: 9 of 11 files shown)


yankay:
I'm not sure, whether additional documentation needs to be supplemented later, like https://github.com/kubernetes-sigs/kueue/pull/4598

ardaguclu:
> I'm not sure, whether additional documentation needs to be supplemented later, like [kubernetes-sigs/kueue#4598](https://github.com/kubernetes-sigs/kueue/pull/4598)

I think, adding documentation in a followup PR is reasonable and useful.

cc: @kerthcet

api/config/v1alpha1/configuration_types.go
```diff
@@ -108,6 +108,10 @@ type InternalCertManagement struct {
 	// Enable controls whether to enable internal cert management or not.
 	// Defaults to true. If you want to use a third-party management, e.g. cert-manager,
 	// set it to false. See the user guide for more information.
+	// This secret is mounted to the lws controller manager pod. The mount
```
kannon92, line 111: What secret do you mean here?

ardaguclu: Thanks for dropping your comment. I think, mentioning this behavior in this flag is not relevant. So that I've removed this comment entirely.

config/default/kustomization.yaml
```diff
@@ -135,10 +144,70 @@ patches:
...
+#          - .spec.dnsNames.1
+#        options:
+#          delimiter: '.'
+#          index: 1
+#          create: true
+#  - source: # Add cert-manager annotation to the webhook Service
+#      kind: Service
+#      version: v1
+#      name: controller-manager-metrics-service
+#      fieldPath: .metadata.name # namespace of the service
```
atiratree, line 159:
```suggestion
#      fieldPath: .metadata.name # name of the service
```

charts/lws/templates/certmanager/certificatemetrics.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.enableCertManager }}
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  labels:
+    app.kubernetes.io/instance: serving-cert
+    {{- include "lws.labels" . | nindent 4 }}
+  name: lws-metrics-cert
```
atiratree, line 8: why do we have two metrics certs?

charts/lws/templates/manager/deployment.yaml
```diff
@@ -88,3 +93,16 @@ spec:
           secret:
             defaultMode: 420
             secretName: lws-webhook-server-cert
+        {{- if .Values.enableCertManager }}
+        - name: metrics-certs
+          secret:
+            secretName: lws-metrics-server-cert
```
atiratree, line 99: in some we use lws-metrics-server-cert

config/default/kustomization.yaml
```diff
@@ -135,10 +144,70 @@ patches:
...
+#    targets:
+#      - select:
+#          kind: Certificate
+#          group: cert-manager.io
+#          version: v1
+#          name: metrics-certs
```
atiratree, line 165: and in some places we use metrics-certs

cmd/main.go
```diff
@@ -150,6 +152,23 @@ func main() {
 			os.Exit(1)
 		}
 	} else {
+		metricsCertPath := "/etc/lws/metrics/certs"
```
atiratree, line 155: IMO it is better to use the default values and watcher that are provided by the  `metricsserver.Options` . Please see https://github.com/kubernetes-sigs/jobset/pull/863 or more specifically

https://github.com/kubernetes-sigs/jobset/pull/863/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R169

config/components/prometheus/monitor_tls_patch.yaml
```diff
@@ -0,0 +1,19 @@
...
+    # SERVICE_NAME and SERVICE_NAMESPACE will be substituted by kustomize
+    serverName: SERVICE_NAME.SERVICE_NAMESPACE.svc
+    insecureSkipVerify: false
+    ca:
+      secret:
+        name: metrics-server-cert
+        key: ca.crt
+    cert:
+      secret:
+        name: metrics-server-cert
```
atiratree, line 15: IMO it is a bit weird to let the prometheus client use the same cert that we use in our server. I have left this to the user to specify the certificate their prometheus instance is using https://github.com/kubernetes-sigs/jobset/pull/863/files#diff-d1323215bef44ba29d323d647a042eed78e05ce849be1e3f8c067a89fd735359R23-R26

atiratree (commented):
Btw, I have opened a similar PR for the jobset controller: https://github.com/kubernetes-sigs/jobset/pull/863

config/default/kustomization.yaml
```diff
@@ -23,6 +23,7 @@ resources:
 - ../webhook
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'. 'WEBHOOK' components are required.
 - ../internalcert
+#- ../certmanager
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
```
atiratree, line 27: btw I think it would be good to allow passing certs to metrics even without a cert manager: https://github.com/kubernetes-sigs/jobset/pull/863/files#diff-c0dbb993e3ea0a70d85d76af178d499d9fe0ffa7fc9ad44d58d1acc7875e71d0R91

atiratree:
cc @kannon92, the questions can apply also to https://github.com/kubernetes-sigs/kueue/pull/4385

ardaguclu:
/hold
to apply the suggested changes from @atiratree

cmd/main.go
ardaguclu: According to here https://github.com/kubernetes-sigs/kubebuilder/blob/fa1d379fba769367def698662517f40e985d7c0d/pkg/plugins/golang/v4/scaffolds/internal/templates/cmd/main.go#L361-L381, this is the recommended way. Maybe `metricsserver.Options` does not rotate?

config/default/kustomization.yaml
ardaguclu: kustomize prepends `lws` https://github.com/kubernetes-sigs/lws/blob/66bcbc7f000dd65e66719aba7f5f9375bf7a00b5/config/default/kustomization.yaml#L9 to the resources and this will be formed to `lws-metrics-certs`. If there is any other value, it wouldn't be correct.

charts/lws/templates/certmanager/certificatemetrics.yaml
ardaguclu: There 2 different types of configurations. Kustomize prepends `lws` to the resource names but helm does not.

config/default/kustomization.yaml
ardaguclu: Thank you for warning me. Helm chart should be `lws-metrics-cert` and kustomize `metrics-cert` (not `metrics-certs`). I've updated the code.

cmd/main.go
ardaguclu: @kannon92 If we mount default paths defined in https://github.com/kubernetes-sigs/controller-runtime/blob/76067279145e76148cf27c01ea4ec63fed8e463b/pkg/metrics/server/server.go#L85-L102, we don't need to manually watch certificates. Controller-runtime manages this for us https://github.com/kubernetes-sigs/controller-runtime/blob/76067279145e76148cf27c01ea4ec63fed8e463b/pkg/metrics/server/server.go#L286-L309. As a result, there is no need to have code change in main.go

config/components/prometheus/monitor_tls_patch.yaml
ardaguclu: We are hardcoding nearly all fields and in my opinion we can hardcode it too.

ardaguclu:
I've tested this PR on a live cluster and it worked without any issue. Certificates are injected correctly and Prometheus successfully scraped the metrics in a secure way.
@kerthcet would you mind taking a look at this PR, when you have a chance?.

charts/lws/templates/certmanager/certificatemetrics.yaml
atiratree: Makes sense now, just noticed the helm support. Added a helm support to https://github.com/kubernetes-sigs/jobset/pull/863 as well.

config/components/prometheus/monitor_tls_patch.yaml
atiratree: IMO it is not entirely correct to let the prometheus use this as a client cert for auth. As mentioned in https://github.com/kubernetes-sigs/jobset/pull/863#discussion_r2032082412 there are many things that have to be configured before running cert manager + prometheus. So I think the best way would be to let user decide if they want to use the server certificate for this. They can easily opt in to it.

I have tried to document this in:
- https://github.com/atiratree/kubernetes-sigs-jobset/blob/enable-tls-metrics/config/components/prometheus/kustomization.yaml
- https://github.com/atiratree/kubernetes-sigs-jobset/blob/enable-tls-metrics/config/components/prometheus/monitor_tls_patch.yaml

charts/lws/templates/certmanager/certificatemetrics.yaml
```diff
@@ -0,0 +1,18 @@
...
+  namespace: {{ .Release.Namespace }}
+spec:
+  dnsNames:
```
atiratree, line 11: We should include a subject (e.g. `commonName`) on our certificates. There is an issue that some libraries might not recognize such self-signed cert as valid. This applies also for the webhook.

Please see https://cert-manager.io/docs/configuration/selfsigned/#certificate-validity for more details.

charts/lws/templates/certmanager/certificatemetrics.yaml
```diff
@@ -0,0 +1,18 @@
...
+    - lws-controller-manager-metrics-service.{{ .Release.Namespace }}.svc
+    - lws-controller-manager-metrics-service.{{ .Release.Namespace }}.svc.cluster.local
```
atiratree, line 13: It would be better if the service name referenced a helm variable.

charts/lws/templates/certmanager/certificatemetrics.yaml
```diff
@@ -0,0 +1,18 @@
...
+  issuerRef:
+    kind: Issuer
+    name: lws-selfsigned-issuer
+  secretName: lws-metrics-server-cert
+{{- end }}
```
atiratree, line 18: nit: newline

charts/lws/templates/prometheus/monitor.yaml
```diff
@@ -14,8 +14,7 @@ spec:
       path: /metrics
       port: https
       scheme: https
-      tlsConfig:
-        insecureSkipVerify: true
+      tlsConfig: {{- toYaml .Values.metrics.serviceMonitor.tlsConfig | nindent 6 }}
   selector:
     matchLabels:
```
atiratree, line 19: does the prometheus collection in the helm deployment work for you?
IMO `controller-manager-metrics-service` should have this label to work properly

charts/lws/values.yaml
```diff
@@ -49,3 +49,7 @@ resources:
 nodeSelector: {}
 tolerations: []
 affinity: {}
+metrics:
```
atiratree, line 52: lws-prometheus-k8s rolebinding monitoring namespace should be configurable IMO

config/certmanager/certificate-metrics.yaml
```diff
@@ -0,0 +1,17 @@
+# The following manifests contain a metrics certificate CR.
+# More document can be found at https://docs.cert-manager.io
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: metrics-cert  # this name should match the one appeared in kustomizeconfig.yaml
```
atiratree, line 6: helm review applies here as well

config/certmanager/kustomization.yaml
```diff
@@ -1,5 +1,6 @@
 resources:
 - certificate.yaml
+- certificate-metrics.yaml
```
atiratree, line 3: does it make sense to enable this by default?

config/components/prometheus/monitor_tls_patch.yaml
ardaguclu: > there are many things that have to be configured before running cert manager + prometheus.

We only require to uncomment some configurations in kustomize. 

If user relies on cert-manager, we can use the same certificates both in client and server because source of truth is the same (i.e. cert-manager) within this cluster.

If user relies on internal certificate management, skipping tls verification in servicemonitor would simply work. If we force users to find and inject a certificate in servicemonitor, users can't inject any arbitrary certificate. This certificate must comply with the one that is managed internally in LWS. I think, this dramatically hurts the user experience. Because I don't think, as a user I can successfully configure this cluster on my own.

charts/lws/templates/certmanager/certificatemetrics.yaml
ardaguclu: I think, users prefer simply applying these configurations to work with whatever set as default. Probably there is no functional need to configure the service name. I'll defer the decision to approvers.

charts/lws/templates/certmanager/certificatemetrics.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.enableCertManager }}
```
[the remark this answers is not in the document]
ardaguclu: If cert-manager is enabled, metrics will  be configured to run with TLS. That is independent from whether prometheus is enabled or not. If prometheus is not enabled, that means nobody scrapes metrics, but metrics server still runs in secure.

config/certmanager/kustomization.yaml
ardaguclu: By default cert-manager is disabled. If we enable this by default, I think we would get `Certificate` resource not found error?.

config/certmanager/certificate-metrics.yaml
ardaguclu: I'll check any missing piece in helm configurations. But I tested kustomize configurations extensively and it works.

charts/lws/templates/certmanager/certificatemetrics.yaml
ardaguclu: If there is a bug with respect to cert-manager configurations that is happening also in webhooks for some libraries, I'd prefer filing an issue to discuss the details and fix it in a followup PR. This PR focuses on building the foundational steps to configure servicemonitor for TLS. I'll again defer the decision to @kubernetes-sigs/lws-maintainers

charts/lws/templates/prometheus/monitor.yaml
```diff
@@ -14,8 +14,7 @@ spec:
       path: /metrics
       port: https
       scheme: https
-      tlsConfig:
-        insecureSkipVerify: true
+      tlsConfig: {{- toYaml .Values.metrics.serviceMonitor.tlsConfig | nindent 6 }}
```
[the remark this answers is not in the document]
ardaguclu: Good catch. Thank you. That was indeed missing. I've updated the code to better align with kustomize and inject the certificates when cert-manager is enabled.

charts/lws/templates/prometheus/monitor.yaml
ardaguclu: You are right. Updated.

charts/lws/templates/certmanager/certificatemetrics.yaml
```diff
@@ -0,0 +1,18 @@
...
+  labels:
+    app.kubernetes.io/instance: serving-cert
+    {{- include "lws.labels" . | nindent 4 }}
+  name: lws-metrics-cert
+  namespace: {{ .Release.Namespace }}
+spec:
+  dnsNames:
+    - lws-controller-manager-metrics-service.{{ .Release.Namespace }}.svc
+    - lws-controller-manager-metrics-service.{{ .Release.Namespace }}.svc.cluster.local
+  issuerRef:
```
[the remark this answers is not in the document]
ardaguclu: Done

config/components/prometheus/monitor_tls_patch.yaml
atiratree: I understand the usability perspective, but technically that certificate is wrong. In mutual TLS the domain name would be off.

I am okay, by adding it as a default, but IMO it should be configurable and we should write a disclaimer there.

charts/lws/templates/certmanager/certificatemetrics.yaml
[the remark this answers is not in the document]
atiratree: Right, it makes sense to secure the metrics even without prometheus scraping. I will update my PR as well.

charts/lws/templates/certmanager/certificatemetrics.yaml
atiratree: There is an issue which we run into in the kueue PR, making curl not accepting certificates without the subject. So I think it is better to include the subject it in our certificates as well.

- https://github.com/kubernetes-sigs/kueue/pull/4800#discussion_r2014226523
- https://github.com/kubernetes-sigs/kueue/pull/4800#discussion_r2032151059

charts/lws/templates/certmanager/certificatemetrics.yaml
atiratree: I am not saying it has to be configurable. I just mean to share a variable so we do not have to harcdode the name in multiple places.

charts/lws/templates/manager/deployment.yaml
```diff
@@ -88,3 +93,16 @@ spec:
           secret:
             defaultMode: 420
             secretName: lws-webhook-server-cert
+        {{- if .Values.enableCertManager }}
```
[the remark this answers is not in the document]
atiratree: not relevant anymore: https://github.com/kubernetes-sigs/lws/pull/478/files#r2032112051

charts/lws/templates/prometheus/monitor.yaml
```diff
@@ -15,7 +15,23 @@ spec:
       port: https
       scheme: https
       tlsConfig:
-        insecureSkipVerify: true
+        {{- if .Values.enableCertManager }}
+        serverName: {{ include "lws.fullname" . }}-controller-manager-metrics-service.{{ .Release.Namespace }}.svc
```
atiratree, line 19: we could use a variable here

config/certmanager/certificate-metrics.yaml
atiratree: I meant mostly this https://github.com/kubernetes-sigs/lws/pull/478/files#r2032107950

charts/lws/templates/prometheus/monitor.yaml
atiratree: I think this still applies as there is no service updated in the commits

config/certmanager/kustomization.yaml
atiratree: Ok makes sense, let's secure metrics regardless of prometheus.

charts/lws/templates/certmanager/certificatemetrics.yaml
ardaguclu: That sounds reasonable. Updated the code by adding `commonName` for metrics certificates.

charts/lws/templates/prometheus/monitor.yaml
```diff
@@ -15,7 +15,23 @@ spec:
...
+            key: ca.crt
+        cert:
+          secret:
+            name: lws-metrics-server-cert
+            key: tls.crt
+        keySecret:
+          name: lws-metrics-server-cert
+          key: tls.key
+        {{- else }}
+          insecureSkipVerify: false
```
kerthcet, line 33: Shouldn't this be true?

kerthcet (commented):
Only one comment, otherwise I think we can take off now, but I didn't test locally, this wouldn't impact users who use the internal cert.

ardaguclu:
> Only one comment, otherwise I think we can take off now, but I didn't test locally, this wouldn't impact users who use the internal cert.

@kerthcet thank you for review. I tested this change on a live cluster and verified that it works. Besides, as you mentioned this won't affect the cluster that rely on internal cert. 

Also if you don't mind, I'm planning to add a new CI job that enables cert-manager (disables internal cert) and runs the current e2e test on this configuration setting. This would ensure these kind of changes don't break anything. Does it make sense?.

kerthcet:
> This would ensure for us that these kind of changes don't break anything. Does it make sense?.

Yeah, definitely make sense. Is that build on helm chart? I think we don't have helm chart installation validation right now IIRC.

PR #488, opened 2025-04-07
Fix issue with rolling update when some existing replicas are unhealthy

pierewoj:
#### What type of PR is this?

/kind bug

The bug is as follows:
* assume replicas readiness is [NOT_READY, READY]
* update is triggered, with MaxUnavailable=1
* update starts updating the replica-1, causing complete downtime, while the desired behavior would be to wait for replica-0 to become ready

The replica-0 might be unavailable for a variety of reasons like hardware issues, node patching etc.

The fix is to reduce partition less aggressively - taking into account pods that are not ready.

Note that the fix was tested on a cluster with MaxUnavailableStatefulSet feature disabled (I don't have access to a cluster with this feature enabled, maybe the issue is less impactful there, however arguably still beneficial since the lws considers readiness of leader+workers and not only leader).

I have not created a dedicated issue for this in github

I also refactored the "iteratedReplicas" a bit into separate functions: a) reading the replica state and b) reading the state to get values used for rolling update params computation.

LWS user will see update progress less aggressively when not all replicas are healthy when update is triggered.

```

### DIFF
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -297,10 +297,12 @@ func (r *LeaderWorkerSetReconciler) rollingUpdateParameters(ctx context.Context,
 		return 0, lwsReplicas, nil
 	}
 
-	continuousReadyReplicas, lwsUnreadyReplicas, err := r.iterateReplicas(ctx, lws, stsReplicas, revisionKey)
+	states, err := r.getReplicaStates(ctx, lws, stsReplicas, revisionKey)
 	if err != nil {
 		return 0, 0, err
 	}
+	lwsUnreadyReplicas := calculateLWSUnreadyReplicas(states, lwsReplicas)
+	continuousReadyReplicas := calculateContinuousReadyReplicas(states)
 
 	originalLwsReplicas, err := strconv.Atoi(sts.Annotations[leaderworkerset.ReplicasAnnotationKey])
 	if err != nil {
@@ -326,7 +328,9 @@ func (r *LeaderWorkerSetReconciler) rollingUpdateParameters(ctx context.Context,
 
 	// When updated replicas become not ready again or scaled up replicas are not ready yet,
 	// we'll not modify the Partition field. That means Partition moves in one direction to make it simple.
-	return min(partition, utils.NonZeroValue(stsReplicas-int32(rollingStep)-continuousReadyReplicas)), wantReplicas(lwsUnreadyReplicas), nil
+	desiredPartition := utils.NonZeroValue(stsReplicas - int32(rollingStep) - continuousReadyReplicas)
+	partitionAccountingForUnavailable := partitionAccountingForUnavailable(states, desiredPartition)
+	return min(partition, partitionAccountingForUnavailable), wantReplicas(lwsUnreadyReplicas), nil
 }
 
 func (r *LeaderWorkerSetReconciler) SSAWithStatefulset(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, partition, replicas int32, revisionKey string) error {
@@ -511,19 +515,21 @@ func (r *LeaderWorkerSetReconciler) updateStatus(ctx context.Context, lws *leade
 	return updateDone, nil
 }
 
-// iterateReplicas will iterate the leader pods together with corresponding worker statefulsets
-// to check the replica state, and return two values and an error in the end:
-//   - The first value represents the number of continuous ready replicas ranging from the last index to 0,
-//     to help us judge whether we can update the Partition or not.
-//   - The second value represents the unready replicas whose index is smaller than leaderWorkerSet Replicas.
-func (r *LeaderWorkerSetReconciler) iterateReplicas(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, stsReplicas int32, revisionKey string) (int32, int32, error) {
+type replicaState struct {
+	ready   bool
+	updated bool
+}
+
+func (r *LeaderWorkerSetReconciler) getReplicaStates(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, stsReplicas int32, revisionKey string) ([]replicaState, error) {
+	states := make([]replicaState, stsReplicas)
+
 	podSelector := client.MatchingLabels(map[string]string{
 		leaderworkerset.SetNameLabelKey:     lws.Name,
 		leaderworkerset.WorkerIndexLabelKey: "0",
 	})
 	var leaderPodList corev1.PodList
 	if err := r.List(ctx, &leaderPodList, podSelector, client.InNamespace(lws.Namespace)); err != nil {
-		return 0, 0, err
+		return nil, err
 	}
 
 	// Get a sorted leader pod list matches with the following sorted statefulsets one by one, which means
@@ -537,7 +543,7 @@ func (r *LeaderWorkerSetReconciler) iterateReplicas(ctx context.Context, lws *le
 	})
 	var stsList appsv1.StatefulSetList
 	if err := r.List(ctx, &stsList, stsSelector, client.InNamespace(lws.Namespace)); err != nil {
-		return 0, 0, err
+		return nil, err
 	}
 	sortedSts := utils.SortByIndex(func(sts appsv1.StatefulSet) (int, error) {
 		return strconv.Atoi(sts.Labels[leaderworkerset.GroupIndexLabelKey])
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Edwinhr716** [comment]
/assign @kerthcet since he implemented the original logic

Also, #490 changes the logic behind rolling update a bit, so might need to rebase once that is merged

**Comment by congcongke** [comment]
> #### What this PR does / why we need it
> The bug is as follows:
> 
> * assume replicas readiness is [NOT_READY, READY]
> * update is triggered, with MaxUnavailable=1
> * update starts updating the replica-1, causing complete downtime, while the desired behavior would be to wait for replica-0 to become ready
> 
> The replica-0 might be unavailable for a variety of reasons like hardware issues, node patching etc.
> 
> The fix is to reduce partition less aggressively - taking into account pods that are not ready.
> 
> Note that the fix was tested on a cluster with MaxUnavailableStatefulSet feature disabled (I don't have access to a cluster with this feature enabled, maybe the issue is less impactful there, however arguably still beneficial since the lws considers readiness of leader+workers and not only leader).
> 
@pierewoj 
I didn't get the key point.

* update is rolling, new replicas will be created, right?
* after new replica ready, the old one will be updated.
* if the old is not ready forever, the replicas will be keeped burst.

**Comment by pierewoj** [comment]
> > #### What this PR does / why we need it
> > The bug is as follows:
> > 
> > * assume replicas readiness is [NOT_READY, READY]
> > * update is triggered, with MaxUnavailable=1
> > * update starts updating the replica-1, causing complete downtime, while the desired behavior would be to wait for replica-0 to become ready
> > 
> > The replica-0 might be unavailable for a variety of reasons like hardware issues, node patching etc.
> > The fix is to reduce partition less aggressively - taking into account pods that are not ready.
> > Note that the fix was tested on a cluster with MaxUnavailableStatefulSet feature disabled (I don't have access to a cluster with this feature enabled, maybe the issue is less impactful there, however arguably still beneficial since the lws considers readiness of leader+workers and not only leader).
> 
> @pierewoj I didn't get the key point.
> 
>     * update is rolling, new replicas will be created, right?
> 
>     * after new replica ready, the old one will be updated.
> 
>     * if the old is not ready forever, the replicas will be keeped burst.

Consider maxSurge=0, replicas=2, maxUnavail=1 and cluster with MaxUnavailableStatefulSet disabled. If replicas readiness is [NOT_READY, READY] then the rolling update would bring down the 2nd replica and cause complete unavailability.

**Comment by congcongke** [comment]
> > @pierewoj I didn't get the key point.
> > ```
> > * update is rolling, new replicas will be created, right?
> > 
> > * after new replica ready, the old one will be updated.
> > 
> > * if the old is not ready forever, the replicas will be keeped burst.
> > ```
> 
> Consider maxSurge=0, replicas=2, maxUnavail=1 and cluster with MaxUnavailableStatefulSet disabled. If replicas readiness is [NOT_READY, READY] then the rolling update would bring down the 2nd replica and cause complete unavailability.

Got it.

Rolling update will be blocked if the actual count of unavailable **GE** the maxUnavail.

**Comment by pierewoj** [comment]
> Got it.
> 
> Rolling update will be blocked if the actual count of unavailable **GE** the maxUnavail.

Yes:
* current behavior would be avail drop and update of the replica-1
* with the fix, the rolling update would wait for the replica-0 to become ready before proceeding to update replica-1 (for example, finish loading the LLM)

**Comment by pierewoj** [comment]
> Overall I agree with this change. If I understand correctly, the main issue is that our core logic doesn't take differentiate between a replica not being ready and a replica not being updated (essentially assumes that if the replica is not updated, it means that it must be running) here:
> 
> https://github.com/kubernetes-sigs/lws/blob/cb41374204030aabd434f3082ad07c49dbb64aab/pkg/controllers/leaderworkerset_controller.go#L558
> 
> To simplify it, could add an extra count for `readyReplicas` and return three ints here instead?
> 
> https://github.com/kubernetes-sigs/lws/blob/cb41374204030aabd434f3082ad07c49dbb64aab/pkg/controllers/leaderworkerset_controller.go#L520

This was my original implementation actually but I think that it is insufficient. See the following scenario:
* replicas=[NOT_READY, NOT_READY], maxUnavail=1, maxSurge=0
* existing lws/main algo will conpute partition=1 (which is actually ok)
* proposed simplified algorithm will produce a counter that would move the partition to partition=2 due to replicas not being ready and the update would get stuck
* this is not desired, since the replicas are not ready anyway so it's safe to update
* this is why the proposed algorithm in PR has the for loop at L633

**Comment by Edwinhr716** [comment]
We can keep the proposed algorithm to calculate maxUnavailable, but instead of passing the states struct just pass the `readyReplicas` value instead. Current implementation requires iterating through the number of replicas four different times. Would be better to just iterate through them twice, and keep the core logic of calculating `readyReplicas`,`continousReady` and `unreadyReplicas` in `iterateReplicas`

**Comment by congcongke** [comment]
If maxSurge is set to 1 and maxUnavailable to 0, then [NotReady, Ready] happened, will 1 new replica be created?

**Comment by pierewoj** [comment]
> We can keep the proposed algorithm to calculate maxUnavailable, but instead of passing the states struct just pass the readyReplicas value instead. Current implementation requires iterating through the number of replicas four different times. Would be better to just iterate through them twice, and keep the core logic of calculating readyReplicas,continousReady and unreadyReplicas in iterateReplicas

I think that the proposed algorithm would need to change, since the logic to move the partition depends on the "desiredPartition" right now, which is known later on only. Also, the iterate replicas would have to return 3 (or 4) values and the readability is not going to be great imo.

Do you insist on this refactor? Is iterating through the replica statuses a performance/scalability concern? If you strongly insist, I can follow this as maintainer guidance and implement (probably in a few days).

> If maxSurge is set to 1 and maxUnavailable to 0, then [NotReady, Ready] happened, will 1 new replica be created?

That should be the case, and not affected by this PR since it only affects the returned partition and not the wantReplicas.

**Comment by Edwinhr716** [comment]
> I think that the proposed algorithm would need to change, since the logic to move the partition depends on the "desiredPartition" right now

DesiredPartition can still be calculated without using the states struct no? Isn't it the same calculation as the one currently done for partition? 

> Is iterating through the replica statuses a performance/scalability concern?

Yeah mostly scalability/performance concern when performing rolling updates. We are 4x the amount of times we iterate through all of the replicas. I think 2x is a good compromise

> Also, the iterate replicas would have to return 3 (or 4) values and the readability is not going to be great imo.

Mmm yeah, we could also return a struct, but with continousReady, lwsUnReady, and readyReplicas as its values instead of ready, unready, to maintain readability.

**Comment by kerthcet** [comment]
I'll take a look today.

File: pkg/controllers/leaderworkerset_controller.go [hunk 1]
```diff
@@ -537,51 +543,101 @@ func (r *LeaderWorkerSetReconciler) iterateReplicas(ctx context.Context, lws *le
...
-	var continuousReadyReplicas, lwsUnreadyReplicas int32
+	return states, nil
+}
 
-	for index := stsReplicas - 1; index >= 0; index-- {
-		replicaReady := processReplica(index)
-		skip = skip || !replicaReady
+func calculateLWSUnreadyReplicas(states []replicaState, lwsReplicas int32) int32 {
+	var unreadyCount int32
+	for idx := int32(0); idx < min(int32(len(states)), lwsReplicas); idx++ {
```
kerthcet, line 596: Is this right? Shouldn't we use lwsReplicas as the evaluation baseline？For example, lwsReplicas=10, the len(stats)=5, because the other 5 is rebuilding, then we at least have 5 unready replicas.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -511,19 +515,21 @@ func (r *LeaderWorkerSetReconciler) updateStatus(ctx context.Context, lws *leade
...
 
-// iterateReplicas will iterate the leader pods together with corresponding worker statefulsets
-// to check the replica state, and return two values and an error in the end:
-//   - The first value represents the number of continuous ready replicas ranging from the last index to 0,
-//     to help us judge whether we can update the Partition or not.
-//   - The second value represents the unready replicas whose index is smaller than leaderWorkerSet Replicas.
-func (r *LeaderWorkerSetReconciler) iterateReplicas(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, stsReplicas int32, revisionKey string) (int32, int32, error) {
+type replicaState struct {
+	ready   bool
+	updated bool
```
kerthcet, line 520: Seems no need to distinguish ready and updated separately.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -537,51 +543,101 @@ func (r *LeaderWorkerSetReconciler) iterateReplicas(ctx context.Context, lws *le
...
-			lwsUnreadyReplicas++
+		continuousReadyCount++
+	}
+	return continuousReadyCount
+}
+
+// Checks state of not updated replicas and adjusts the rolling update parittion to prevent
+// availability drops below "maxUnavailable".
+func partitionAccountingForUnavailable(states []replicaState, desiredPartition int32) int32 {
+	// Desired partition already accounts for readiness of replicas above partition. Code below
```
kerthcet, line 619: Can we calculate just in one place? Like refactoring the original part, would that make the code more readable?

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -537,51 +543,101 @@ func (r *LeaderWorkerSetReconciler) iterateReplicas(ctx context.Context, lws *le
...
+		}
+	}
+
+	partitionAccountingForUnavailable := desiredPartition + unavailableBelowPartition
+
+	// Reduce the partitionAccountingForUnavailable as long as replicas are unavailable. This is safe to update these
+	// replicas as it does not impact general availability.
+	// Never reduce below desiredPartition.
+	for idx := partitionAccountingForUnavailable; idx >= desiredPartition; idx-- {
+		if idx < int32(len(states)) && states[idx].ready {
```
kerthcet, line 634: I think we're playing a trick here, the first ready partition is actually not the accurate one, which may occur like the partition is getting bigger and bigger, but as mentioned, it won't impact the final result.

Another approach is comparing the current partition with the original one, but we need to pay attention to partition=0.

kerthcet (commented):
left some comments.

pkg/controllers/leaderworkerset_controller.go
pierewoj: calculateLWSUnreadyReplicas/calculateContinuousReadyReplicas could use a joint `ready&updated` state while partitionAccountingForUnavailable uses `ready` only

pkg/controllers/leaderworkerset_controller.go
pierewoj: Do you mean trying to extend the `iterateReplicas` and adding more return values to the result as @Edwinhr716 suggested?

pierewoj:
@kerthcet thanks for review. I finally had some time to come back to this PR, I've updated the diff with fixes, PTAL

The suggestion:
> Another approach is comparing the current partition with the original one, to ensure the partition will not get smaller, but we need to pay attention to partition=0.

was a bit ambiguous for me, so I decided to refactor the partition calculation logic into a single function, where it's obvious that the partition is being decreased. It's a bigger change in the controller logic so please let me know what you think.

kerthcet:
Thanks @pierewoj  I checked the code, most of the codes are easy to read, except the `rollingUpdatePartition`. I was wondering whether we can simplify it like `min(partition, utils.NonZeroValue(stsReplicas-int32(rollingStep)-continuousReadyReplicas + unavailablePods)`, the `unavailablePods` are the ones we forget to consider before. Would like to hear your advices since you have refactored the code a lot.

kerthcet:
> This was actually my initial implementation when I tried to fix this issue (not shared in PR), but I think this is insufficient. We want to be able to update the replicas that are not ready, since it does not impact the availability. Imagine following case (the R/N stand for ready/notready and _X is revision number, for example R_1 means redy revision 1):

That's true, what I proposed is a more safe solution, but less efficient that yours. I'll take another look.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -537,51 +540,105 @@ func (r *LeaderWorkerSetReconciler) iterateReplicas(ctx context.Context, lws *le
...
+			unavailable++
+		}
+	}
+
+	// Iterate over replicas, determine if moving partition will violate the maxUnavailable (rollingStep).
+	// Note that the returned partition will always be smaller than or equal to currentPartition.
+	// When updated replicas become not ready again or scaled up replicas are not ready yet,
+	// we'll not modify the Partition field. That means Partition moves in one direction to make it simple.
+	var partition = currentPartition
+	for idx := partition - 1; idx >= minAllowedPartition; idx-- {
```
kerthcet, line 607: If we agree that the to-update-partition is belong to [minAllowedPartition, currentPartition], can we iterate the [0, minAllowedPartition) to find the unavailable ones, and the to-update-partition should be `min(minAllowedPartition+unavailable, currentPartition)`.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -537,51 +540,105 @@ func (r *LeaderWorkerSetReconciler) iterateReplicas(ctx context.Context, lws *le
...
+	var minAllowedPartition = utils.NonZeroValue(stsReplicas - continuousReadyReplicas - rollingStep)
 
-		if replicaReady && !skip {
-			continuousReadyReplicas++
+	// Replicas that are not ready or expected to be updated (since idx >= partition) given current partition.
+	// Used to determine how much the partition can be reduced.
+	var unavailable int32
+	for idx := 0; idx < int(stsReplicas-continuousReadyReplicas); idx++ {
+		if !states[idx].ready || idx >= int(currentPartition) {
+			unavailable++
```
kerthcet, line 598: I guess the `unavailable` is not right here because `idx >= partition` doesn't mean it's unavailable?

pierewoj:
> If we agree that the to-update-partition is belong to [minAllowedPartition, currentPartition], can we iterate the [0, minAllowedPartition) to find the unavailable ones, and the to-update-partition should be min(minAllowedPartition+unavailable, currentPartition).

Discussed over Slack; I have implemented the proposed change together with adding the logic to reduce partition in case replicas are continuously unready, which should address the risk of deployments getting stuck.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -537,51 +540,102 @@ func (r *LeaderWorkerSetReconciler) iterateReplicas(ctx context.Context, lws *le
...
-		if !replicaReady && index < *lws.Spec.Replicas {
-			lwsUnreadyReplicas++
+	}
+	var partition = rolliingStepPartition + unavailable
+
+	// Reduce the partition if replicas are continously not ready. It is safe since updating these replicas does not impact
+	// the availability of the LWS. This is important to prevent update from getting stuck in case maxUnavailable is already violated
+	// (for example, all replicas are not ready when rolling update is started).
+	// Note that we never drop the partition below rolliingStepPartition.
+	for idx := min(partition, stsReplicas-1); idx >= rolliingStepPartition; idx-- {
```
kerthcet, line 607: What's the take here of `min(partition, stsReplicas-1)` rather than using partition directly.

kerthcet (commented):
Only one question.
/approve

pkg/controllers/leaderworkerset_controller.go
pierewoj: This is because partition (once increased by "unavailable") can be too big and result in index out of bound errors for the states.

This could be the case if we have a lot of unhealthy replicas and a small rolling step.

It's not a problem since we take min(partition, currentPartition), but we have to consider it in the partition reduction loop.

kerthcet: Thank, let's merge this then.
/lgtm

PR #489, opened 2025-04-07
Introduce env variable to modify ns in metrics tests

ardaguclu:
#### What type of PR is this?
/kind cleanup

`should ensure the metrics endpoint is serving metrics` test strictly rely on `lws-system` namespace while checking the resources. However, this test should also work on the environments that LWS is deployed to another namespaces. 

This PR introduces `LWS_NAMESPACE` environment variable. If it is not set, default `lws-system` will be used.

```release-note
None
```


yankay:
Thanks @ardaguclu 
/lgtm

PR #490, opened 2025-04-07
Fix UpdateInProgress conditions for newly created LWS

### LINKED ISSUE #400: Create a new lws or modify replicas, UpdateInProgress will also exist in conditions
**What happened**:
I am adapting lws in volcano, and when adapting the rolling update scenario, I determine whether an lws is in a rolling update based on the presence of an UpdateInProgress condition set to True in the lws. However, now when I create a new lws, I find that sometimes the UpdateInProgress condition is also stably generated, which does not match the description of this condition:
```
// LeaderWorkerSetUpdateInProgress means lws is performing a rolling update. UpdateInProgress
```
When creating a new lws or modifying replicas, this condition should not occur, but currently I can basically reproduce this situation, especially repeatedly creating and deleting the same lws should be able to stably reproduce it

When creating a new lws or modifying replicas, this condition should not occur

My lws yaml:
```
kind: LeaderWorkerSet
metadata:
  name: leaderworkerset-rollout
spec:
  rolloutStrategy:
    type: RollingUpdate
    rollingUpdateConfiguration:
      maxSurge: 2
  replicas: 4
  leaderWorkerTemplate:
    size: 4
    workerTemplate:
      metadata:
        name: worker    
      spec:
        containers:
        - name: nginx
          image: nginx:1.14.2
          resources:
            limits:
              cpu: "100m"
            requests:
              cpu: "50m"
          ports:
          - containerPort: 8080
```

Especially repeatedly creating and deleting the same lws should be able to stably reproduce it

**Environment**:

- Kubernetes version (use `kubectl version`): v1.31
- LWS version (use `git describe --tags --dirty --always`): v0.5.1
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

yankay:
#### What type of PR is this?

/kind bug

**Fixes #400**  

**Bug Scenario**:  
A condition occurs where the LWS worker StatefulSet has not yet been properly initialized.  

**Root Cause**:  
In  https://github.com/kubernetes-sigs/lws/pull/298/files#diff-636763c4aab72861dbe05856e71e11abf551830d79fdcd848f3133ffc9c2d6e1R386 , the introduction of the `continue` keyword causes `currentNonBurstWorkerCount++` to execute prematurely.  

This leads to the condition `updatedNonBurstWorkerCount < currentNonBurstWorkerCount` at https://github.com/kubernetes-sigs/lws/blob/v0.6.0/pkg/controllers/leaderworkerset_controller.go#L441 , triggering the bug.  

**Fix**:  
Simply move `currentNonBurstWorkerCount++` **below** the `continue` statement to resolve the issue.  

```release-note
Fix UpdateInProgress conditions for newly created LWS
```

### DIFF
--- a/test/integration/controllers/leaderworkerset_test.go
+++ b/test/integration/controllers/leaderworkerset_test.go
@@ -464,6 +464,26 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
 				},
 			},
 		}),
+		ginkgo.Entry("Ensure UpdateInProgress condition is not set during LWS initialization", &testCase{
+			makeLeaderWorkerSet: func(nsName string) *wrappers.LeaderWorkerSetWrapper {
+				return wrappers.BuildLeaderWorkerSet(nsName).Replica(4).MaxSurge(2)
+			},
+			updates: []*update{
+				{
+					checkLWSState: func(lws *leaderworkerset.LeaderWorkerSet) {
+						testing.ExpectLeaderWorkerSetProgressing(ctx, k8sClient, lws, "Replicas are progressing")
+						gomega.Expect(func() error {
+							for _, c := range lws.Status.Conditions {
+								if c.Type == string(leaderworkerset.LeaderWorkerSetUpdateInProgress) && c.Status == metav1.ConditionTrue {
+									return fmt.Errorf("UpdateInProgress condition is not expected")
+								}
+							}
+							return nil
+						}()).To(gomega.Succeed())
+					},
+				},
+			},
+		}),
 		ginkgo.Entry("Test default progressing state", &testCase{
 			makeLeaderWorkerSet: wrappers.BuildLeaderWorkerSet,
 			updates: []*update{
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -389,9 +389,6 @@ func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *l
 		if err != nil {
 			return false, false, err
 		}
-		if index < int(*lws.Spec.Replicas) {
-			currentNonBurstWorkerCount++
-		}
 
 		var sts appsv1.StatefulSet
 		if !noWorkerSts {
@@ -402,6 +399,10 @@ func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *l
 				}
 				continue
 			}
+		}
+
+		if index < int(*lws.Spec.Replicas) {
+			currentNonBurstWorkerCount++
 		}
 
 		var ready, updated bool


Edwinhr716:
Thanks for taking a look! Would it be possible to add an integration/unit test to cover this test case?

yankay:
> Thanks for taking a look! Would it be possible to add an integration/unit test to cover this bug?

HI @Edwinhr716 

A test case has been added :-)

test/integration/controllers/leaderworkerset_test.go
```diff
@@ -464,6 +464,26 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
...
 		}),
+		ginkgo.Entry("Ensure UpdateInProgress condition is not set during LWS initialization", &testCase{
+			makeLeaderWorkerSet: func(nsName string) *wrappers.LeaderWorkerSetWrapper {
+				return wrappers.BuildLeaderWorkerSet(nsName).Replica(4).MaxSurge(2)
+			},
+			updates: []*update{
+				{
+					checkLWSState: func(lws *leaderworkerset.LeaderWorkerSet) {
+						testing.ExpectLeaderWorkerSetProgressing(ctx, k8sClient, lws, "Replicas are progressing")
+						gomega.Expect(func() error {
```
Edwinhr716, line 475: Why not use this function instead? https://github.com/kubernetes-sigs/lws/blob/370b8692a62ed6135d70c6a944b3be78a043312c/test/testutils/validators.go#L387

test/integration/controllers/leaderworkerset_test.go
```diff
@@ -464,6 +464,26 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
 				},
 			},
 		}),
+		ginkgo.Entry("Ensure UpdateInProgress condition is not set during LWS initialization", &testCase{
```
Edwinhr716, line 467: Does this test fail without the fix?

test/integration/controllers/leaderworkerset_test.go
yankay: HI @Edwinhr716 

In `ExpectLeaderWorkerSetNoUpgradeInProgress`, `CheckLeaderWorkerSetHasCondition` is designed to verify the presence of the UpdateInProgress condition.
However, the expected behavior is that the `UpdateInProgress condition` ​​should not exist yet​​, so the check will fail (timeout error) and cause the test to fail.

This is why we need this code segment—to explicitly handle this validation.

test/integration/controllers/leaderworkerset_test.go
yankay: When the fix is not applied, the test will fail. Below is a screenshot of the error:

img_v3_02l6_70f49207-0347-4b85-bed9-7ffc93cb1aag

test/integration/controllers/leaderworkerset_test.go
Edwinhr716: That makes sense

test/integration/controllers/leaderworkerset_test.go
Edwinhr716: Thanks

PR #491, opened 2025-04-07
fix: don't release the burst replica until the unreadyReplicas less than maxSurge

congcongke:
#### What type of PR is this?

/kind bug

In my opinion, the replica should be decreased when the (unready replica < maxSurge). It means the rolling update is in progress and newly created one is partly stable.

Background:
1. LWS is in running with 1 replica, 25% maxSurge and 25% minUnavailable
2. Update the image or cpu quota of LWS

What I want:
1. Scale **Statefulset** up to 2 replica, with the second instance with new image.
2. After the second one is in running, update the first instance
3. Scale **Statefulset** down to 1 replica, with all the instance updated

Actual
1. The replica of **Statefulset** is stay at 1 with partition is set to 1

```release-note

```


pkg/controllers/leaderworkerset_controller_test.go
```diff
@@ -536,3 +538,163 @@ func TestSetCondition(t *testing.T) {
 		})
 	}
 }
+
+func TestRollingUpdateParameters(t *testing.T) {
```
Edwinhr716, line 542: This is a copy of the existing integration tests, I don't think it is needed

congcongke: yes, removed it

kerthcet:
What's the status here? @Edwinhr716

Edwinhr716:
Only saw the duplication of tests, haven't really taken a look at it or thought about if it is something we want

Issue #407, opened 2025-03-04
Support custom scheulder to enable gang scheduling


JesseStutler:
**What would you like to be added**:
Support custom scheulder to enable gang scheduling ( e.g. Volcano, Scheduler plugins etc)

**Why is this needed**:
#167 , currently LeaderWorkerSet does not support Gang Scheduling. Although currently 1 Leader + (size-1) workers form a Group, in terms of scheduling, they do not have a Group relationship and are scheduled separately

**Completion requirements**:

This enhancement requires the following artifacts:

- [x] Design doc
- [ ] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

I have conducted an initial exploration of adapting LeaderWorkerSet in Volcano, and here are my current insights:
## Create LeaderWorkerSet
When the LeaderWorkerSet is created, create replicas number of PodGroups. The configuration of the PodGroup has two types:
### StartUpPolicy is LeaderCreated (default)
When StartUpPolicy is LeaderCreated, the Leader Pod and Worker Pods are created simultaneously. The MinMember of the PodGroup is size, indicating that these Pods must be scheduled together. If there are insufficient resources for one pod, the entire group cannot be scheduled. The resource amount of the PodGroup is the total resource request of these Pods
### StartUpPolicy is LeaderReady
When StartUpPolicy is LeaderReady, the Leader Pod needs to be scheduled and Running first, so MinMember is set to 1, but the resource amount of the PodGroup is still the total resource amount of the Leader Pod and worker Pods. This setting relies on the enqueue judgment; if the requested resources exceed the available cluster resources, then the PodGroup cannot be enqueued, and thus the Leader Pod will not be scheduled. The LeaderReady approach can only try to meet the concept of gang scheduling as much as possible, but it essentially still conflicts somewhat with gang scheduling.

## LeaderWorkerSet scale up/down
When the LeaderWorkerSet undergoes a scale down/scale up, the number of corresponding PodGroups also needs to be adjusted
## LeaderWorkerSet Rolling Update
When a Rolling Update occurs for the LeaderWorkerSet, it is possible that the user has configured maxSurge, which would require additional PodGroups to be created. Therefore, if the lws's `.status.Replicas` is updated, the difference between the current number of PodGroups and `.status.Replicas` should be compared. If there are fewer PodGroups, additional ones need to be created so that the rolling update of the LeaderWorkerSet can proceed successfully. After the rolling update is completed, the number of PodGroups will return to the replicas.

[...]

Issue #444, opened 2025-03-17
Support for LWS level ResourceClaimTemplate


Edwinhr716:
**What would you like to be added**:
I would like to be able to associate a ResourceClaimTemplate with a leader-worker group in LWS, such that each time a new LWS replica is created, a new ResourceClaim is created based on the template, and all the pods in the LWS replica are pointed to that Resource Claim

**Why is this needed**:
Similar reason as JobSet https://github.com/kubernetes-sigs/jobset/issues/762. In order to support ResourceClaims that represent multi-node resource, pods need to share the ResourceClaim. The main idea is that when an LWS replica is created, you will allocate a TPU slice or an IMEX channel that is shared amongst all Pods in the LWS replica. 

This enhancement requires the following artifacts:

- [X] Design doc
- [X] API change
- [X] Docs update

The artifacts should be linked in subsequent comments.
cc @johnbelamaric

Edwinhr716:
Target is to have this ready for the v0.7.0 release

kannon92:
So in jobset issue, our main concern is trying to implement this in each operator may be challenging.

If you want to POC this that would be great! I worry that this should be an API change in Kubernetes rather than LWS/JobSet/MPIJob/etc all supporting this.

But I can see why you want to go ahead with it..

ahg-g:
I think it will be hard to generalize it. For example, even if StatefulSet supports ResourceClaim templates, it wouldn't work for LWS because we may want to have the leader point to the same claim with its workers. Also, for disaggregated serving, we want to create a claim be subgroup.

kannon92:
Yea, I'd love to see a KEP on this then!

I haven't had time to think this through on JobSet side.

We brought this up in wg-device-management a few weeks ago.

Edwinhr716:
> Also, for disaggregated serving, we want to create a claim be subgroup

We should have this as a follow up item, not part of the initial KEP, wdyt? @ahg-g

ahg-g:
We don't need to implement everything in the KEP in one go, but the KEP should try to describe a future proof solution that includes subgroups

kerthcet:
Glad to see this happen.

JesseStutler:
Glad to have this feature, just as @ahg-g said: https://github.com/kubernetes-sigs/lws/issues/407#issuecomment-2704252088, can we extract a public interface for these lws relica level resources? I'm also looking forward to create podgroup for each lws replica.

Edwinhr716:
[...]

PR #496, opened 2025-04-10
Add KEP#407: Gang Scheduling

JesseStutler:
#### What type of PR is this?
/kind documentation

Ref #407, Without gang scheduling support, partial scheduling may lead to resource deadlocks. 
As reported in [issue #167](https://github.com/kubernetes-sigs/lws/issues/167), when cluster resources are limited, 
the scheduler may prioritize scheduling leader pods while leaving their worker pods pending. 
This results in a situation where no inference service can serve properly, as each service requires both leader and worker pods to be running.
Gang scheduling support would ensure that pods within a replica are either all scheduled together or not scheduled at all, 
thus preventing resource deadlocks and maintaining service availability.

Ref #407 

```release-note
NONE
```


keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,403 @@
...
+		...
+	} else {
+		// Since the worker sts has already been set the groupIndex label according to the leader pod's groupIndex label during creation in pod controller, 
+		// no special handling is needed for worker pods
+		...
+	}
+
+	// Set the podgroup affiliation for the leader pod and worker pod
+	if utilfeature.DefaultFeatureGate.Enabled(features.PodGroupPerReplica) {
+		err = p.podGroupProvider.SetPodMeta(pod)
```
Edwinhr716, line 222: is `podGroupProvider` a new field in the `pod_webhook` and `pod_controller` struct? Is this field defined at startup time?

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,403 @@
...
+	// SetPodMeta sets pod meta for PodGroup association, called by webhook
+	SetPodMeta(pod *corev1.Pod) error
+}
+```
+Define a new interface called `Provider` to manage PodGroup resources. Whenever a leader pod is created, 
+the PodGroup will also be created with its OwnerReference set to the leader pod, so its lifecycle will follow the leader pod. 
+This way, whether the leader pod is deleted due to scale-down of lws, deletion of lws, or any other reason, the PodGroup will also be deleted. 
+Besides, during a lws rolling update, the PodGroup will be recreated along with the leader pod.
+- `CreatePodGroupIfNotExists`: This method will be called by the pod controller to create a PodGroup resource for each lws replica if it doesn't already exist.
+  For schedulers with PodGroup CRD (such as Volcano, coscheduling scheduler plugin), the creation logic of the corresponding PodGroup resource needs to be
```
Edwinhr716, line 181: Is the expectation that the scheduler implements the `Provider` interface? If so, how would it interact if LWS also implements the interface to create a headless service for each PodGroup?

keps/407-gang-scheduling/README.md
JesseStutler: Yes, in pod webhook or pod controller initialization, I will init the podGroupProvider field, depending on what type of PodGroupProvider the user wants to use, for example, like this:
```go
func SetupPodWebhook(mgr ctrl.Manager) error {
	pw := &PodWebhook{}
	if utilfeature.DefaultFeatureGate.Enabled(features.PodGroupPerReplica) {
		providerType := os.Getenv("LWS_PODGROUP_PROVIDER")
		podGroupProvider, err := podgroup.NewPodGroupProvider(podgroup.ProviderType(providerType), mgr.GetClient())
		if err != nil {
			return err
		}
		pw.podGroupProvider = podGroupProvider
	}

	return ctrl.NewWebhookManagedBy(mgr).
		For(&corev1.Pod{}).
		WithDefaulter(&PodWebhook{}).
		WithValidator(&PodWebhook{}).
		Complete()
}
```

keps/407-gang-scheduling/README.md
JesseStutler: No, each scheduler still needs to implement the Provider interface in lws, with the code written in the lws repository. I expect lws to create podgroup resources as the upstream and set the pod's associated podgroup in their label/annotation, while the downstream custom scheduler focuses only on the scheduling of podgroups. If possible, I can actually draft a WIP PR for the Volcano example I've written, to facilitate the discussion of the proposal and review of this KEP

JesseStutler: So this Provider interface, in the future, can be refactored to include the creation of ResourceClaim and headlessService as well, I think?

JesseStutler: https://github.com/kubernetes-sigs/lws/pull/498, I submitted a WIP PR to facilitate the review of the KEP's rationality. This PR includes an example of a Volcano PodGroup Provider, which can be referenced. For other custom schedulers, it is the same. If we want to integrate a new custom scheduler, we need to implement the Provider interface and add a new implementation under `pkg/utils/podgroup`

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,407 @@
...
+Gang scheduling support would ensure that pods within a replica are either all scheduled together or not scheduled at all, 
+thus preventing resource deadlocks and maintaining service availability.
+
+### Goals
+
+<!--
+List the specific goals of the KEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+- Create different PodGroup resources for each lws replica for different custom schedulers (Volcano, scheduler-plugins, YuniKorn etc.) to enable gang scheduling
```
googs1025, line 95: Do we need to indicate that this has dependencies, such as coscheduling with scheduler-plugins or gang scheduling features of other scheduling projects? 🤔

JesseStutler: Sure, I can add some explanations

keps/407-gang-scheduling/README.md
Edwinhr716: > So this Provider interface, in the future, can be refactored to include the creation of ResourceClaim and headlessService as well, I think?

If you could provide an example of this in #498 as well that would be great. As I see it right now, the Provider interfaces are mutually exclusive, so if we create a default LWS PodGroup Provider interface for LWS native features, they won't be supported on the other interfaces. Unless we have a method that we force all Provider implementations to call

JesseStutler: You mean that if we want to adapt the creation of ResourceClaim and HeadlessService, directly adding an interface in the Provider would not be very convenient because the code of the already integrated custom scheduler would also need to adapt (originally only needing to implement the `CreatePodGroupIfNotExist` and `SetPodMeta` interfaces, but due to changes in the Provider, it would also need to adapt to the new interface). Actually, there will be many inheritors of the PodGroupProvider interface implementation because there are many custom gang schedulers, but the creation of ResourceClaim and HeadlessService should only have one implementation in lws. Is there any way to make the successor to PodGroupProvider have to integrate the default implementation of ResourceClaim and Headless Service? Do you have any suggestion?

JesseStutler: @Edwinhr716 Currently, what I'm thinking is to add another interface, for example, called `BaseResourceProvider`, which would include interfaces like `CreateHeadlessService` and `CreateResourceClaim`. This is because these resource implementations are fixed for lws. Then, a struct named `baseResource` would implement this interface. `ReplicaResourceProvider` would be a composite interface of `BaseResourceProvider` and `PodGroupProvider`. The new attributes added to pod-controller and pod-webhook would also be of type `ReplicaResourceProvider`. If we want to add a new implementation of a custom scheduler, it needs to implement the `ReplicaResourceProvider interface`. Either the custom scheduler directly embeds the `baseResource` struct, or it implements the `BaseResourceProvider` interface on its own. This way, if we want to provide new resources for the replica dimension in the future, we just need to modify the `BaseResourceProvider` interface. What do you think about this?

```go
// BaseResourceProvider defines the interface for managing base resources, like headless services, resource claims, etc.
type BaseResourceProvider interface {
	CreateHeadlessService(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet) error

	CreateResourceClaim(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet) error
}

// PodGroupProvider defines the interface for managing pod group resources
type PodGroupProvider interface {
	// CreatePodGroupIfNotExists creates a PodGroup if it doesn't exist, called by pod controller
	CreatePodGroupIfNotExists(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, leaderPod *corev1.Pod) error

	// SetPodMeta sets pod meta for PodGroup association, called by webhook
	SetPodMeta(pod *corev1.Pod) error
}

type ReplicaResourceProvider interface {
	BaseResourceProvider
	PodGroupProvider
}
```

```go
type baseProvider struct {
}

func (p *baseProvider) CreateHeadlessService(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet) error {
	return nil
}

func (p *baseProvider) CreateResourceClaim(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet) error {
	return nil
}
```

```go
type VolcanoProvider struct {
	*baseProvider // Implement the ReplicaResourceProvider interface with embedded baseResource struct
	client client.Client
}
```

Edwinhr716: This makes more sense, and lets us add new native LWS methods without interfering with the provider implementations

JesseStutler: Yes, if there are new replica-level resources to be created in the future, just modify the `BaseResourceProvider` interface and the implementation of `defaultBaseResourceProvider`. For new custom schedulers that need to implement `ReplicaResourceProvider` interface, just embed the `defaultBaseResourceProvider` struct and then implement the `PodGroupProvider` interface is the easiest way. Do you have any better implementation recommendations? If there are no issues, maybe I can update this design into my KEP now?

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,407 @@
...
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation. What is the desired outcome and how do we measure success?.
+The "Design Details" section below is for the real
+nitty-gritty.
+-->
+LeaderWorkerSet will create a PodGroup resource for each replica and handle the association between pods and PodGroup. When creating or updating a LeaderWorkerSet, the controller will:
+
+1. Create PodGroup resources for each replica
```
ardaguclu, line 117: What does "each replica" refer to in here?. All the pods including leader and workers?

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,407 @@
...
+A good summary is probably at least a paragraph in length.
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+This KEP aims to enable gang scheduling support for LeaderWorkerSet by creating a PodGroup for each replica.
```
ardaguclu, line 66: I think, it is better to mention this https://github.com/kubernetes/enhancements/issues/4671 in this KEP.

ardaguclu: Native `PodGroup` resource in Kubernetes would clear away the in-tree custom scheduler implementation necessity.

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,407 @@
...
+	SetPodMeta(pod *corev1.Pod) error
+}
+```
+Define a new interface called `Provider` to manage PodGroup resources. Whenever a leader pod is created, 
+the PodGroup will also be created with its OwnerReference set to the leader pod, so its lifecycle will follow the leader pod. 
+This way, whether the leader pod is deleted due to scale-down of lws, deletion of lws, or any other reason, the PodGroup will also be deleted. 
+Besides, during a lws rolling update, the PodGroup will be recreated along with the leader pod.
+- `CreatePodGroupIfNotExists`: This method will be called by the pod controller to create a PodGroup resource for each lws replica if it doesn't already exist.
+  For schedulers with PodGroup CRD (such as Volcano, coscheduling scheduler plugin), the creation logic of the corresponding PodGroup resource needs to be 
+  implemented in `CreatePodGroupIfNotExists`. For schedulers like Yunikorn that do not include PodGroup CRD, simply return nil if fine.
```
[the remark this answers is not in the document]
JesseStutler: Currently, the most popular gang scheduling frameworks in the industry are Volcano, coscheduling plugins, and Yunikorn. Yunikorn does not require creating additional PodGroup resources; it sets the group to which the pod belongs in the pod's annotations:
https://yunikorn.apache.org/docs/user_guide/gang_scheduling/#example, in my understanding, `SetPodMeta` is necessary because it sets which pods belong to a group, and `CreatePodGroupIfNotExists` is for custom schedulers like Volcano and coscheduling plugins that require additional CRDs. These two interfaces align with the current design of creating PodGroups.

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,407 @@
...
+The sequence diagram of the implementation is as follows:
+![gang-scheduling-sequence.png](gang-scheduling-sequence.png)
+This diagram omits the creation of lws. After lws is created, the leader sts will be created immediately, and then as shown in the sequence diagram:
+1. The sts-controller in kcm then creates leader pods
+2. The pod webhook will mutate the leader pod, adding the podgroup affiliation to the labels or annotations of the leader pod. The naming convention for the podgroup is `lws.Name-groupIndex`, 
+each podgroup has the same groupIndex as pods.
+3. Pod controller watches the creation of the leader pod, creates the podgroup, sets the ownerReference to the leader pod, and sets the minimum resource requirements and minAvailable count for the podgroup.
+4. Pod controller creates worker sts
+5. Sts controller watches the creation of worker sts, creates worker pods
+6. Pod webhook sets the podgroup affiliation for the worker pod as well
```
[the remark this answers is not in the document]
JesseStutler: Yes, setting annotation for the pod can indicate which pods belong to the same group

keps/407-gang-scheduling/README.md
JesseStutler: Yes, 1 leader pod + (size-1) worker pods is a group, and it means each replica

keps/407-gang-scheduling/README.md
JesseStutler: As far as I know, this KEP has been inactive for a long time. Volcano, coscheduling scheduler plugin, and Yunikorn are currently the most mature gang schedulers. Integrating a custom scheduler can help to expand the lws ecosystem more quickly, especially during this popular period for inference. It's uncertain when this KEP will be merged, and I believe that integrating custom schedulers does not conflict with this KEP. It can be added later as the default implementation :)

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,407 @@
...
+Besides, during a lws rolling update, the PodGroup will be recreated along with the leader pod.
+- `CreatePodGroupIfNotExists`: This method will be called by the pod controller to create a PodGroup resource for each lws replica if it doesn't already exist.
+  For schedulers with PodGroup CRD (such as Volcano, coscheduling scheduler plugin), the creation logic of the corresponding PodGroup resource needs to be 
+  implemented in `CreatePodGroupIfNotExists`. For schedulers like Yunikorn that do not include PodGroup CRD, simply return nil if fine.
+- `SetPodMeta`: This method will be called by pod webhook to set the PodGroup affiliation for leader pods and worker pods, 
+  as different schedulers set the PodGroup affiliation in different ways. For example, in Volcano, it is done by setting the pod's annotation: 
+  `scheduling.k8s.io/group-name`, while in coscheduling scheduler plugin, it is done by setting the pod's label: `scheduling.x-k8s.io/pod-group`.
+  Therefore, `SetPodMeta` directly passes in the pod object and sets labels and annotations for this pod
+
+### Sequence Diagram
```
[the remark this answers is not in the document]
JesseStutler: Thanks for your suggestions, I will think about it and clarify it later

JesseStutler:
@ardaguclu https://github.com/kubernetes-sigs/lws/pull/496#discussion_r2044152579, Yes, as an implementation of InTree, because there are only a few popular custom gang schedulers in the industry, and they are not many. For example, the Kubeflow training operator is InTree integration: 
- docs:
   https://www.kubeflow.org/docs/components/trainer/legacy-v1/user-guides/job-scheduling/, 
- code implementation:https://github.com/kubeflow/trainer/blob/378a9e62d18137810a6110488bebc957a879e822/pkg/controller.v1/common/job.go#L250-L335

Including Kuberay which is also InTree integration:
- docs:
  - https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/volcano.html, 
  - https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/yunikorn.html, - 
  - https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/kueue.html
- code implementation: 
  https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/batchscheduler/schedulermanager.go

So I think lws can also integrate gang scheduling in this way to expand ecosystem :)

JesseStutler:
@ardaguclu https://github.com/kubernetes-sigs/lws/pull/496#discussion_r2044152579, Yes, as an implementation of InTree, because there are only a few popular custom gang schedulers in the industry, and they are not many. For example, the Kubeflow training operator is InTree integration: 
- docs:
   https://www.kubeflow.org/docs/components/trainer/legacy-v1/user-guides/job-scheduling/, 
- code implementation:https://github.com/kubeflow/trainer/blob/378a9e62d18137810a6110488bebc957a879e822/pkg/controller.v1/common/job.go#L250-L335

Including Kuberay which is also InTree integration:
- docs:
  - https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/volcano.html, 
  - https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/yunikorn.html, - 
  - https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/kueue.html
- code implementation: 
  https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/batchscheduler/schedulermanager.go

So I think lws can also integrate gang scheduling in this way to expand ecosystem :)

keps/407-gang-scheduling/README.md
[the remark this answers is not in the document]
ardaguclu: In order to be this KEP functional, for example, user has to create separate `PodGroup` resource which is not reflected in this KEP?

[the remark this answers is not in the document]
JesseStutler: Yes, step 3 has mentioned: "3. Pod controller watches the creation of the leader pod, **creates the podgroup**, sets the ownerReference to the leader pod, and sets the minimum resource requirements and minAvailable count for the podgroup"

keps/407-gang-scheduling/README.md
JesseStutler: Yes, as an implementation of InTree, because there are only a few popular custom gang schedulers in the industry, and they are not many. For example, the Kubeflow training operator is InTree integration: 
- docs:
   https://www.kubeflow.org/docs/components/trainer/legacy-v1/user-guides/job-scheduling/, 
- code implementation:https://github.com/kubeflow/trainer/blob/378a9e62d18137810a6110488bebc957a879e822/pkg/controller.v1/common/job.go#L250-L335

Including Kuberay which is also InTree integration:
- docs:
  - https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/volcano.html, 
  - https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/yunikorn.html,
  - https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/kueue.html
- code implementation: 
  https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/batchscheduler/schedulermanager.go

So I think lws can also integrate gang scheduling in this way to expand ecosystem. The maintainers of the custom scheduler can upgrade it periodically according to versions :)

ardaguclu: Just an idea; if LWS is only supposed to create `PodGroup` resource and that's it. Why we don't create unstructured objects compatible with the whatever custom scheduler we support?. Therefore, we don't need to import. Of course, if the `PodGroup` objects are subject to change in the future, this would be an issue.

JesseStutler: Will unstructured have performance issues? When using an unstructured object, you also need to specify the apiVersion and kind of the podgroup, which makes configuring the unstructured object very complex. Moreover, the configuration of the Podgroup in Volcano and the podgroup in the coscheduling plugin are different, so it's not possible to use only one Provider. Currently, all upstream communities are integrated InTree, but I only have listed kuberay and kubeflow.

ardaguclu: > Will unstructured have performance issues?

I don't think so

> When using an unstructured object, you also need to specify the apiVersion and kind of the podgroup. And configuring the unstructured object will be very complex. Moreover, the configuration of the Podgroup in Volcano and the podgroup in the coscheduling plugin are different, so it's not possible to use only one Provider.

Provider interface can  still be implemented per each custom scheduler without importing the code. Volcano provider will create it's own compatible `PodGroup` unstructured and so on.

> Currently, all upstream communities are integrated InTree, I only have listed kuberay and kubeflow here

I don't know the others. But if there is a requirement to update, modify, etc. for `PodGroup` resources, it might make sense to import it for a better maintainability. However, from what I understand from the KEP, we just need to create it and delete it which still can be managed via unstructured without needing to create this tight coupling.

ardaguclu: I'll defer the decision to approvers.

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,407 @@
...
+
+## Design Details
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+### Interface
```
andreyvelich, line 169: In Kubeflow TrainJob which uses JobSet, we implement gang-scheduling via PodGroupPolicy API as described here: https://github.com/kubeflow/trainer/tree/master/docs/proposals/2170-kubeflow-trainer-v2#the-podgrouppolicy-api

Then, we have co-scheduling plugin that Build and Creates PodGroup for the JobSet to perform gang-scheduling: https://github.com/kubeflow/trainer/blob/master/pkg/runtime/framework/plugins/coscheduling/coscheduling.go#L105

I am curious, whether we want to re-use some of these APIs for the LWS gang-scheduling.

E.g. 

```go
type LeaderWorkerSetSpec struct {
  PodGroupPolicy *PodGroupPolicy `json:"podGroupPolicy,omitempty"`
}
```


WDYT @JesseStutler @tenzen-y @kannon92 @kerthcet @ahg-g  @akshaychitneni?

andreyvelich (commented):
Thank you for driving this @JesseStutler!

keps/407-gang-scheduling/README.md
vsoch, line 66: I would mention LeaderWorkerSet as an example, but it's not the main or only reason. Gang scheduling is immensely important for scientific (HPC) use cases where we use the message passing interface (MPI) and jobs absolutely cannot start without reaching quorum. When groups aren't taken into account you get a blocked queue - partial allocations of each group, each waiting for final members (that will never fit).

keps/407-gang-scheduling/README.md
vsoch, line 95: I would again not scope this to one specific project, but to have that as an example. Other projects that add examples make the KEP stronger I think?

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,407 @@
...
+
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation. What is the desired outcome and how do we measure success?.
+The "Design Details" section below is for the real
+nitty-gritty.
+-->
+LeaderWorkerSet will create a PodGroup resource for each replica and handle the association between pods and PodGroup. When creating or updating a LeaderWorkerSet, the controller will:
```
vsoch, line 115: I do something similar to this with having an associated mutating webhook with the operator, which is probably what you had in mind?

vsoch: Does this also assume that one pod is a group of 1 (in other words, does everything become a group)? That is a strategy we took.

vsoch: Is there more than one step here? What about in the case of a group that changes size, and what happens with cleanup or change of state of members?

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,407 @@
...
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
+As a user running distributed inference services with LeaderWorkerSet, I want to prevent resource deadlocks and ensure service availability in resource-constrained environments.
+
+Currently, when cluster resources are limited, the scheduler might only schedule leader pods while leaving worker pods pending. This results in inference services being unavailable despite consuming cluster resources.
+
+With gang scheduling support, leader and worker pods will be treated as a single scheduling unit, ensuring they are either all scheduled together or remain pending as a group. This prevents resource waste and ensures resources are only used by fully functional inference services.
```
vsoch, line 134: +1

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,407 @@
...
+
+<!--
+What are the caveats to the proposal?
+What are some important details that didn't come across above?
+Go in to as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
+
+### Risks and Mitigations
+
```
vsoch, line 148: There are cases when a pod needs to be re-created for a group, or a group needs to scale (and create new pods). How is this handled for the group? You can imagine a case where the operation is not possible (or would take a long time in the queue of pending pods) and you get something akin to blocking.

keps/407-gang-scheduling/README.md
vsoch: Could there be a common API that all projects share? This would also be useful for us.

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,407 @@
...
+type Provider interface {
+	// CreatePodGroupIfNotExists creates a PodGroup if it doesn't exist, called by pod controller
+	CreatePodGroupIfNotExists(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, leaderPod *corev1.Pod) error
+
+	// SetPodMeta sets pod meta for PodGroup association, called by webhook
+	SetPodMeta(pod *corev1.Pod) error
+}
+```
+Define a new interface called `Provider` to manage PodGroup resources. Whenever a leader pod is created, 
+the PodGroup will also be created with its OwnerReference set to the leader pod, so its lifecycle will follow the leader pod.
```
vsoch, line 181: From reading this I am guessing it is not possible for lws to scale up?

keps/407-gang-scheduling/README.md
JesseStutler: Thanks for sharing! 👍 This is also a method, some fields of PodGroup can be configured by users, but still need to introduce the api libraries of various custom schedulers, then directly create their PodGroup CRD right? Just like my implementation but can't be configured by users currently: https://github.com/kubernetes-sigs/lws/pull/498

@ardaguclu concerns that if we implemented it as InTree, it needs to maintain the api version of custom scheduler: https://github.com/kubernetes-sigs/lws/pull/496#discussion_r2046046980, whether there will be compatibility issues. I also listed kubeflow as an InTree implementation, I think lws can also do it like this way, but I don't know how kubeflow maintains the dependency versions of Volcano and coscheduling, upgrade regularly by version?

keps/407-gang-scheduling/README.md
JesseStutler: Hi, what do approvers think :)? @ahg-g @kerthcet @Edwinhr716

keps/407-gang-scheduling/README.md
JesseStutler: Other projects require gang scheduling, such as spark, flink, training job mpi, etc., but other projects have already integrated gang scheduling. My KEP is only for the lws project, so I think the goal here is reasonable, right?

vsoch: I came to this KEP from an upstream (more general) link and I thought it was being referenced as a general solution (that happened to be under this repo). My misunderstanding. I would like to see a community maintained solution.

keps/407-gang-scheduling/README.md
vsoch: Gotcha, so you can resize (and create a new set of workers or leader to generate a new group) but an existing set cannot dynamically scale.

JesseStutler: Yes, this is also designed based on the current features of lws. Otherwise, I would directly create PodGroups in the lws controller, but this would require a lot of extra handling and be more complex

keps/407-gang-scheduling/README.md
ardaguclu: This https://github.com/kubernetes/enhancements/issues/4671 can be the general solution.

JesseStutler: Yes, this is the general solution, but what about the custom scheduler? Can't https://github.com/kubernetes/enhancements/issues/4671 be the default implementation, while other scheduler can be configurable :)?

keps/407-gang-scheduling/README.md
andreyvelich: >Just like my implementation but can't be configured by users currently: https://github.com/kubernetes-sigs/lws/pull/498

Yes, that makes sense.
So rather than selecting provider via `LWS_PODGROUP_PROVIDER` envs, we could allow user to configure it via `PodGroupPolicy` API in the LWS spec.

>I don't know how kubeflow maintains the dependency versions of Volcano and coscheduling, upgrade regularly by version?

In the Kubeflow Trainer V2 we only allow to create specific version of PodGroup as you can see here: https://github.com/kubeflow/trainer/blob/master/pkg/runtime/framework/plugins/coscheduling/coscheduling.go#L135

Maybe @tenzen-y can share details on whether we want to support multiple versions of Volcano or Co-scheduling.

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,407 @@
...
+  For schedulers with PodGroup CRD (such as Volcano, coscheduling scheduler plugin), the creation logic of the corresponding PodGroup resource needs to be 
+  implemented in `CreatePodGroupIfNotExists`. For schedulers like Yunikorn that do not include PodGroup CRD, simply return nil if fine.
+- `SetPodMeta`: This method will be called by pod webhook to set the PodGroup affiliation for leader pods and worker pods, 
+  as different schedulers set the PodGroup affiliation in different ways. For example, in Volcano, it is done by setting the pod's annotation: 
+  `scheduling.k8s.io/group-name`, while in coscheduling scheduler plugin, it is done by setting the pod's label: `scheduling.x-k8s.io/pod-group`.
+  Therefore, `SetPodMeta` directly passes in the pod object and sets labels and annotations for this pod
+
+### Sequence Diagram
+The sequence diagram of the implementation is as follows:
+![gang-scheduling-sequence.png](gang-scheduling-sequence.png)
```
sunnylovestiramisu, line 194: What are the error scenarios if any of these sequences fail?

keps/407-gang-scheduling/README.md
Edwinhr716: sgtm to add the updated design to the KEP

kerthcet:
/assign

Will take another look

keps/407-gang-scheduling/README.md
JesseStutler: Sorry for the late reply, I was busy with other matters. In the sequence diagram, the only parts involving lws code modification are pod webhook and pod controller. In the pod webhook, `SetPodMeta` is called. This interface is designed to be able to return an error, but in the implementation of the custom scheduler, it only modifies pod labels or annotations, and does not involve API calls. If there is indeed more complex operations in the future that involve errors return, the kcm sts controller will retry, trying to re-create the sts pod again. For the pod controller, it involves the `CreatePodGroupIfNotExists` operation, which checks if the podgroup exists. If the podgroup does not exist, it will create one. If the creation fails, such as being disconnected from the API server, it will also perform a back-off retry, which is determined by the controller-runtime framework.

keps/407-gang-scheduling/README.md
JesseStutler: @Edwinhr716 Should we add a `PodGroupPolicy` field to lws like kubeflow? I think this would help users customize some PodGroup fields

keps/407-gang-scheduling/README.md
JesseStutler: I have updated. Please review again. @Edwinhr716 Please also help review @ardaguclu opinion, whether using  unstructured object is the better way? And the recommendation from the kubeflow maintainer @andreyvelich , do we need to add a `PodGroupPolicy` field to allow users to customize some fields of PodGroup?

keps/407-gang-scheduling/README.md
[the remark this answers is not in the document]
JesseStutler: I added a section on Error Handling, please help review it again. For the situation when there are not enough cluster resources, lws can still create podgroups, it's fine, it's just that the pods under the podgroup cannot be gang scheduled

keps/407-gang-scheduling/README.md
Edwinhr716: If I understand correctly, the proposal here is to add a field in the LWS spec instead to define the fields specific for the scheduling plugin? 

So for example, we would have the field
```
  podGroupPolicy:
       volcano:
            <fields that volcano needs specified>
spec:
    leaderWokerTemplate:
         workerTemplate:
               spec:
                     scheduler: volcano
```

My worry is if we can have multiple `podGroupPolicySource` coexist, in order to also use this field to include features such as Resource Claim Template and NetworkConfig. I'm guessing having multiple custom schedulers is not something that kubeflow has to worry about, so I think that would be the main point of contention if we want to implement something similar in LWS.

JesseStutler: I think the meaning of Kubeflow's definition of this PodGroupPolicy should be that some custom fields of PodGroup can be configured by users, and kubeflow provides a default PodGroup implementation, which has default configuration fields that users cannot manually modify: https://github.com/kubeflow/trainer/blob/master/pkg/runtime/framework/plugins/coscheduling/coscheduling.go#L140 
Perhaps we can make this field more general and define it as a Replica-level resource that can be configured by users? For example, we can call it `ReplicaResourcePolicy` or other names, and then `PodGroupPolicy` is part of this field. `ReplicaResourcePolicy` can also be used to configure custom fields of ResourceClaim or custom fields of Headless Service. What do you think?
@Edwinhr716

JesseStutler: @andreyvelich Hi Andrey, I would like to ask how kubeflow solves the problem of integrating custom scheduler and introducing API dependencies such as Volcano and coscheduling. Is it also to directly introduce dependencies in go.mod and then upgrade them regularly according to the version?

Edwinhr716: > @Edwinhr716 Please can you give an example when two schedulers should be presented for a single LWS ?

We want PodGroupPolicy to be generic so we can use the concept for other features such as ServicePerReplica and  ResourceClaim.

JesseStutler: Yes if there are better suggestions, feel free to comment @tenzen-y @astefanutti 😄 thanks. However, for now, I think the only good solution is to separate the lws API into its own repository. But I believe it's urgent to integrate the custom scheduler with lws at this stage, as we are at a critical phase for the inference API, and more inference APIs are being rolled out. Integrating lws with the custom scheduler can help lws gain a dominant position in the short term. So, should we integrate it into go.mod for now? Both kueue and Kubeflow integrate third-party dependencies this way. @Edwinhr716

JesseStutler: > > @Edwinhr716 Please can you give an example when two schedulers should be presented for a single LWS ?
> 
> We want PodGroupPolicy to be generic so we can use the concept for other features such as ServicePerReplica and ResourceClaim.

`Perhaps we can make this field more general and define it as a Replica-level resource that can be configured by users? For example, we can call it ReplicaResourcePolicy or other names, and then PodGroupPolicy is part of this field. ReplicaResourcePolicy can also be used to configure custom fields of ResourceClaim or custom fields of Headless Service. What do you think?` 

@Edwinhr716 Can we wrap the `PodGroupPolicy` field like this to also serve the resource claim template and headless service?

JesseStutler:
@Edwinhr716 @kerthcet @ardaguclu Hi, do you have any new feedback? I hope we can push this KEP forward soon, as 0.7.0 is not far away :) I'm looking forward to expanding the LWS ecosystem

keps/407-gang-scheduling/README.md
Edwinhr716: Mmm that makes more sense, so for example to run a scheduler, RCT, and headless service replica it would look like:
```
ReplicaResourcePolicy:
  podGroupPolicy:
       volcano:
            <fields that volcano needs specified>
  ResourceClaimTemplate:
       ....
   HeadlessService: PerReplica
spec:
    leaderWokerTemplate:
         workerTemplate:
               spec:
                     scheduler: volcano
```

From my side it LGTM, I'll leave it to @kerthcet for final approval

JesseStutler: @kerthcet Hi, kerthcet, would you take a look this PR? 😄  I think we should push forward this week

kerthcet:
Reviewing now!

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,447 @@
...
+-->
+
+#### Story 1
+As a user running distributed inference services with LeaderWorkerSet, I want to prevent resource deadlocks and ensure service availability in resource-constrained environments.
+
+Currently, when cluster resources are limited, the scheduler might only schedule leader pods while leaving worker pods pending. This results in inference services being unavailable despite consuming cluster resources.
+
+With gang scheduling support, leader and worker pods will be treated as a single scheduling unit, ensuring they are either all scheduled together or remain pending as a group. This prevents resource waste and ensures resources are only used by fully functional inference services.
+
+#### Story 2
```
kerthcet: remove this since no other stories.

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,447 @@
...
+What are the caveats to the proposal?
+What are some important details that didn't come across above?
+Go in to as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
+
+### Risks and Mitigations
+
+<!--
+What are the risks of this proposal, and how do we mitigate? Think broadly.
```
kerthcet, line 149: Add one risk like once more and more custom schedulers wants to integrate with lws, it will increase the maintenance cost of the code base.

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,447 @@
...
+```go
+// BaseResourceProvider defines the interface for managing base resources, like headless services, resource claims, etc.
+type BaseResourceProvider interface {
+    CreateHeadlessService(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet) error
+
+    CreateResourceClaim(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet) error
+}
+
+// PodGroupProvider defines the interface for managing pod group resources
+type PodGroupProvider interface {
```
kerthcet: SchedulerProvider seems more general.

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,447 @@
...
+    // CreatePodGroupIfNotExists creates a PodGroup if it doesn't exist, called by pod controller
+    CreatePodGroupIfNotExists(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, leaderPod *corev1.Pod) error
+
+    // SetPodMeta sets pod meta for PodGroup association, called by webhook
+    SetPodMeta(pod *corev1.Pod) error
```
kerthcet: InjectPodGroupMetadata?

keps/407-gang-scheduling/README.md
kerthcet: IIRC, volcano doesn't have timeout right?

kerthcet: I think in the future, we may have such a API for configurations:
- per LWS per scheduler support, just like podSpec.schedulerName
- populate the configurations to podGroup

but I have no idea whether it's a general one for all the schedulers or different scheduler with dedicated field just like @andreyvelich posted here. 

Maybe the final solution is a common API in upstream, but right now, I'm ok to start with all the defaulting values.

keps/407-gang-scheduling/README.md
kerthcet: I like version controlled APIs, makes me feel more stable. But maybe we should use controller flag rather than ENV since we can add validation at startup.

kerthcet (commented):
Thanks @JesseStutler for this proposal

Regarding this proposal, I have two additional questions:

- Can we start with the scheduler provider first, I saw we have a lot of new abstractions I believe not quite related to his proposal, we can develop that part in the future if needed. Also cc @Edwinhr716 who participated in the discussion before.
- Can you give more details about the integration with different startupPolicy, especially the leaderFirst.

keps/407-gang-scheduling/README.md
JesseStutler: > IIRC, volcano doesn't have timeout right?

Yes.

> Maybe the final solution is a common API in upstream, but right now, I'm ok to start with all the defaulting values.

OK I think we can start with default podgroup first, when user need to specify sepecial fields of podgroup, then we can open another KEP to discuss in the future whether to add a new API, or add a field to lws like Kubeflow

keps/407-gang-scheduling/README.md
JesseStutler: Removed, thanks for reminding me.

keps/407-gang-scheduling/README.md
JesseStutler: OK, I have added risk and mitigation, please check. As for mitigation, I suggest we can write a version compatibility documentation, maintain a clear matrix documenting LWS versions with their supported scheduler versions and PodGroup API dependencies. And if more and more schedulers are willing to integrate, we can provide clear maintainer contact information, involve the maintenance teams of these schedulers to regularly upgrade the scheduler dependency versions to reduce the pressure on the lws core team.

keps/407-gang-scheduling/README.md
JesseStutler: OK, I have modified

keps/407-gang-scheduling/README.md
JesseStutler: OK, I have modified

keps/407-gang-scheduling/README.md
JesseStutler: OK. Do we need to `enable PodGroup` as a feature gate or just directly as a controller flag? @kerthcet

JesseStutler:
Thanks for your review @kerthcet 
> * Can we start with the scheduler provider first, I saw we have a lot of new abstractions I believe not quite related to his proposal, we can develop that part in the future if needed. Also cc @Edwinhr716 who participated in the discussion before.

/cc @Edwinhr716 Please cc, @kerthcet thinks we don't need to add resourceclaim/headless service abstractions in this KEP, what do you think? But I'm already halfway through this part, I don't know whether I should delete it

> * Can you give more details about the integration with different startupPolicy, especially the leaderFirst.

OK I have added details of startup policy in `StartupPolicy Handling` part, please take a look

keps/407-gang-scheduling/README.md
```diff
@@ -0,0 +1,511 @@
...
+When an update occurs to lws, the leader pod will be recreated, and the podgroup should also be recreated. 
+During scaling down/deletion, the podgroup should also be deleted. During rolling updates, if the user has configured maxSurge, 
+then additional leader pods should be created, and corresponding additional podgroups should also be created to schedule these additional pods, 
+so that lws can perform rolling updates successfully.
+
+### StartupPolicy Handling
+
+Different `StartupPolicy` should have different MinMember settings:
+- **LeaderCreated Policy (Default):** MinMember is set to the full replica size (1 leader + (size-1) workers). All pods in the replica must be scheduled together. 
+- **LeaderReady Policy:** MinMember is set to 1, therefore leader pod can be scheduled immediately and won't be blocked by gang scheduler.
```
kerthcet, line 323: so one risk here is if the podGroup doesn't support both minMember and minResources, it can not match with startupPolicy perfectly. Let's add it as a risk.

kerthcet:
> But I'm already halfway through this part, I don't know whether I should delete it

Sorry for the delayed reply, if these are not blocking issues, I would prefer to delete them as a MVP implementation, we can move them to the alternatives for further revisit . But would like to hear @Edwinhr716 opinions as well.

kerthcet:
General LGTM. Thanks.

Edwinhr716:
> But I'm already halfway through this part, I don't know whether I should delete it
How hard do you expect it would be to retroactively add the abstraction @JesseStutler ? If it is hard, I think it would be worth it to include it now, especially since the concept of treating them as a group would be a useful to have. If it is easy, we can revisit this when implementing Resource Claim Template support

JesseStutler:
> > But I'm already halfway through this part, I don't know whether I should delete it
> > How hard do you expect it would be to retroactively add the abstraction @JesseStutler ? If it is hard, I think it would be worth it to include it now, especially since the concept of treating them as a group would be a useful to have. If it is easy, we can revisit this when implementing Resource Claim Template support

I haven't evaluated the implementation difficulty of Headless Service and Resource Claim Template yet, but I think it is easy to add "abstraction" only. It is already mid-July, and I think we need to speed up the progress. So do we need to delete `BaseResourceProvider`? Or Edwin @Edwinhr716 , do you think this abstraction does not affect KEP and is very useful? Then I will keep it.

Edwinhr716:
>  do you think this abstraction does not affect KEP and is very useful? Then I will keep it.

I don't think it affects the KEP and it is really useful. Let's keep it if you are already halfway done and completing the implementation is easy. Though you **don't** need to add anything specific for Resource Claim Template and Headless Service, we will worry about that later.

Edwinhr716:
General LGTM as well, will let @kerthcet merge it

keps/407-gang-scheduling/README.md
JesseStutler: OK I have added, please take a look

JesseStutler:
@kerthcet Please help review it, thanks. I have modified it according to the latest review opinions. Looking forward to writing codes asap 😄.

kerthcet:
I personally don't think adding the new interface is hard, also haven't see any recent plans for them, like https://github.com/kubernetes-sigs/lws/issues/444 is not on anyone's radar now, so if it's possible, would like to see a pure implementation. But wouldn't like to block the KEP, we can discuss that part during PR review. It's not a big problem to me either way.

/lgtm
/approve

Thanks @JesseStutler

PR #498, opened 2025-04-11
[WIP] Add Gang Scheduling Support for lws and volcano implementation example

JesseStutler:
#### What type of PR is this?
/kind feature

Related with https://github.com/kubernetes-sigs/lws/pull/496, provide the current implementation code for Gang Scheduling and an example code for Volcano's PodGroup provider, to facilitate everyone's review of whether the KEP implementation is reasonable. This current PR is just for reference, and the actual implementation will need to be rewritten after the KEP is finalized

Fixes #

This current PR is just for reference, the actual implementation will need to be rewritten after the KEP https://github.com/kubernetes-sigs/lws/pull/496 is finalized

```release-note

```


pkg/utils/replicaresource/volcano_provider.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+type VolcanoProvider struct {
+	*defaultBaseResourceProvider
+	client client.Client
+}
+
+func NewVolcanoProvider(client client.Client) *VolcanoProvider {
+	return &VolcanoProvider{
+		client: client,
```
Edwinhr716, line 42: Missing the `defaultBaseResourceProvider` right? Which would be the same for every provider

JesseStutler: Yes, you are right. I add it now.

JesseStutler:
/cc @Edwinhr716 @kerthcet 
Please review it thanks, I'm still following the design in https://github.com/kubernetes-sigs/lws/pull/496, using the ReplicaResourceProvider interface. The CreateHeadlessService and CreateResourceClaim methods are reserved, please check whether you want to delete these two methods and keep only the SchedulerProvider interface in this PR

JesseStutler:
I added a volcano implementation for you to refer to. I can also add coscheduling plugin implementation if we also need it in v0.7.0 😄

kerthcet:
Will finish the review today, sorry I'm a bit busy these days.

api/config/v1alpha1/configuration_types.go
```diff
@@ -55,6 +55,15 @@ type ControllerManager struct {
...
 	Health ControllerHealth `json:"health,omitempty"`
+
+	// EnableGangScheduling controls whether to enable gang-scheduling.
+	// +optional
+	EnableGangScheduling bool `json:"enableGangScheduling,omitempty"`
+
+	// SchedulerProvider is the name of the scheduler that provides gang-scheduling capabilities.
+	// This is required if EnableGangScheduling is true.
+	// +optional
+	SchedulerProvider string `json:"schedulerProvider,omitempty"`
```
kerthcet: Can we use a struct pointer instead? Like 
```
type GangSchedulingManagement struct {
  schedulerProvider *string
}
```

Similar to InternalCertManagement, if `gangSchedulingManagement` not nil then we enable the gang scheduling.  We can extend the field in the future if needed.

kerthcet: Also can we put it under the `Configuration`?

charts/lws/templates/manager/deployment.yaml
```diff
@@ -37,6 +37,10 @@ spec:
         - args:
           - --config=/controller_manager_config.yaml
           - --zap-log-level=2
+          {{- if .Values.gangScheduling.enable }}
```
kerthcet: Shouldn't we use the --config instead, do we still need this?

charts/lws/README.md
```diff
@@ -65,24 +65,26 @@ for more information on installing LWS with metrics using our Helm chart.
...
+| `service.type`                             | Type of lws controller service                 | `ClusterIP`                                       |
+| `service.port`                             | Lws controller service port                    | `9443`                                            |
+| `resources.requests.cpu`                   | CPU request for resources                      | `1`                                               |
+| `resources.requests.memory`                | Memory request for resources                   | `1Gi`                                             |
+| `nodeSelector`                             | Node selector                                  | `{}`                                              |
+| `tolerations`                              | Tolerations                                    | `{}`                                              |
+| `affinity`                                 | Affinity                                       | `{}`                                              |
+| `gangScheduling.enable`                    | enable gang scheduling capabilities            | `false`                                           |
```
kerthcet: If we agree with the field change above, I think we should polish here as well.

cmd/main.go
```diff
@@ -103,6 +110,8 @@ func main() {
 		"The controller will load its initial configuration from this file. "+
 			"Command-line flags will override any configurations set in this file. "+
 			"Omit this flag to use the default configuration values.")
+	flag.BoolVar(&enableGangScheduling, "enable-gang-scheduling", false, "Enable gang-scheduling for a group of lws pods [ 1 leader pod + (size-1) worker pods].")
```
kerthcet: I think we encourage users to use the config file instead, so maybe we don't need this here any more.

docs/examples/sample/gang-scheduling/README.md
```diff
@@ -0,0 +1,57 @@
+# Enabling Gang Scheduling with Custom Schedulers
+
+This document provides guidance for how to enable gang scheduling capabilities in LeaderWorkerSet using custom schedulers.
```
kerthcet:
```suggestion
This document provides guidance about how to enable gang scheduling capabilities in LeaderWorkerSet with different schedulers.
```

pkg/utils/replicaresource/default_provider.go
```diff
@@ -0,0 +1,52 @@
...
+	leaderworkerset "sigs.k8s.io/lws/api/leaderworkerset/v1"
+)
+
+// The defaultBaseResourceProvider is the default implementation for creating resources like HeadlessService, ResourceClaim, etc.
+// Generally, in lws, there is only one implementation for creating these resources, and you can embed defaultBaseResourceProvider into the newly written replica resource provider,
+// you can refer to the implementation in volcano_provider.go.
+type defaultBaseResourceProvider struct {
+}
+
+func (d *defaultBaseResourceProvider) CreateHeadlessService(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet) error {
```
kerthcet: Should we move the headlessservice creation logic to this place in the future? Like we always have a default provider?

pkg/controllers/pod_controller.go
```diff
@@ -116,6 +134,13 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 		}
 	}
 
+	if r.replicaResourceProvider != nil {
```
kerthcet: Will this be nil? Will we always have a default provider?

pkg/utils/replicaresource/utils.go
```diff
@@ -0,0 +1,54 @@
...
+	SupportedSchedulerProviders = sets.New("volcano")
+)
+
+const (
+	PodGroupNameFmt = "%s-%s-%s"
+)
+
+// GetPodGroupName returns the name of the PodGroup for a given LeaderWorkloadSet name, group index and revision (leaderworkerset.sigs.k8s.io/template-revision-hash).
+func GetPodGroupName(lwsName, groupIndex, revision string) string {
+	return fmt.Sprintf(PodGroupNameFmt, lwsName, groupIndex, revision)
```
kerthcet: Will the name be too long? Will it lead to any problem if we only use `<lwsname>-<groupIndex>`?

pkg/utils/replicaresource/volcano_provider.go
```diff
@@ -0,0 +1,123 @@
...
+		// If no leader template is specified, use worker template.
+		leaderTemplate = &lws.Spec.LeaderWorkerTemplate.WorkerTemplate
+	}
+	totalResources := resource.PodRequests(&corev1.Pod{Spec: leaderTemplate.Spec}, resource.PodResourcesOptions{})
+
+	// Calculate and add worker resources for (size-1) workers.
+	workerCount := ptr.Deref(lws.Spec.LeaderWorkerTemplate.Size, 1) - 1
+	if workerCount > 0 {
+		workerResources := resource.PodRequests(&corev1.Pod{Spec: lws.Spec.LeaderWorkerTemplate.WorkerTemplate.Spec}, resource.PodResourcesOptions{})
+		for i := int32(0); i < workerCount; i++ {
```
kerthcet: Can we use workerResources * (size-1) here?

pkg/utils/replicaresource/volcano_provider.go
```diff
@@ -0,0 +1,123 @@
...
+		// as at least one Worker will definitely not be able to be scheduled.
+		if lws.Spec.StartupPolicy == leaderworkerset.LeaderReadyStartupPolicy {
+			pg.Spec.MinMember = 1
+		}
+
+		if queueName, ok := lws.Annotations[volcanov1beta1.QueueNameAnnotationKey]; ok {
+			pg.Spec.Queue = queueName
+		}
+
+		err = ctrl.SetControllerReference(leaderPod, &pg, v.client.Scheme())
```
kerthcet: We should have integration test to cover that podGroup is created successfully once lws is created, also cleaned up once lws is deleted.

kerthcet (commented):
Thanks @JesseStutler for this, it's awesome. But seems we don't have integration tests?

pkg/utils/replicaresource/default_provider.go
JesseStutler: Yes, do you accept `CreateHeadlessService` and `CreateResourceClaim` methods and this `defaultBaseResourceProvider`?  Because I think there will only be one implementation of `CreateHeadlessService` and `CreateResourceClaim` for lws, so I let `ReplicaresourceProvider` interface Implementer embed this `defaultBaseResourceProvide` struct and implement its own `SchedulerProvider`. If you think this implementation should not be in this PR or the implementation is not good, then I can also delete it in this PR and refactor it later.

pkg/utils/replicaresource/utils.go
JesseStutler: I used `<lwsname>-<groupIndex>` before. However, during a rolling upgrade, the leader pod will be rebuilt and the podgroup will be rebuilt as well. I am not sure if this will happen:
1. The podgroup is still being deleted
2. The new leader pod has been rebuilt, and the new podgroup also needs to be recreated
3. A podgroup with the same name already exists (it is still being deleted in step 1), so a new podgroup is not created
4. Podgroup deletion completed
5. The new leader pod and worker pods cannot be scheduled because they have no podgroup and remain pending.
6. Rolling update interrupt
If I change to `<lwsname>-<groupIndex>-<revision>`. At least the podgroup will not have the same name before and after the rebuild, so this situation will not happen.

pkg/controllers/pod_controller.go
JesseStutler: If gang scheduling is not enabled, here replicaResourceProvider will be nil. So here need to use if `r.SchedulerProvider != nil` to check

pkg/utils/replicaresource/volcano_provider.go
JesseStutler: Sorry I don't understand, you mean minResources == `workerResources` * (size-1) ? But we don't consider the resource requirement of the leader pod? I thought the resource requirement of leader pod is important

kerthcet:
Thanks for your response @JesseStutler I'll take a look later today.

JesseStutler:
> Thanks for your response @JesseStutler I'll take a look later today.

Thanks 😄 , I'm going to add integration tests in recent days, and some comments I have not responsed yet but I already modified it in my laptop @kerthcet

pkg/utils/replicaresource/default_provider.go
kerthcet: If not a big effort, let's remove it. 😏

api/config/v1alpha1/configuration_types.go
JesseStutler: OK, I have modified it, please check

charts/lws/templates/manager/deployment.yaml
JesseStutler: OK, I have removed these commend args, gang scheduling configuration can be enabled in lws-manager-config configmap instead

charts/lws/README.md
JesseStutler: OK I have modified it, please check

cmd/main.go
JesseStutler: OK, I have removed these commend args, gang scheduling configuration can be enabled in lws-manager-config configmap instead

docs/examples/sample/gang-scheduling/README.md
JesseStutler: OK, I have modified it

pkg/utils/replicaresource/default_provider.go
JesseStutler: OK, only SchedulerProvider interface is kept now

pkg/utils/replicaresource/volcano_provider.go
JesseStutler: OK. I have added several integration tests in both webhook and controller, please check. Regarding the controller, now I added two use cases:
1. After LWS is created, podgroups also need to be created, and the number should be the same as spec.replicas
2. For rolling updates, the leader pod needs to be rebuilt, and the podgroup also needs to be rebuilt.

Regarding deleting the lws, the podgroup also needs to be deleted. I feel that this use case is not very meaningful in integration tests? Because we don't have kube-controller-manager, we can't automatically delete the podgroup( no gc controler). All leader pods and podgroups must be deleted manually. I feel that this does not reflect the semantics that after deleting the lws, the podgroup is also "automatically" deleted.

pkg/webhooks/pod_webhook.go
```diff
@@ -31,16 +31,23 @@ import (
...
+type PodWebhook struct {
+	SchedulerProvider schedulerprovider.SchedulerProvider
+}
+
+func NewPodWebhook() *PodWebhook {
+	return &PodWebhook{}
+}
 
-func SetupPodWebhook(mgr ctrl.Manager) error {
+func (p *PodWebhook) SetupPodWebhook(mgr ctrl.Manager) error {
```
ardaguclu: I understand that the reason behind this change. But curious about is this the conventional way of dealing with this?.

JesseStutler: The reason is this: Actually, I didn't change this method initially. Instead, I initialized the `SchedulerProvider` by passing a `Configuration` parameter. Since I need to add gang scheduling integration tests, but I don't want to modify the old use cases and avoid affecting them, I expected to write a `BeforeAll` in my test case so that the `SchedulerProvider` can be set up after the `PodWebhook` is initialized: https://github.com/JesseStutler/lws/blob/4889aec939fbd182580cf2d92848ad70645cea46/test/integration/webhooks/pod_test.go#L870-L872. So, I want to add a `NewPodWebhook` method and also changed the `SetupPodWebhook` to setup mgr after webhook initialized

charts/lws/README.md
```diff
@@ -65,24 +65,25 @@ for more information on installing LWS with metrics using our Helm chart.
...
+| `resources.requests.memory`                | Memory request for resources                                                                                                           | `1Gi`                                               |
+| `nodeSelector`                             | Node selector                                                                                                                          | `{}`                                                |
+| `tolerations`                              | Tolerations                                                                                                                            | `{}`                                                |
+| `affinity`                                 | Affinity                                                                                                                               | `{}`                                                |
+| `gangSchedulingManagement`                 | Configuration for gang scheduling. When set the scheduler name, gang scheduling will be enabled. For example: `schedulerName: volcano` | `{}`                                                |
```
kerthcet:
```suggestioin
| `gangSchedulingManagement`                 | Configuration for gang scheduling.  | `{}`      
```

Let's be simple here, because the description is a bit unclear to me.

charts/lws/values.yaml
```diff
@@ -48,3 +48,9 @@ resources:
 nodeSelector: {}
 tolerations: []
 affinity: {}
+
+# gangSchedulingManagement is configuration for gang scheduling management.
+# For example:
+# gangSchedulingManagement:
+#   schedulerName: volcano
```
kerthcet: It should be schedulerProvider, but actually maybe schedulerName is not a bad choice, the schedulerProvider is the interface we defined. WDYT?

cmd/main.go
```diff
@@ -186,18 +190,37 @@ func setupControllers(mgr ctrl.Manager, certsReady chan struct{}) {
...
+	// Set up scheduler provider
+	var sp schedulerprovider.SchedulerProvider
+	if cfg.GangSchedulingManagement != nil {
+		var err error
+		sp, err = schedulerprovider.NewSchedulerProvider(schedulerprovider.ProviderType(*cfg.GangSchedulingManagement.SchedulerProvider), mgr.GetClient())
+		if err != nil {
+			setupLog.Error(err, "unable to create scheduler provider", "provider", *cfg.GangSchedulingManagement.SchedulerProvider)
+			os.Exit(1)
+		}
+		setupLog.Info("Enabled gang scheduling", "provider", *cfg.GangSchedulingManagement.SchedulerProvider)
```
kerthcet:
```suggestion
		setupLog.Info("Gang scheduling enabled", "provider", *cfg.GangSchedulingManagement.SchedulerProvider)
```

pkg/config/validation.go
```diff
@@ -24,18 +24,36 @@ import (
...
+	allErrs = append(allErrs, validateSchedulerProvider(c)...)
 	allErrs = append(allErrs, validateInternalCertManagement(c)...)
 	return allErrs
 }
 
+func validateSchedulerProvider(c *configapi.Configuration) field.ErrorList {
+	var allErrs field.ErrorList
+	if c.GangSchedulingManagement != nil {
+		if c.GangSchedulingManagement.SchedulerProvider == nil || *c.GangSchedulingManagement.SchedulerProvider == "" {
+			allErrs = append(allErrs, field.Required(schedulerProviderPath, "must be set when enableGangScheduling is true"))
```
kerthcet:
```suggestion
			allErrs = append(allErrs, field.Required(schedulerProviderPath, "must be set when gang scheduling is enabled"))
```

pkg/controllers/pod_controller.go
```diff
@@ -116,6 +118,13 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 		}
 	}
 
+	if r.SchedulerProvider != nil {
+		err = r.SchedulerProvider.CreatePodGroupIfNotExists(ctx, &leaderWorkerSet, &pod)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
```
kerthcet: Move below L131, once leading Pod is deleting, no need to create pod group then.

pkg/webhooks/pod_webhook.go
kerthcet: I'm ok with this, just different way to initialize the webhook, can we rename to Setup then?

README.md
```diff
@@ -22,6 +22,7 @@ Read the [documentation](https://lws.sigs.k8s.io/docs/) or watch the LWS-related
 - **Group of Pods as a unit:** Supports a tightly managed group of pods that represent a “super pod”
   - **Unique pod identity:** Each pod in the group has a unique index from 0 to n-1.
   - **Parallel creation:** Pods in the group will have the same lifecycle and be created in parallel.
+  - **Gang Scheduling:** The group of pods can be scheduled in an all-or-nothing manner.
```
kerthcet:
```suggestion
  - **Gang Scheduling:** Each replica with a group of pods can be scheduled in an all-or-nothing manner.
```

cmd/main.go
```diff
@@ -186,18 +190,37 @@ func setupControllers(mgr ctrl.Manager, certsReady chan struct{}) {
...
+	}
 	// Set up pod reconciler.
 	podController := controllers.NewPodReconciler(mgr.GetClient(), mgr.GetScheme(), mgr.GetEventRecorderFor("leaderworkerset"))
```
kerthcet: we can pass the sp to the NewPodReconciler instead?

cmd/main.go
```diff
@@ -186,18 +190,37 @@ func setupControllers(mgr ctrl.Manager, certsReady chan struct{}) {
...
 		os.Exit(1)
 	}
+	// Set up webhooks
 	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
 		if err := webhooks.SetupLeaderWorkerSetWebhook(mgr); err != nil {
 			setupLog.Error(err, "unable to create leaderworkerset webhook", "webhook", "LeaderWorkerSet")
 			os.Exit(1)
 		}
-		if err := webhooks.SetupPodWebhook(mgr); err != nil {
+		pw := webhooks.NewPodWebhook()
```
kerthcet: ditto

docs/examples/sample/gang-scheduling/README.md
```diff
@@ -0,0 +1,79 @@
...
+This document provides guidance about how to enable gang scheduling capabilities in LeaderWorkerSet with different schedulers.
+
+## Supported Schedulers
+
+| Custom Scheduler Community | GitHub ID       | Email                           |
+| -------------------------- |-----------------|---------------------------------|
+| Volcano (`@volcano-sh`)    | `@JesseStutler` | `jesseincomparable@hotmail.com` |
+| Your-Scheduler-Here        | `@your-id`      | `your@email.com`                |
+
+We welcome contributions for integrating more custom schedulers that support gang scheduling capabilities. When developing an integration, please:
```
kerthcet:
```suggestion
We welcome integrations for more custom schedulers that support gang scheduling capabilities. Getting started with:

File: pkg/utils/schedulerprovider/interface.go [hunk 22]
```diff
@@ -0,0 +1,34 @@
+/*
+Copyright 2025.
```
kerthcet: Can we just move this to pkg/schedulerprovider? Since it's used by both controller and webhook.  Utils is used for general functions, I think not fit this.

pkg/utils/schedulerprovider/utils.go
```diff
@@ -0,0 +1,54 @@
...
+// GetPodGroupName returns the name of the PodGroup for a given LeaderWorkloadSet name, group index and revision (leaderworkerset.sigs.k8s.io/template-revision-hash).
+func GetPodGroupName(lwsName, groupIndex, revision string) string {
+	return fmt.Sprintf(PodGroupNameFmt, lwsName, groupIndex, revision)
+}
+
+// ProviderType defines the type of scheduler provider
+type ProviderType string
+
+const (
+	Volcano ProviderType = "volcano"
```
kerthcet: Let's move this to volcano provider.

pkg/utils/schedulerprovider/utils.go
```diff
@@ -0,0 +1,54 @@
+/*
+Copyright 2025.
```
kerthcet: I'm move inclined to move this to interface.go directly, not a complex one.

test/integration/controllers/leaderworkerset_test.go
```diff
@@ -2195,6 +2196,163 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
...
+					ns := &corev1.Namespace{
+						ObjectMeta: metav1.ObjectMeta{
+							GenerateName: "lws-gang-ns-",
+						},
+					}
+					gomega.Expect(k8sClient.Create(ctx, ns)).To(gomega.Succeed())
+
+					// Create LeaderWorkerSet with volcano scheduler
+					lws := tc.makeLeaderWorkerSet(ns.Name).Obj()
+					lws.Spec.LeaderWorkerTemplate.LeaderTemplate.Spec.SchedulerName = "volcano"
```
kerthcet: Add a check here once leaderTemplate == nil, skip.

test/testutils/util.go
```diff
@@ -159,39 +171,67 @@ func CreateLeaderPodsFromRevisionNumber(ctx context.Context, leaderSts appsv1.St
...
+	// Set the controller owner reference for garbage collection and reconciliation.
+	if err := ctrl.SetControllerReference(&leaderSts, pod, scheme.Scheme); err != nil {
+		return nil, err
+	}
+
+	return pod, nil
+}
+
+// createLeaderPodsWithInjectedMeta creates leader pods with injected metadata using the provided inject function.
+func createLeaderPodsWithInjectedMeta(ctx context.Context, k8sClient client.Client, leaderSts appsv1.StatefulSet, lws *leaderworkerset.LeaderWorkerSet, revisionKey string, start, end int, injectFn func(*corev1.Pod)) error {
```
kerthcet: createLeaderPodsWithInjectFn?

test/testutils/util.go
```diff
@@ -141,6 +144,15 @@ func CreateLeaderPods(ctx context.Context, leaderSts appsv1.StatefulSet, k8sClie
 	return createLeaderPods(ctx, k8sClient, leaderSts, lws, revisionutils.GetRevisionKey(cr), start, end)
 }
 
+func CreateLeaderPodsWithInjectedMeta(ctx context.Context, leaderSts appsv1.StatefulSet, k8sClient client.Client, lws *leaderworkerset.LeaderWorkerSet, start int, end int, injectFn func(*corev1.Pod)) error {
```
kerthcet: CreateLeaderPodsWithInjectFn?

test/testutils/util.go
```diff
@@ -696,3 +736,156 @@ func SetLwsPartition(ctx context.Context, k8sClient client.Client, lws *leaderwo
...
+	gomega.Expect(ownerRefMap["uid"]).To(gomega.Equal(string(leaderPod.UID)))
+
+	controller, found, err := unstructured.NestedBool(ownerRefMap, "controller")
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	gomega.Expect(found).To(gomega.BeTrue())
+	gomega.Expect(controller).To(gomega.BeTrue())
+}
+
+// validatePodGroupSpec validates the PodGroup spec using unstructured access
+func validatePodGroupSpec(podGroup *unstructured.Unstructured, lws *leaderworkerset.LeaderWorkerSet) {
```
kerthcet, line 777: Should we consider the leader creation policy here?

test/testutils/util.go
```diff
@@ -696,3 +736,156 @@ func SetLwsPartition(ctx context.Context, k8sClient client.Client, lws *leaderwo
...
+		gomega.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: expectedPgName, Namespace: lws.Namespace}, podGroup)).To(gomega.Succeed())
+
+		// Validate podgroup owner reference and spec
+		validateOwnerReference(podGroup, leaderPod)
+		validatePodGroupSpec(podGroup, lws)
+	}
+}
+
+// UpdatePodGroupAtIndex manually updates the PodGroup for a specific group index
+func UpdatePodGroupAtIndex(ctx context.Context, k8sClient client.Client, schedulerProvider schedulerprovider.SchedulerProvider, lws *leaderworkerset.LeaderWorkerSet, groupIndex string) {
```
kerthcet: We have mixed the podGroup names in the file, we may modify the name like `SetPodGroupToReady` to `SetSuperPodToReady` in the follow up.

pkg/utils/replicaresource/volcano_provider.go
kerthcet: fair, we need a e2e test.

pkg/utils/replicaresource/utils.go
kerthcet: Yes, we should make sure the podGroup has the same lifecycle as the leaderpod. Would you mind to paste a real podGroup name here? Thanks.

kerthcet (commented):
Thanks @JesseStutler The total logic is very clear. I left some comments there I think we can catch up with the 0.7.0. Besides, I think we should:
- add one e2e test to make sure it works as expected with volcano

charts/lws/values.yaml
JesseStutler: Actually, I wrote it wrong here. I wrote schedulerProvider in Configuration. Let me correct it.

cmd/main.go
JesseStutler: This is still related to Integration Tests. In suite_test.go, `BeforeSuite` starts a global `PodReconciler`. In order to avoid affecting the old use cases, I don’t want to initialize the `SchedulerProvider` at the beginning. So I set up the sp for PodReconciler later. Or, following your approach, I can directly pass in sp here, but in the integration tests, I can pass nil in the `BeforeSuite`first and then initialize it after. WDYT?

cmd/main.go
JesseStutler: The reason is similar to above, https://github.com/kubernetes-sigs/lws/pull/498#discussion_r2246701786, WDYT?

test/testutils/util.go
JesseStutler: Since the normal `MakeXXXPodSpec` in integration test does not request for resources, the MinMember of `LeaderReady` and `LeaderCreated` are actually the same. I think it is more appropriate to add it to e2e tests.

test/testutils/util.go
JesseStutler: Yes I'm also confused about this `SetPodGroupToReady` func, I have renamed it now

pkg/webhooks/pod_webhook.go
JesseStutler: > I'm ok with this, just different way to initialize the webhook, can we rename to Setup then?

OK I have modified it, please check

charts/lws/README.md
JesseStutler: OK I have modified it, please check

cmd/main.go
JesseStutler: OK I have modified it, please check

pkg/config/validation.go
JesseStutler: OK I have modified it, please check

pkg/controllers/pod_controller.go
JesseStutler: Good suggestion, I have modified it, please check thanks

README.md
JesseStutler: OK I have modified it, please check

docs/examples/sample/gang-scheduling/README.md
JesseStutler: OK I have modified it, please check

pkg/utils/schedulerprovider/interface.go
JesseStutler: OK I have modified it, please check

pkg/utils/schedulerprovider/utils.go
JesseStutler: OK I have modified it, please check

pkg/utils/schedulerprovider/utils.go
JesseStutler: OK I have modified it, please check

test/integration/controllers/leaderworkerset_test.go
JesseStutler: OK I have added a judgement, if `lws.Spec.LeaderWorkerTemplate.LeaderTemplate` isn't nil, then should add volcano schedulerName for the leaderTemplate

test/testutils/util.go
JesseStutler: OK I have renamed it, please check

test/testutils/util.go
JesseStutler: OK I have renamed it, please check

pkg/utils/replicaresource/utils.go
JesseStutler: Sure, I have added a comment:
`// An example can be like "lws-1-dd6699c7c", where "lws" is the LeaderWorkerSet name, "1" is the group index, and "dd6699c7c" is the revision hash.`

JesseStutler:
> Thanks @JesseStutler The total logic is very clear. I left some comments there I think we can catch up with the 0.7.0. Besides, I think we should:
> 
> * add one e2e test to make sure it works as expected with volcano
> * add another e2e test with resizePolicy=Recreate and schedulerProvider=volcano, try to make sure it works well with the resize feature.
> 
> Once ready, I think we can merge then.

Sure, 

> Thanks @JesseStutler The total logic is very clear. I left some comments there I think we can catch up with the 0.7.0. Besides, I think we should:
> 
> * add one e2e test to make sure it works as expected with volcano
> * add another e2e test with resizePolicy=Recreate and schedulerProvider=volcano, try to make sure it works well with the resize feature.
> 
> Once ready, I think we can merge then.

But if so, I need to install volcano in e2e tests as well, would the lws community accept it? Should I separate an e2e test in the Makefile? @kerthcet

cmd/main.go
kerthcet: >  I can directly pass in sp here, but in the integration tests, I can pass nil in the BeforeSuitefirst and then initialize it after. WDYT?

Yeah, that's what I'm thinking about.

README.md
```diff
@@ -22,6 +22,7 @@ Read the [documentation](https://lws.sigs.k8s.io/docs/) or watch the LWS-related
 - **Group of Pods as a unit:** Supports a tightly managed group of pods that represent a “super pod”
   - **Unique pod identity:** Each pod in the group has a unique index from 0 to n-1.
   - **Parallel creation:** Pods in the group will have the same lifecycle and be created in parallel.
+  - **Gang Scheduling:** Each replica with a group of pods can be scheduled in an all-or-nothing manner.
```
kerthcet:
```suggestion
  - **Gang Scheduling:** Each replica with a group of pods can be scheduled in an all-or-nothing manner (Alpha level, API may change in the future).
```

kerthcet:
@JesseStutler can we finish this week? I don't want to last until next week. Thanks.

JesseStutler:
> @JesseStutler can we finish this week? I don't want to last until next week. Thanks.

Sorry for the late reply, I have something to do on the weekend. Can I finish it on Monday and ask you to review again if it's okay? @kerthcet

JesseStutler:
> @JesseStutler can we finish this week? I don't want to last until next week. Thanks.

@kerthcet You can review first about the e2e testing, I'll verify it on Monday. Is that OK?

test/e2e/e2e_test.go
```diff
@@ -450,6 +451,116 @@ var _ = ginkgo.Describe("leaderWorkerSet e2e tests", func() {
 		_, err = testutils.Run(cmd)
 		gomega.Expect(err).To(gomega.BeNil())
 	})
+
+	ginkgo.Context("with gang scheduling enabled", ginkgo.Ordered, func() {
```
kerthcet: Can we move to a different file?

test/testutils/util.go
```diff
@@ -776,11 +779,61 @@ func validatePodGroupSpec(podGroup *unstructured.Unstructured, lws *leaderworker
...
+
+		actualQuantity, err := resource.ParseQuantity(actualQuantityStr.(string))
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		gomega.Expect(actualQuantity.Equal(expectedQuantity)).To(gomega.BeTrue(),
+			"Resource %s: expected %s, got %s", resourceName, expectedQuantity.String(), actualQuantity.String())
+	}
+}
+
+// calculateMinResources calculates the minimum resources needed for the entire PodGroup [1 Leader + (size-1) Worker pods]
+// This mirrors the logic in volcano_provider.go getMinResources method
```
kerthcet: Since a mirror, can we move to util file?

test/e2e/e2e_test.go
```diff
@@ -450,6 +451,116 @@ var _ = ginkgo.Describe("leaderWorkerSet e2e tests", func() {
...
+					Size(3).
+					SchedulerName("volcano").
+					Obj()
+				testing.MustCreateLws(ctx, k8sClient, lws)
+				testing.ExpectLeaderWorkerSetAvailable(ctx, k8sClient, lws, "All replicas are ready")
+
+				// Get initial revision
+				var leaderSts appsv1.StatefulSet
+				gomega.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: lws.Name, Namespace: lws.Namespace}, &leaderSts)).To(gomega.Succeed())
+				initialRevision := leaderSts.Labels[v1.RevisionKey]
```
kerthcet: Can we move fetching revision to the `ExpectPodGroupsCountWithRevision` as well? Or can we move all of them to `ExpectValidPodGroups`?

test/e2e/e2e_test.go
```diff
@@ -450,6 +451,116 @@ var _ = ginkgo.Describe("leaderWorkerSet e2e tests", func() {
...
+					Obj()
+				testing.MustCreateLws(ctx, k8sClient, lws)
+				testing.ExpectLeaderWorkerSetAvailable(ctx, k8sClient, lws, "All replicas are ready")
+				// Verify initial PodGroups with size=2
+				testing.ExpectValidPodGroups(ctx, k8sClient, schedulerprovider.Volcano, lws, 2)
+				// Resize to size=4
+				testing.UpdateSize(ctx, k8sClient, lws, int32(4))
+				testing.ExpectLeaderWorkerSetAvailable(ctx, k8sClient, lws, "All replicas are ready")
+				// Verify PodGroups are recreated with new size
+				testing.ExpectValidPodGroups(ctx, k8sClient, schedulerprovider.Volcano, lws, 2)
```
kerthcet: Why the account is still 2? If the test passed, is that a bug?

pkg/schedulerprovider/volcano_provider.go
```diff
@@ -60,6 +60,11 @@ func (v *VolcanoProvider) CreatePodGroupIfNotExists(ctx context.Context, lws *le
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      pgName,
 				Namespace: lws.Namespace,
+				Labels: map[string]string{
+					leaderworkerset.SetNameLabelKey:    lws.Name,
```
kerthcet, line 62: Seems some fields set again in the `InjectPodGroupMetadata`. Better to move to here.

hack/e2e-test.sh
```diff
@@ -74,13 +102,24 @@ leaderElection:
...
     else
-      cd $CWD/config/manager && $KUSTOMIZE edit set image controller=$IMAGE_TAG
       $KUSTOMIZE build $CWD/test/e2e/config | $KUBECTL apply --server-side -f -
     fi
 }
+
+function run_tests() {
+    if [ -n "$SCHEDULER_PROVIDER" ]; then
+        # Run gang scheduling tests
+        $GINKGO --junit-report=junit.xml --output-dir=$ARTIFACTS -v --focus="with gang scheduling enabled" $CWD/test/e2e/...
```
kerthcet:
```suggestion
        $GINKGO --junit-report=junit.xml --output-dir=$ARTIFACTS -v --focus="gang scheduling enabled" $CWD/test/e2e/...
```

hack/e2e-test.sh
```diff
@@ -74,13 +102,24 @@ leaderElection:
...
+    else
+        # Run normal tests, skip gang scheduling tests
+        $GINKGO --junit-report=junit.xml --output-dir=$ARTIFACTS -v --skip="with gang scheduling enabled" $CWD/test/e2e/...
```
kerthcet: ditto

test/e2e/e2e_test.go
kerthcet:
```suggestion
	ginkgo.Context("with volcano gang scheduling enabled", ginkgo.Ordered, func() {
```

pkg/schedulerprovider/volcano_provider.go
JesseStutler: This is added when creating a PodGroup in order to facilitate the use of label to find pg in the future. `InjectPodGroupMetadata` is used to inject the podgroup information to the pod

kerthcet: Sorry, I misread the code.

test/e2e/e2e_test.go
JesseStutler: I have removed it to `test/e2e/e2e_gang_scheduling_test.go` now

test/testutils/util.go
JesseStutler: OK, now I have removed it to `pkg/utils/utils.go`

test/e2e/e2e_test.go
JesseStutler: OK, I have removed all the logic into `ExpectValidPodGroups` and deleted `ExpectPodGroupsCountWithRevision`, revision will get first in the `ExpectValidPodGroups`

test/e2e/e2e_test.go
JesseStutler: I think actually size update scenario is the same as rolling update? Now I have added some logic, just like in some other rolling update use cases, to ensure that after the rolling update is completed, the number of pods is as expected, and the number of podgroups for the new revision is also as expected.

hack/e2e-test.sh
JesseStutler: Since I have moved all the gang scheduling related tests into a new file, now my current use case is organized like this:
``` 
ginkgo.Describe("leaderWorkerSet e2e gang scheduling tests", func() {
    ginkgo.Context("with volcano gang scheduling enabled", ginkgo.Ordered, func() {
           ....
    })
})
```
So I changed the focus description to: "leaderWorkerSet e2e gang scheduling tests" is that okay?

hack/e2e-test.sh
JesseStutler: Just like this: https://github.com/kubernetes-sigs/lws/pull/498#discussion_r2251162251, is that ok?

test/e2e/e2e_test.go
JesseStutler: OK I have modified it, please check

JesseStutler:
> Once you fix the env error in the e2e test, I'll create a new ci workflow in test-infra, so we can verify the gang scheduling tests.

@kerthcet Now I have fixed all the CIs, please help me create a workflow, thanks. Does this require codes to merge?

README.md
JesseStutler: I have changed it now, please check

test/e2e/e2e_test.go
kerthcet: Since we changed the size to 4, shouldn't we have 4 podGroups?

test/testutils/util.go
```diff
@@ -696,3 +738,194 @@ func SetLwsPartition(ctx context.Context, k8sClient client.Client, lws *leaderwo
...
+	}
+}
+
+// ExpectValidPodGroups validates that PodGroups are correctly created for a LeaderWorkerSet with proper owner references and spec configuration.
+// When expectedCount is 0, it validates that no PodGroups exist for the LWS, checking cleanup.
+// When expectedCount > 0, it validates both count and detailed configuration of each PodGroup.
+func ExpectValidPodGroups(ctx context.Context, k8sClient client.Client, provider schedulerprovider.ProviderType, lws *leaderworkerset.LeaderWorkerSet, expectedCount int) {
+	gvk := getPodGroupGVK(provider)
+	if gvk.Empty() {
+		ginkgo.Skip("Unsupported scheduler provider for PodGroup validation")
```
kerthcet: I think we should return error here? Empty gvk is not allowed I think.

test/testutils/util.go
```diff
@@ -696,3 +738,194 @@ func SetLwsPartition(ctx context.Context, k8sClient client.Client, lws *leaderwo
...
+		// Validate podgroup owner reference and spec
+		validateOwnerReference(podGroup, leaderPod)
+		validatePodGroupSpec(podGroup, lws)
+	}
+}
+
+// UpdatePodGroupAtIndex manually updates the PodGroup for a specific group index
+func UpdatePodGroupAtIndex(ctx context.Context, k8sClient client.Client, providerType schedulerprovider.ProviderType, schedulerProvider schedulerprovider.SchedulerProvider, lws *leaderworkerset.LeaderWorkerSet, groupIndex string) {
+	gvk := getPodGroupGVK(providerType)
+	if gvk.Empty() {
```
kerthcet, line 870: ditto.

test/testutils/util.go
```diff
@@ -696,3 +738,194 @@ func SetLwsPartition(ctx context.Context, k8sClient client.Client, lws *leaderwo
...
+		err = schedulerProvider.InjectPodGroupMetadata(pod)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	}
+	gomega.Expect(CreateLeaderPodsWithInjectFn(ctx, leaderSts, k8sClient, lws, startIndex, startIndex+1, injectFn)).To(gomega.Succeed())
+}
+
+// ExpectValidPodGroupAtIndex checks that a PodGroup exists for the specified group index and has the correct owner reference
+func ExpectValidPodGroupAtIndex(ctx context.Context, k8sClient client.Client, provider schedulerprovider.ProviderType, lws *leaderworkerset.LeaderWorkerSet, groupIndex string) {
+	gvk := getPodGroupGVK(provider)
+	if gvk.Empty() {
```
kerthcet, line 910: ditto

pkg/utils/utils.go
```diff
@@ -74,3 +79,25 @@ func GetOperatorNamespace() string {
 	}
 	return defaultNamespace
 }
+
+// CalculatePGMinResources calculates the minimum resources needed for an entire PodGroup [1 Leader + (size-1) Worker pods]
```
kerthcet, line 83: we need to have a unit test for this later, it could be a follow up PR.

test/e2e/e2e_test.go
JesseStutler: > Since we changed the size to 4, shouldn't we have 4 podGroups?

Yes size is updated to 4, more worker pods will come out. But the replicas of lws is still 2, the number of podgroups is same as replicas. There will be two podgroups for (2 leader pods and 6 worker pods)

kerthcet: got it.

test/testutils/util.go
JesseStutler: OK I have changed to `ginkgo.Fail`, please check

test/testutils/util.go
JesseStutler: OK I have changed to `ginkgo.Fail`, please check

test/testutils/util.go
JesseStutler: OK I have changed to `ginkgo.Fail`, please check

pkg/utils/utils.go
JesseStutler: OK, I'll do it in later PR, thanks for your reminder

kerthcet:
/lgtm
/approve

Thanks @JesseStutler

ardaguclu:
@JesseStutler I was expecting changes under `/config` directory which aligns with the changes under `/charts`. Is it missing or intentional?. For example, where is clusterrole changes for volcano under config directory?

ardaguclu:
Currently, I don't think gang scheduler is supported via kustomize, unless user manually applies required configurations.

JesseStutler:
> Currently, I don't think gang scheduler is supported via kustomize, unless user manually applies required configurations.

Thanks for your reminder @ardaguclu I was also thinking about this question. I'm not very familiar with kustomize. Are the configurations under `/config` enabled by default? Since gang scheduling is turned off by default, I didn't add any volcano-related configurations under `/config`. I don't know if kustomize can work like Helm, allowing you to specify which to enable and which to disable?

ardaguclu:
For example, if the only requirement is to define a ClusterRole, you need to add it in here https://github.com/kubernetes-sigs/lws/tree/main/config/rbac. After that you need to add this file https://github.com/kubernetes-sigs/lws/blob/b73c72c99cca0ff70e73b0ab7b50f51d9b98cff2/config/rbac/kustomization.yaml#L1 in commented format by mentioning "if you want to enable gang scheduling, you need to comment out this section, etc."

JesseStutler:
> For example, if the only requirement is to define a ClusterRole, you need to add it in here https://github.com/kubernetes-sigs/lws/tree/main/config/rbac. After that you need to add this file
> 
> https://github.com/kubernetes-sigs/lws/blob/b73c72c99cca0ff70e73b0ab7b50f51d9b98cff2/config/rbac/kustomization.yaml#L1
> 
> in commented format by mentioning "if you want to enable gang scheduling, you need to comment out this section, etc."

Good idea, thanks! @ardaguclu I will add it in later PRs

Issue #514, opened 2025-04-28
Add LWS configuration in Helm Chart


caozhuozi:
**What happened**:

internal cert manager option is always true if is not set.

**What you expected to happen**:

As a boolean, by default it should be false unless user explicitly set it to true.

**How to reproduce it (as minimally and precisely as possible)**:

1. start the manager without `--config` flags.
    the config obj will be: (`internalCertManagement:\n  enable: true`)
```
2025-04-28T03:57:01Z	INFO	setup	Successfully loaded configuration	{"config": "apiVersion: config.lws.x-k8s.io/v1alpha1\nclientConnection:\n  burst: 500\n  qps: 500\nhealth:\n  healthProbeBindAddress: :8081\n  livenessEndpointName: /healthz\n  readinessEndpointName: /readyz\ninternalCertManagement:\n  enable: true\n  webhookSecretName: lws-webhook-server-cert\n  webhookServiceName: lws-webhook-service\nkind: Configuration\nleaderElection:\n  leaderElect: true\n  leaseDuration: 15s\n  renewDeadline: 10s\n  resourceLock: leases\n  resourceName: b8b2488c.x-k8s.io\n  resourceNamespace: \"\"\n  retryPeriod: 2s\nmetrics:\n  bindAddress: :8443\nwebhook:\n  certDir: /tmp/k8s-webhook-server/serving-certs\n  port: 9443\n"}
```
2. cert-roration will be automatically started.
```
2025-04-28T03:57:02Z INFO    cert-rotation    starting cert rotator controller
```
followed by error logs of the cert-controller:
```
2025-04-28T03:57:23Z	ERROR	cert-rotation	could not refresh CA and server certs	{"error": "Operation cannot be fulfilled on secrets \"lws-webhook-server-cert\": the object has been modified; please apply your changes to the latest version and try again"}
github.com/open-policy-agent/cert-controller/pkg/rotator.(*CertRotator).refreshCertIfNeeded.func1
	/go/pkg/mod/github.com/open-policy-agent/cert-controller@v0.12.0/pkg/rotator/rotator.go:329
k8s.io/apimachinery/pkg/util/wait.runConditionWithCrashProtection
	/go/pkg/mod/k8s.io/apimachinery@v0.32.3/pkg/util/wait/wait.go:145
k8s.io/apimachinery/pkg/util/wait.ExponentialBackoff
[...]

PR #517, opened 2025-04-28
Add LWS configuration in Helm Charts

ardaguclu:
#### What type of PR is this?
/kind bug

Helm chart does not support configuring LWS configuration like Kustomize. This PR defines this configuration as configmap and binds it to deployment as a volume to be used by the application. 

Fixes https://github.com/kubernetes-sigs/lws/issues/514

```release-note
None
```

### DIFF
--- a/charts/lws/templates/manager/configmap.yaml
+++ b/charts/lws/templates/manager/configmap.yaml
@@ -1,0 +1,16 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "lws.fullname" . }}-manager-config
+  namespace: {{ .Release.Namespace}}
+  labels:
+    {{- include "lws.labels" . | nindent 4 }}
+data:
+  controller_manager_config.yaml: |2
+
+    apiVersion: config.jobset.x-k8s.io/v1alpha1
+    kind: Configuration
+    leaderElection:
+      leaderElect: true
+    internalCertManagement:
+      enable: {{ not .Values.certManager.enable }}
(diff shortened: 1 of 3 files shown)


caozhuozi:
Do we need to also add corresponding values for each option item in the config?

ardaguclu:
> Do we need to also introduce corresponding values for each option item in the config?

In my opinion, other configurations are not needed in most of the time and that would be redundant to explicitly specify each of them.

charts/lws/templates/manager/configmap.yaml
```diff
@@ -0,0 +1,32 @@
+{{- /*
```
kerthcet, line 1: Seems we don't require for this, let's keep consistent.

kerthcet (commented):
/approve

charts/lws/templates/manager/configmap.yaml
ardaguclu: Thank you for review. Removed.

andrew0li:
Why does this use the JobSet config resource? I'm getting this in the LWS manager logs

```
unable to load the configuration        {"error": "no kind \"Configuration\" is registered for version \"config.jobset.x-k8s.io/v1alpha1\" in scheme \"pkg/runtime/scheme.go:100\""}
```

ardaguclu:
> "config.jobset.x-k8s.io

@andrew0li thank you for reporting this. This is not correct. I've opened https://github.com/kubernetes-sigs/lws/pull/527

PR #523, opened 2025-05-06
fix: support foreground deletion

congcongke:
#### What type of PR is this?

/kind bug

Fixes

The deletion will fall into infinite loop with command `kubectl delete lws --cascade='foreground'`.

Simple e2e test appended.

```release-note

```

### DIFF
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -100,6 +100,11 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	if err := r.Get(ctx, types.NamespacedName{Name: req.Name, Namespace: req.Namespace}, lws); err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+
+	if lws.DeletionTimestamp != nil {
+		return ctrl.Result{}, nil
+	}
+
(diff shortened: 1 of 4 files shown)


pkg/controllers/leaderworkerset_controller.go
```diff
@@ -109,6 +109,11 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, err
 	}
 
+	// lws is in deleting and leader sts is also in deleting, return directly.
+	if lws.DeletionTimestamp != nil && (leaderSts == nil || leaderSts.DeletionTimestamp != nil) {
+		return ctrl.Result{}, nil
+	}
+
```
yankay, line 116: Hi @congcongke,

I'm not entirely sure, but could you please clarify why we need to check `(leaderSts == nil || leaderSts.DeletionTimestamp != nil)`? Wouldn't it suffice to just check `lws.DeletionTimestamp != nil` to avoid reconciling LWS that are marked for deletion?

Additionally, do you think it would make more sense to place this piece of code before `leaderSts, err := r.getLeaderStatefulSet(ctx, lws)`?

congcongke: yes, you are right!

kerthcet:
Thanks @congcongke

PR #534, opened 2025-05-19
fix: update the subgroup-index caculation

### LINKED ISSUE #533: Polish the subgroup-index formula

About ` KEP-115 Subgroup Support` and its implementation:

The `getSubGroupIndex()`( `subgroup-index = (workerIndex - 1) / subGroupSize `)  may encounter problems.

(as my understanding , `subGroupSize = N` means we want to divide the pods (belong to one LWS instances) into N sub-groups)

for example
```
size=3( 1 leader 2 workers), subGroupSize=2
```

```
leader (no TPU):  subgroup-index = (0-1)/2 = 0
worker-1            :  subgroup-index = (1-1)/2 = 0
worker-2            :  subgroup-index = (2-1)/2 = 0
```
But the expected behavior is :
- worker-1 belongs to subgroup=0, 
- worker-2 belongs to subgroup=1 !!!

another example:

```
size=7( 1 leader 6 workers)，subGroupSize=3
```

```
- Leader    :    (0-1)/3 =0
- Worker 1：(1-1)/3 = 0
- Worker 2：(2-1)/3 = 0
- Worker 3：(3-1)/3 = 0
- Worker 4：(4-1)/3 = 1
- Worker 5：(5-1)/3 = 1
- Worker 6：(6-1)/3 = 1
```
We can see all pods are either group 0 or group 1, 

but we expect there will be group 3 ( because `subGroupSize=3`) but there is none.

So, I proposed to **change** the formula to **`subgroup-index = workerIndex % subGroupSize`**

**Why is this needed**:

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

panpan0000:
#### What type of PR is this?
kind bug

refer to #533  for explanation 

There's flaw for original subgroup calculation .
So make the calculation more accurate. 

Fixes #533

```release-note

```

### DIFF
--- a/pkg/webhooks/pod_webhook.go
+++ b/pkg/webhooks/pod_webhook.go
@@ -149,7 +149,7 @@ func (p *PodWebhook) Default(ctx context.Context, obj runtime.Object) error {
 				return err
 			}
 			leaderName := pod.Annotations[leaderworkerset.LeaderPodNameAnnotationKey]
-			subGroupIndexKey := getSubGroupIndex(podCount, subGroupSizeInt, workerIndex)
+			subGroupIndexKey := getSubGroupIndex(subGroupSizeInt, workerIndex)
 			pod.Labels[leaderworkerset.SubGroupIndexLabelKey] = subGroupIndexKey
 			subGroupUniqueKey := genGroupUniqueKey(leaderName, subGroupIndexKey)
 			pod.Labels[leaderworkerset.SubGroupUniqueHashLabelKey] = subGroupUniqueKey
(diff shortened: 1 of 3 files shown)


pkg/webhooks/pod_webhook.go
```diff
@@ -242,10 +242,6 @@ func exclusiveAffinityApplied(pod corev1.Pod, topologyKey string) bool {
 	return hasAffinity && hasAntiAffinity
 }
 
-func getSubGroupIndex(podCount int, subGroupSize int, workerIndex int) string {
-	if (podCount-1)%subGroupSize == 0 {
```
kerthcet, line 246: It's a new feature called LeaderExcluded just incase the leader is a proxy, should not be part of the subgroup, see.
https://github.com/kubernetes-sigs/lws/blob/a2ae101b56d498c193074ee0b45b07439d62bf6b/api/leaderworkerset/v1/leaderworkerset_types.go#L224

kerthcet: Wait, let me revisit the design.

kerthcet: https://github.com/kubernetes-sigs/lws/issues/533#issuecomment-2899802572

panpan0000: thank you !

PR #538, opened 2025-05-21
fix: close leader elect

ls-2018:
close


kerthcet:
Please fix the tests @ls-2018

PR #2, opened 2024-02-28
Small change to README

liurupeng:

PR #540, opened 2025-05-23
fix: exclusive topology only affect inside LWS instance

### LINKED ISSUE #539: exclusive-topology  antiAffinity should include matching LWS name
**What happened**:

should add below 
```
          - key: leaderworkerset.sigs.k8s.io/group-key
            operator: In
            values:
            - $MY_LWS_NAME
```

the whole worker affinity looks like 

```
  affinity:
    podAntiAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchExpressions:
          - key: leaderworkerset.sigs.k8s.io/group-key
            operator: In
            values:
            - $MY_LWS_NAME
          - key: leaderworkerset.sigs.k8s.io/group-key
            operator: Exists
          - key: leaderworkerset.sigs.k8s.io/group-key
            operator: NotIn
            values:
            - 8fb.....23d207e113
        topologyKey: topology.kubernetes.io/zone
```

in the same namespace:

1. run a LWS(named "A" ) with `leaderworkerset.sigs.k8s.io/exclusive-topology` annotation.
2. run another LWS(named "B" ) with `leaderworkerset.sigs.k8s.io/exclusive-topology` annotation.
you will see B will be pending due to A already live in those nodes (but B may use diff resource than A, they should not compete and A should not stop B from running )

**Environment**:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

panpan0000:
#### What type of PR is this?
kind bug

say, we have two LWS instance on the same nodes,
LWS-1 requires CPU
LWS-2 requires GPU
they should share the same topology , and the exclusive policy should apply inside LWS-1 and LWS-2 respectively.

But now, if LWS-1 occupied the nodes, the LWS-2 will be pending   

Fixes #539

```release-note

```


Edwinhr716:
/hold
Does this work if LWS "A" and LWS "B" have different names? We inject the Pod AntiAffinity of not having the GroupKey label, where `GroupKey` depends on the pod name, and the pod name changes based on LWS name.

Edwinhr716:
> = (Same Namespace) AND (Same LWS Set Name ) [ OR (Diff GroupID)]

How are you setting the OR here? If a pod has multiple pod anti-affinities they will be treated as AND no?

panpan0000:
> > = (Same Namespace) AND (Same LWS Set Name ) [ OR (Diff GroupID)]
> 
> How are you setting the OR here? If a pod has multiple pod anti-affinities they will be treated as AND no?

Oops, you found a typo @Edwinhr716 
image


I tried to keep the original code `GroupIDKey` (`LabelSelectorOpNotIn` with `SubGroupIndexLabelKey`) to make minimal changes:

so my PR just adds  a `LabelSelectorOpIn` with 'lwsName', 
to achieve the final effect of expected  **`= (Same Namespace) AND (Same LWS Set name) AND (Different GroupID)`**

Let me explain the The PR result again (below are mathematical logic operation calculation):
```
Anti Affinity 
= (Same LWS Set name) AND    (Different GroupID-Key)         <--- current PR code
```
// because  "GroupID-Key = NS + LWSName + GroupID", so Derivation begins below:
```
= (Same LWS Set name) AND    [(Diff namespace) OR (Diff LWS Set name) OR (Diff GroupID) ] 
```

make it shorter :
```
A AND (B OR !A OR C) 
= (A AND B) OR (A AND !A) OR (A AND C)
= (A AND B) OR   False    OR (A AND C)
= (A AND B) OR (A AND C)
```

So the result is 
```
= [ (Same LWS Set name) AND (Diff namespace) ] OR   [ (Same LWS Set name) AND (Diff GroupID) ]
```



Aha... with the `GroupIDKey` ,  everything become quite **obscure** T_T



That's why I asked your experts opinion:  if we can remove `GroupIDKey`,  but just plainly use LWSName & GroupIndex as condotion.

> But I think changing it to below will be more clear , What do. you think ?
>
>				{
>					Key:      leaderworkerset.GroupIndexLabelKey,
>					Operator: metav1.LabelSelectorOpNotIn,
>					pod.Labels[leaderworkerset.SubGroupIndexLabelKey],
>				},

panpan0000:
> ```
> = (A AND B) OR (A AND C)
> ```
> 
> So the result is
> 
> ```
> = [ (Same LWS Set name) AND (Diff namespace) ] OR   [ (Same LWS Set name) AND (Diff GroupID) ]
> ```
> 
> Aha... with the `GroupIDKey` , everything become quite **obscure** T_T
> 


## In short
 the code logic **can work**,  but very very **obscure** .


## When if we abandon the GroupIDKey

it will be easier to understand like below, what do you think if we abandon GroupKey and change the code like this ? If yes, I could rework this PR?

```
    podAntiAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchExpressions:
          - key: leaderworkerset.sigs.k8s.io/name
            operator: In
            values:
            - $LWS_NAME
          - key: leaderworkerset.sigs.k8s.io/group-index
            operator: Exists
          - key: leaderworkerset.sigs.k8s.io/group-index       <-------- change from GroupKEY to Group ID
            operator: NotIn
            values:
            - $MY_GROUP_ID
        namespaces:                 <-------- add namespace restriction 
        - $MY_NAMESPACE
        topologyKey: topology.kubernetes.io/$MY_TOPOLOGY_KEY
```

If you agree, I've another PR for this fix method no.2 :
https://github.com/kubernetes-sigs/lws/pull/551

Edwinhr716:
Thanks for the thorough explanation, now it makes sense to me what happens if they have diff group ID.

One last question, what happens if LWS A and LWS B both request the same type of resource? Then exclusive placement won't be respected in that case correct?

panpan0000:
> Thanks for the thorough explanation, now it makes sense to me what happens if they have diff group ID.
> 
> One last question, what happens if LWS A and LWS B both request the same type of resource? Then exclusive placement won't be respected in that case correct?

Good point , @Edwinhr716 . 

let's assume LWS A & B both request GPU:
Denote 4 nodes with 8 GPU each (total 32 GPUs):

Case #1 ( no conflict  )
- LWS A : size 4, each leader & worker requests **4** GPU.        LWS B is the same.
- Then LWS A and B will be co-located in the 4 nodes, A uses 50% GPU and B uses other 50%. 


Case #2:  ( with conflict  )
- LWS A : size 4, each leader & worker requests **8** GPU.        LWS B is the same.
- If A comes first, A will occupied all 32 GPUs, B will not be scheduled. If some pod in A restarts , B's pod may preemp (it requires gang-scheduling, maybe now with help of Kueue?)
- If A & B comes at the same time, it requires gang-scheduling( already in LWS roadmap) , otherwise, some pods in A or B will be pending and cause dead-lock



So, Do you think it's acceptable ?

Edwinhr716:
So in the case of 2, we are no longer guaranteeing exclusive placement, even if the flag is set.

panpan0000:
> So in the case of 2, we are no longer guaranteeing exclusive placement, even if the flag is set.

hmm...yes.  Do you think is there any way to statify both case 1 and 2？ or case2 should wait gang being implemented？

Edwinhr716:
I don't think we should break existing functionality for the purpose of adding new one. @kerthcet @ahg-g @ardaguclu @yankay thoughts?

panpan0000:
@Edwinhr716  Or we add another  policy field as a flag ?

ahg-g:
We can do that with a parameter, perhaps call it `exclusive-topology-scope` that takes values like 
- "Global": means across all LWS instances
- "WithinSameLWS" means within an LWS instance

Edwinhr716:
Adding a new field sounds good to me, though we should add a disclaimer that using the feature can cause a deadlock

kerthcet:
I was on holiday before, quote my comments here https://github.com/kubernetes-sigs/lws/issues/539#issuecomment-2933976121:
```
However, I think `You will see B will be pending due to A already live in those nodes (but B may use diff resource than A, they should not compete and A should not stop B from running )` 
exactly reflects the meaning of exclusive. And I don't think we should change the semantic.
```

Not objecting the idea but I would defer this until we see other requests.

> though we should add a disclaimer that using the feature can cause a deadlock

What do you mean deadlock, if so, I would say we should avoid it.

Edwinhr716:
> What do you mean deadlock, if so, I would say we should avoid it.

See
> Case https://github.com/kubernetes-sigs/lws/pull/2: ( with conflict )
LWS A : size 4, each leader & worker requests 8 GPU. LWS B is the same.
If A comes first, A will occupied all 32 GPUs, B will not be scheduled. If some pod in A restarts , B's pod may preemp (it requires gang-scheduling, maybe now with help of Kueue?)
If A & B comes at the same time, it requires gang-scheduling( already in LWS roadmap) , otherwise, some pods in A or B will be pending and cause dead-lock

kerthcet:
Got it, we should defer this until gang supports at least.

panpan0000:
> #539 (comment)

copy my reply from #539 to here :

 I would say the existing implementation combines topology-aware with exclusive together and kind of mixing up.

But actually, it would be better that topology-aware as a major function, and exclusive as a flag of sub-function, like

```
spec:
 ...
 topology:
       nodeSelectorKey: "xxx"          <-- it will add `nodeSelector:           topology.kubernetes.io/supernode: xxx` for pods, and affinity rule for same groups
       exclusive:  true                <-- it will add antiAffinity to other groups
```

So what's your suggestion ? @Edwinhr716 @ahg-g  @kerthcet ?  

(1) should we split the topology and exclusive semantic apart , like above yaml ?


or (2) adding a flag `exclusive-topology-scope`  like @ahg-g said ?

>"Global": means across all LWS instances
>"WithinSameLWS" means within an LWS instance

Issue #511, opened 2025-04-27
Support lws partitioned for update


lixin010:
**What would you like to be added**:
I'd like to add the `Partition` field to `RollingUpdateConfiguration` struct like sts.
```go
// leaderworkerset_types.go
type RollingUpdateConfiguration struct {
        Partition *int32

	MaxUnavailable intstr.IntOrString `json:"maxUnavailable,omitempty"`

	MaxSurge intstr.IntOrString `json:"maxSurge,omitempty"`
}
```
Meanwhile, the `rollingUpdateParameters` function in lws controller also needs to be restructured to add the partition update feature.

**Why is this needed**:
It enable lws to have the ability to perform partitioned updates. This is helpful in being able to do a canary based deployment.

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [x] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

Edwinhr716:
This feature makes sense to me, also brings features that StatefulSet already has to the LWS layer. @kerthcet, @ahg-g thoughts?

kerthcet:
+1 to me, @lixin010 would you like to draft a KEP then?

PR #541, opened 2025-05-26
add kep-511 partition update

lixin010:
#### What type of PR is this?

/kind documentation

add KEP for #511 

Fixes #

```release-note

```

### DIFF
--- a/keps/511-partition-update/README.md
+++ b/keps/511-partition-update/README.md
@@ -167,15 +167,6 @@ type RollingUpdateConfiguration struct {
 }
 ```
 
-In partition updates, there will be two versioned replicas. The `ReadyReplicas` and `UpdatedReplicas` filed are insufficient to represent the current status of the lws. So adding the `UpdatedReadyReplicas` field can better indicate the state of the new version replicas.
-```go
-// leaderworkerset_types.go
-type LeaderWorkerSetStatus struct {
-    // UpdatedReadyReplicas track the number of groups that have been updated and are in ready state.
-    UpdatedReadyReplicas int32 `json:"updatedReadyReplicas,omitempty"`
-}
-```
-
 ### Implementation
 #### lws controller
 In the `rollingUpdatePartition`function, an intermediate variable "partition" is calculated, which means the groups from `partition`to `replicas-1` will be updated. The function returns `min(partition, currentPartition)` that means Partition moves in one direction to make it simple. 


lixin010:
@Edwinhr716 @kerthcet @ardaguclu please have a look when you have time. thks.

lixin010:
@Edwinhr716  Thaks. I fixed the toc error.

keps/511-partition-update/README.md
```diff
@@ -0,0 +1,318 @@
...
+    // new code, to implement lws partition update
+    lwsPartition = getLwsPartition(lws)
+    partition = max(partition, lwsPartition)
+    
+    // That means Partition moves in one direction to make it simple.
+    return min(partition, currentPartition)
+}
+```
+
+During a rolling update, the `partition` field updated to the leader sts will not be smaller than `lwsPartition`, which keeps the groups smaller than `lwsPartition`will not be updated.
```
Edwinhr716: We also need to update the logic that determines when a rolling update is done 
https://github.com/kubernetes-sigs/lws/blob/7b8a941ed516a598d3bc6757651cbc6ca22ea3d6/pkg/controllers/leaderworkerset_controller.go#L446
Otherwise it will get stuck as the updateAndReadyCount will never equal the number of replicas

keps/511-partition-update/README.md
```diff
@@ -0,0 +1,318 @@
...
+demonstrate the interest in a KEP within the wider Kubernetes community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+This KEP is to enable lws to have the ability to perform partitioned updates, facilitating canary or gray release. In disaggregated prefilling cases, it can be utilized to align the version ratio between p and d.
+
+### Goals
+
+- Add partitionfield to ebable partitioned updates for LWS, similar to the [StatefuleSet partition](https://kubernetes.io/zh-cn/docs/concepts/workloads/controllers/statefulset/#partitions).
```
kerthcet: Please refer to an english version.

keps/511-partition-update/README.md
```diff
@@ -0,0 +1,318 @@
...
+### User Stories (Optional)
+
+<!--
+Detail the things that people will be able to do if this KEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
```
kerthcet, line 117: Could you also add the story about canary deployment?

keps/511-partition-update/README.md
```diff
@@ -0,0 +1,318 @@
...
+    
+    // That means Partition moves in one direction to make it simple.
+    return min(partition, currentPartition)
+}
+```
+
+During a rolling update, the `partition` field updated to the leader sts will not be smaller than `lwsPartition`, which keeps the groups smaller than `lwsPartition`will not be updated.
+
+#### lws webhook
+We need to set default value of `Spec.RolloutStrategy.RollingUpdateConfiguration.Partition` to `0` in the lws webhook.
```
kerthcet: The kubebuilder tag can help us, no need to set the webhook.

keps/511-partition-update/README.md
```diff
@@ -0,0 +1,318 @@
...
+when drafting this test plan.
+
+[testing-guidelines]: https://git.k8s.io/community/contributors/devel/sig-testing/testing.md
+-->
+
+[x] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+##### Prerequisite testing updates
```
kerthcet, line 227: We need to verify the replicas with `index < partition`, during the rolling update, it always populates with the old version.

keps/511-partition-update/kep.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - kerthcet
+  - Edwinhr716
+approvers:
+  - kerthcet
+  - Edwinhr716
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v0.6.1"
```
kerthcet: v0.7.0 I believe, it refers to the upcoming milestone.

keps/511-partition-update/kep.yaml
```diff
@@ -0,0 +1,17 @@
+title: Partition Update
+kep-number: 511
+authors:
+  - lixin010
+status: provisional
```
kerthcet: implementable

keps/511-partition-update/README.md
```diff
@@ -0,0 +1,318 @@
...
+    // ...
+    // ...
+    partition = ...
+
+    // new code, to implement lws partition update
+    lwsPartition = getLwsPartition(lws)
+    partition = max(partition, lwsPartition)
+    
+    // That means Partition moves in one direction to make it simple.
+    return min(partition, currentPartition)
```
kerthcet: Shouldn't we always set `partition = max(lwsPartition, calculatedPartition)`.

keps/511-partition-update/README.md
```diff
@@ -0,0 +1,318 @@
...
+https://testgrid.k8s.io/sig-testing-canaries#ci-kubernetes-coverage-unit
+
+This can inform certain test coverage improvements that we want to do before
+extending the production code to implement this enhancement.
+-->
+
+<!-- - `<package>`: `<date>` - `<test coverage>` -->
+- Test whether the `partition` value is correct.
+
+##### Integration tests
```
kerthcet, line 258: Also let's have integration tests to make sure the partitioned rolling update is working as expected with maxUnavailable step by step.

keps/511-partition-update/README.md
```diff
@@ -0,0 +1,318 @@
...
+<!--
+Major milestones in the lifecycle of a KEP should be tracked in this section.
+Major milestones might include:
+- the `Summary` and `Motivation` sections being merged, signaling SIG acceptance
+- the `Proposal` section being merged, signaling agreement on a proposed design
+- the date implementation started
+- the first Kubernetes release where an initial version of the KEP was available
+- the version of Kubernetes where the KEP graduated to general availability
+- when the KEP was retired or superseded
+-->
```
kerthcet, line 321: Add the history here, like 
- 2025-06-01: KEP initialized and submitted for review

kerthcet (commented):
I think most of the part is good to me since the feature is really straightforward.

keps/511-partition-update/README.md
lixin010: Yes, we need to update these.

keps/511-partition-update/README.md
lixin010: This test point will be verified in the e2e test case I mentioned.

keps/511-partition-update/README.md
lixin010: It makes sense.

lixin010:
Thanks for @Edwinhr716 @kerthcet 's review. I have revised kep according to the suggestions.

keps/511-partition-update/README.md
```diff
@@ -0,0 +1,329 @@
...
+    // +optional
+    Partition *int32
+    
+    MaxUnavailable intstr.IntOrString `json:"maxUnavailable,omitempty"`
+    
+    MaxSurge intstr.IntOrString `json:"maxSurge,omitempty"`
+}
+```
+
+In partition updates, there will be two versions replicas. the `ReadyReplicas` and `UpdatedReplicas` filed are insufficient to represent the current status of the lws. So adding the `UpdatedReadyReplicas` field can better indicate the state of the new version replicas.
```
kerthcet:
```suggestion
In partition updates, there will be two versioned replicas. The `ReadyReplicas` and `UpdatedReplicas` filed are insufficient to represent the current status of the lws. So adding the `UpdatedReadyReplicas` field can better indicate the state of the new version replicas.
```

I would suggest to follow the statefulset design, would `currentReplicas` good enough for you?

lixin010: I think `currentReplicas` cannot cover the semantics of `updatedReadyReplicas`. I hope to adopt this field, which contains the version infomation of ready replicas. What do you think?

Edwinhr716:
Another case we should consider: We have an LWS object with X replicas on version A, and Y replicas on version B, where A is the newest version, and PodRecreateOnGroupRestart is triggered. The leader recreation is handled by the sts controller so it should work fine. However, when creating the workers to match the leader, we fetch the revision that matches the leader https://github.com/kubernetes-sigs/lws/blob/b1cb09073200f1c23f0ac8a3b213e854966517fa/pkg/controllers/pod_controller.go#L137 

But, since the upgrade is done, only the Revision of A will exist because we delete all revisions aside from the newest one https://github.com/kubernetes-sigs/lws/blob/b1cb09073200f1c23f0ac8a3b213e854966517fa/pkg/controllers/leaderworkerset_controller.go#L178-L179

So when the Leader Pod with version B is recreated, and the pod controller is triggered to create the worker pod, when we fetch the revision of B, the function returns nil, causing a crashing loop.

lixin010:
> Another case we should consider: We have an LWS object with X replicas on version A, and Y replicas on version B, where A is the newest version, and PodRecreateOnGroupRestart is triggered. The leader recreation is handled by the sts controller so it should work fine. However, when creating the workers to match the leader, we fetch the revision that matches the leader
> 
> https://github.com/kubernetes-sigs/lws/blob/b1cb09073200f1c23f0ac8a3b213e854966517fa/pkg/controllers/pod_controller.go#L137
> 
> But, since the upgrade is done, only the Revision of A will exist because we delete all revisions aside from the newest one
> 
> https://github.com/kubernetes-sigs/lws/blob/b1cb09073200f1c23f0ac8a3b213e854966517fa/pkg/controllers/leaderworkerset_controller.go#L178-L179
> 
> So when the Leader Pod with version B is recreated, and the pod controller is triggered to create the worker pod, when we fetch the revision of B, the function returns nil, causing a crashing loop.



There is indeed a problem here. One solution is to clean up all other controller revisions except the current vision when all replicas have been updated. So we need change `updateConditions` func to return `allUpdateDone` which means all replicas have been updated and completed, not `updateDone`.  @Edwinhr716 WDYT?

Edwinhr716:
Yes that makes sense to me. Doing that lets us differentiate between `LeaderWorkerSetAvailable` and when all replicas are using the same version.

lixin010:
Thanks to @Edwinhr716 and @kerthcet for the reivew.  I have updated the kep. If there are no other problems, could this PR be merged? I will write code to implement the feature.

Edwinhr716:
From my side it is ready to merge, but will wait for https://github.com/kubernetes-sigs/lws/pull/541#discussion_r2128038841 to be addressed to merge.

Feel free to create the implementation PR, we can start the review of the PR in parallel since the KEP review is mostly done

kerthcet:
Sorry I was on vacation, will take a look today.

kerthcet:
> There is indeed a problem here. One solution is to clean up all other controller revisions except the current vision when all > replicas have been updated. So we need change updateConditions func to return allUpdateDone which means all replicas > have been updated and completed, not updateDone. @Edwinhr716 WDYT?

We can introduce PartitionedUpdateDone to distinguish with the updateDone. Would that be simple to delete all the revision versions only we reach the updateDone, which means keep the current logic unchanged. I guess the versions wouldn't be too much.

keps/511-partition-update/README.md
kerthcet: I didn't see any reference to this field in the proposal, so could you help me understand why it's not enough here? Basically, we're sts on sts, so I think it should be enough for us. Maybe I'm wrong.

kerthcet:
BTW, do you think we should at some point move the revision to the lws status just like statefulset? @Edwinhr716

lixin010:
> > There is indeed a problem here. One solution is to clean up all other controller revisions except the current vision when all > replicas have been updated. So we need change updateConditions func to return allUpdateDone which means all replicas > have been updated and completed, not updateDone. @Edwinhr716 WDYT?
> 
> We can introduce PartitionedUpdateDone to distinguish with the updateDone. Would that be simple to delete all the revision versions only we reach the updateDone, which means keep the current logic unchanged. I guess the versions wouldn't be too much.

@kerthcet  Yes, the solution I mentioned is to delete all the revision versions only we reach the updateDone, which is exactly the same logic as what you said.  It is because after introducing the partition updated feature, the `updateDone` returned by the `updateConditions` function does not indicate `allUpdatedDone`, but `PartitionedUpdateDone`. So we need change this.

keps/511-partition-update/README.md
lixin010: The `currentReplicas` is the number of groups which are current revision. Consider such a scenario where `Spec.Replicas` = 4 and `partition` = 2. So the `Status.CurrentReplicas` = 2 and `Status.UpdatedReplicas` = 2. If the `Status.ReadyReplicas` = 2, I don't know which version of groups is ready. We may want to know whether the updated group is ready.

kerthcet: I don't think we should inference the status from these fields, instead they just reflect the status of rolling update. Can we just move this section and let's discuss the implementation details in the PR instead.

For currentReplicas, it could be another PR not related to KEP. WDYT?

lixin010: ok, I will remove the content about the `updatedReadyReplicas` field from the kep.

kerthcet:
/lgtm
/approve

Thanks @lixin010 Let's move to the implementation then.

Edwinhr716:
> BTW, do you think we should at some point move the revision to the lws status just like statefulset? @Edwinhr716

If a feature needs it I'm not against adding it to the status. But I don't see a need for us to add it to the status at its current state. When implementing controller revision we opted to keep it as simple as possible, which is why we didn't add it to the status.

PR #557, opened 2025-06-09
KEP #552: allow updates to `spec.leaderWorkerTemplate.size` field

### LINKED ISSUE #552: Allow to update `spec.leaderWorkerTemplate.size`

I'd like to be able to update the `spec.leaderWorkerTemplate.size` field instead of recreating the `LWS` instance.

**Why is this needed**:

This would allow to rollout new topologies without having to handle two `LWS` instances.

**Completion requirements**:

- Updating the `spec.leaderWorkerTemplate.size` is allowed.

- [ ] Design doc
- [ ] API change
- [ ] Docs update

synthe102:
#### What type of PR is this?

/kind feature
/kind api-change

New KEP to allow updates to the `spec.leaderWorkerTemplate.size` field.

Fixes #552 

```release-note

```

### DIFF
--- a/keps/552-worker-resizing/README.md
+++ b/keps/552-worker-resizing/README.md
@@ -1,0 +1,319 @@
+# KEP-552: allow worker size updates
+
+<!--
+This is the title of your KEP. Keep it short, simple, and descriptive. A good
+title can help communicate what the KEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a KEP and for
+highlighting any additional information provided beyond the standard KEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+<!--
+This section is incredibly important for producing high-quality, user-focused
+documentation such as release notes or a development roadmap. It should be
+possible to collect this information before implementation begins, in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself. KEP editors and SIG Docs
+should help to ensure that the tone and content of the `Summary` section is
+useful for a wide audience.
+
+A good summary is probably at least a paragraph in length.
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+This KEP aims to allow resizing of workers by introducing a new field that defines the type of resize strategy. The current strategy will become `None` to keep the same behavior. `Recreate` will be added. The `spec.leaderWorkerTemplate.size` field will become editable by relaxing the validating webhook.
+
+## Motivation
+
+<!--
+This section is for explicitly listing the motivation, goals, and non-goals of
+this KEP.  Describe why the change is important and the benefits to users. The
+motivation section can optionally provide links to [experience reports] to
+demonstrate the interest in a KEP within the wider Kubernetes community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+It is not currently possible to update the `spec.leaderWorkerTemplate.size` field. This complicates updates to existing `LeaderWorkerSets` by forcing to recreate resources, as well a being a blocker for more innovative topologies where in-place resize of worker pools is supported beyond `spec.replicas`.
+
+### Goals
+
+<!--
+List the specific goals of the KEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+The current goals are:
+- Allow to update the `spec.leaderWorkerTemplate.size` field.
+- Add an optional field to control the way replicas are resized.
+
+
+### Non-Goals
+
+<!--
+What is out of scope for this KEP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+- This KEP doesn't propose a way to target `spec.leaderWorkerTemplate.size` with an autoscaling mecanism.
+- InPlace update policy is not part of this KEP since it adds more complexities like dynamically changing the topology envs at runtime. We may visit this in the future if highly required by the community.
+
+## Proposal
+
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation. What is the desired outcome and how do we measure success?.
+The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+### User Stories (Optional)
+
+<!--
+Detail the things that people will be able to do if this KEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
+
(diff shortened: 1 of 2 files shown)


keps/552-worker-resizing/README.md
```diff
@@ -0,0 +1,306 @@
...
+	// +kubebuilder:default=None
+	// +kubebuilder:validation:Enum={None,InPlace,Restart}
+	// +optional
+    ResizePolicy ResizePolicyType `json:"resizePolicy,omitempty`
+}
+```
+
+An update will be made to the `LeaderWorkerSet` validating webhook to retain its current behavior which prevents updates to `spec.leaderWorkerTemplate.size` only if the `ResizePolicy` is `None`.
+If the policy is `InPlace`, the underlying leader pod and statefulset will be resized accordingly.
+For the `Restart` policy, a new annotation `leaderworkerset.sigs.k8s.io/size` describing the targeted size will be added to the `StatefulSets` pod template to trigger a rollout of the worker pods.
```
Edwinhr716: If you trigger a rollout on the worker pods here, all of the replicas will be restarted at the same time. And if RestartGroupOnPodRestart is set (which is the default policy), all leader + worker pods will be restarted at the same time. Is this the intended behavior? Or should we go through rolling update for this feature?

keps/552-worker-resizing/README.md
```diff
@@ -0,0 +1,306 @@
...
+    // Restart updates the size and restarts existing pods.
+	// +kubebuilder:default=None
+	// +kubebuilder:validation:Enum={None,InPlace,Restart}
+	// +optional
+    ResizePolicy ResizePolicyType `json:"resizePolicy,omitempty`
+}
+```
+
+An update will be made to the `LeaderWorkerSet` validating webhook to retain its current behavior which prevents updates to `spec.leaderWorkerTemplate.size` only if the `ResizePolicy` is `None`.
+If the policy is `InPlace`, the underlying leader pod and statefulset will be resized accordingly.
```
Edwinhr716: For the `InPlace` size policy, what happens with the `size` annotation? Are we just not adding it? Or do the extra pods not have this label, but the original pods do? 

How does this feature interact with the SubGroupPolicy feature?

keps/552-worker-resizing/README.md
```diff
@@ -0,0 +1,306 @@
...
+<!--
+Detail the things that people will be able to do if this KEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
+
+As a user, I should be able to update the size of the workers without having to recreate the `LeaderWorkerSet` instance, and be assured that my workloads will be restarted if needed.
```
kerthcet: Could you elaborate more here, I just saw what we need rather than the reason which will help people to understand this feature better. 

Maybe a real user case would be better. Thanks.

keps/552-worker-resizing/README.md
kerthcet: Yes, this is critical. We not only have them in annotations but also the envs.

kerthcet: Will you use the size env in your environment?

keps/552-worker-resizing/README.md
synthe102: This is the intended behavior. I guess that rolling update could be added in a second phase ?

keps/552-worker-resizing/README.md
synthe102: In this proposal, the `size` annotation is used just to trigger the rollout. Not provide actual information. My plan with `InPlace` was just to not add the annotation.
If this is too confusing regarding the usefulness of the annotation, we could use a less confusing one: `leaderworkerset.sigs.k8s.io/resized-at: <timestamp>` or event `leaderworkerset.sigs.k8s.io/restarted-at: <timestamp>` ?

synthe102: It seems that `InPlace` brings more problems to the table than solutions. At least for us it doesn't make sense. I added it to the KEP because it might be nice for other users.
Based on this discussion, it might be worth to consider only `None` and `Restart`.

keps/552-worker-resizing/README.md
```diff
@@ -0,0 +1,312 @@
...
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+This KEP aims to allow resizing of workers by introducing a new field that defines the type of resize strategy. The current strategy will become `None` to keep the same behavior. `InPlace` and `RollingRestart` will be added. The `spec.leaderWorkerTemplate.size` field will become editable by relaxing the validating webhook.
```
kerthcet: No `InPlace` now right?

keps/552-worker-resizing/README.md
```diff
@@ -0,0 +1,312 @@
...
+
+
+### Non-Goals
+
+<!--
+What is out of scope for this KEP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+- This KEP doesn't propose a way to target `spec.leaderWorkerTemplate.size` with an autoscaling mecanism.
```
kerthcet, line 97: Could we also add the InPlace here since we've discussed, just like
- InPlace update policy is not part of this KEP since it adds more complexities like dynamically changing the topology envs at runtime. We may visit this in the future if highly required by the community.

synthe102: Sure ! I'll update the KEP.

keps/552-worker-resizing/README.md
Edwinhr716: If `RollingUpdate` is set in the yaml, the expected behavior is for the controller to update each replica one by one. Not sure if we want to keep that behavior when only changing `leaderWorkerTemplate`. @kerthcet thoughts?

keps/552-worker-resizing/README.md
```diff
@@ -0,0 +1,310 @@
...
+proposal will be implemented, this is the place to discuss them.
+-->
+
+A new field is added to the `LeaderWorkerTemplate` struct: `ResizePolicy`. It controls the way updates to `spec.leaderWorkerTemplate.size` are handled
+
+```go
+type ResizePolicyType string
+
+const ResizePolicyNone ResizePolicyType = "None"
+const ResizePolicyRestart ResizePolicyType = "Restart"
```
Edwinhr716: Thoughts on renaming this to "Recreate"? In this scenario, we are deleting the pods and creating them again, forcing it to go through the whole scheduling loop again. 

See https://github.com/kubernetes-sigs/jobset/issues/918#issuecomment-2970408748 for a related discussion on Recreate vs Restart

synthe102: That makes sense, let's rename it for cohesion. Thanks for bringing this up.

keps/552-worker-resizing/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+<!--
+This question should be filled when targeting a release.
+For Alpha, describe what tests will be added to ensure proper quality of the enhancement.
+
+For Beta and GA, add links to added tests together with links to k8s-triage for those tests:
+https://storage.googleapis.com/k8s-triage/index.html
+-->
+
+- <test>: <link to test coverage>
```
kerthcet: We should add two more integration tests here:
- Once resizePolicy=recreate, resize will trigger the recreate process successfully
- Once resizePolicy=recreate, resize in rolling update will trigger the recreate process successfully

Also, we should add the webhool validation tests here.

keps/552-worker-resizing/README.md
```diff
@@ -0,0 +1,310 @@
...
+This question should be filled when targeting a release.
+For Alpha, describe what tests will be added to ensure proper quality of the enhancement.
+
+For Beta and GA, add links to added tests together with links to k8s-triage for those tests:
+https://storage.googleapis.com/k8s-triage/index.html
+
+We expect no non-infra related flakes in the last month as a GA graduation criteria.
+-->
+
+- <test>: <link to test coverage>
```
kerthcet: One e2e test: resize will trigger the recreate process once resizePolicy=recreate and recover the service as expected.

keps/552-worker-resizing/README.md
```diff
@@ -0,0 +1,310 @@
...
+<!--
+Major milestones in the lifecycle of a KEP should be tracked in this section.
+Major milestones might include:
+- the `Summary` and `Motivation` sections being merged, signaling SIG acceptance
+- the `Proposal` section being merged, signaling agreement on a proposed design
+- the date implementation started
+- the first Kubernetes release where an initial version of the KEP was available
+- the version of Kubernetes where the KEP graduated to general availability
+- when the KEP was retired or superseded
+-->
```
kerthcet, line 303: Add the history here, like:
- 2025-05-29: KEP initialized and submitted for review

kerthcet:
And sorry for the late response, I'm relocating. @synthe102

kerthcet:
Hi @synthe102 do you have time to push this forward? Also cc @Edwinhr716 for further comments.

synthe102:
@kerthcet Sorry I had a busy week offsite. I'll push an update in the next 2 days. Thanks for your feedbacks !

Edwinhr716:
>  Also cc @Edwinhr716 for further comments.

The only discussion still open is on whether or not the `Recreate` policy triggers rolling update or not. I'm my opinion, if I have rollingUpdate set, I would expect that changing the size field would trigger a rolling update.

synthe102:
> > Also cc @Edwinhr716 for further comments.
> 
> The only discussion still open is on whether or not the `Recreate` policy triggers rolling update or not. I'm my opinion, if I have rollingUpdate set, I would expect that changing the size field would trigger a rolling update.

I'm in favor of this behavior as well. @kerthcet if you agree, I'll do a final update on the KEP.

kerthcet:
> I'm in favor of this behavior as well. @kerthcet if you agree, I'll do a final update on the KEP.

+1 I think rolling update is the only way now since we hardcoded some environments like LWS_GROUP_SIZE, have on idea how to support the in-place update in the future 😕. 

Would you like to add that part then we're happy to merge this one.

synthe102:
Added a line to specify that if `rollingUpdate` is set, updating the size will trigger a rolling update.
I think we're good then :D

kerthcet:
> The only discussion still open is on whether or not the Recreate policy triggers rolling update or not. I'm my opinion, if I  have rollingUpdate set, I would expect that changing the size field would trigger a rolling update.

I think @Edwinhr716 concern is solved now, so we're ready to merge.
/lgtm
/approve

PR #563, opened 2025-06-16
[WIP] add partition update

### LINKED ISSUE #511: Support lws partitioned for update

lixin010:
#### What type of PR is this?
/kind feature

add partition update for the canary deployment

Fixes #511 

```release-note

```

### DIFF
--- a/api/leaderworkerset/v1/leaderworkerset_types.go
+++ b/api/leaderworkerset/v1/leaderworkerset_types.go
@@ -248,6 +248,16 @@ const (
 
 // RollingUpdateConfiguration defines the parameters to be used for RollingUpdateStrategyType.
 type RollingUpdateConfiguration struct {
+	// Partition indicates the ordinal at which the lws should be partitioned for updates.
+	// During a rolling update, all the groups from ordinal Partition to Replicas-1 will be updated.
+	// The groups from 0 to Partition-1 will not be updated. 
+	// This is helpful in being able to do a canary based deployment. 
+	// The default value is 0.
+	//
+	// +optional
+	// +kubebuilder:default=0
+	Partition *int32 `json:"partition,omitempty"`
+
 	// The maximum number of replicas that can be unavailable during the update.
 	// Value can be an absolute number (ex: 5) or a percentage of total replicas at the start of update (ex: 10%).
 	// Absolute number is calculated from percentage by rounding down.
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -336,6 +336,9 @@ func (r *LeaderWorkerSetReconciler) rollingUpdateParameters(ctx context.Context,
 func (r *LeaderWorkerSetReconciler) SSAWithStatefulset(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, partition, replicas int32, revisionKey string) error {
 	log := ctrl.LoggerFrom(ctx)
 
+	// Limit the replicas with less than lwsPartition will not be updated.
+	partition = max(partition, *lws.Spec.RolloutStrategy.RollingUpdateConfiguration.Partition)
+
 	// construct the statefulset apply configuration
 	leaderStatefulSetApplyConfig, err := constructLeaderStatefulSetApplyConfiguration(lws, partition, replicas, revisionKey)
 	if err != nil {
@@ -384,8 +387,10 @@ func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *l
 	}
 
 	updateStatus := false
-	readyCount, updatedCount, updatedNonBurstWorkerCount, currentNonBurstWorkerCount, updatedAndReadyCount := 0, 0, 0, 0, 0
+	readyCount, updatedCount := 0, 0
+	partitionedUpdatedNonBurstCount, partitionedCurrentNonBurstCount, partitionedUpdatedAndReadyCount := 0, 0, 0
 	noWorkerSts := *lws.Spec.LeaderWorkerTemplate.Size == 1
+	lwsPartition := *lws.Spec.RolloutStrategy.RollingUpdateConfiguration.Partition
 
 	// Iterate through all leaderPods.
 	for _, pod := range leaderPodList.Items {
@@ -405,8 +410,8 @@ func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *l
 			}
 		}
 
-		if index < int(*lws.Spec.Replicas) {
-			currentNonBurstWorkerCount++
+		if index < int(*lws.Spec.Replicas) && index >= int(lwsPartition) {
+			partitionedCurrentNonBurstCount++
 		}
 
 		var ready, updated bool
@@ -417,16 +422,16 @@ func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *l
 		if (noWorkerSts || revisionutils.GetRevisionKey(&sts) == revisionKey) && revisionutils.GetRevisionKey(&pod) == revisionKey {
 			updated = true
 			updatedCount++
-			if index < int(*lws.Spec.Replicas) {
+			if index < int(*lws.Spec.Replicas) && index >= int(lwsPartition) {
 				// Bursted replicas do not count when determining if rollingUpdate has been completed.
-				updatedNonBurstWorkerCount++
+				partitionedUpdatedNonBurstCount++
 			}
 		}
 
 		if ready && updated {
 			// Bursted replicas should not be counted here.
-			if index < int(*lws.Spec.Replicas) {
-				updatedAndReadyCount++
+			if index < int(*lws.Spec.Replicas) && index >= int(lwsPartition) {
+				partitionedUpdatedAndReadyCount++
 			}
 		}
 	}
@@ -442,18 +447,19 @@ func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *l
 	}
 
 	var conditions []metav1.Condition
-	updateDone := false
-	if updatedNonBurstWorkerCount < currentNonBurstWorkerCount {
+	if partitionedUpdatedNonBurstCount < partitionedCurrentNonBurstCount {
 		// upgradeInProgress is true when the upgrade replicas is smaller than the expected
 		// number of total replicas not including the burst replicas
 		conditions = append(conditions, makeCondition(leaderworkerset.LeaderWorkerSetUpdateInProgress))
 		conditions = append(conditions, makeCondition(leaderworkerset.LeaderWorkerSetProgressing))
-	} else if updatedAndReadyCount == int(*lws.Spec.Replicas) {
+	} else if partitionedUpdatedAndReadyCount == int(*lws.Spec.Replicas)-int(lwsPartition) {
 		conditions = append(conditions, makeCondition(leaderworkerset.LeaderWorkerSetAvailable))
-		updateDone = true
 	} else {
 		conditions = append(conditions, makeCondition(leaderworkerset.LeaderWorkerSetProgressing))
 	}
+
+	// updateDone is true when all replicas are updated and ready
+	updateDone := (lwsPartition == 0) && partitionedUpdatedAndReadyCount == int(*lws.Spec.Replicas)
 
 	updateCondition := setConditions(lws, conditions)
 	// if condition changed, record events


kerthcet:
Ping @lixin010 are u still working on this ? We hope to make it part of v0.7.0

lixin010:
> Ping @lixin010 are u still working on this ? We hope to make it part of v0.7.0

That's great. I'll finish it as soon as possible.

lixin010:
@ahg-g @ardaguclu @kerthcet  Please review the code and tag `ok-to-test` label.

kerthcet:
Seems tests failed.

lixin010:
@kerthcet  Now, the test is ok.

lixin010:
@ahg-g @ardaguclu Please help review the code when you have time. Thanks.

lixin010:
Hi @ahg-g @ardaguclu. This PR has very few lines of code. Could you spare some time to review?

ardaguclu:
From my side, it is better to be reviewed by @kerthcet, when he is available.

kerthcet:
Will do this week.

xigang:
@lixin010 I suggest you calculate the partition directly within the `rollingUpdateParameters` function.

Add a case:

```
	if lws.Spec.RolloutStrategy.RollingUpdateConfiguration.Partition != nil {
		userPartition := *lws.Spec.RolloutStrategy.RollingUpdateConfiguration.Partition

		// Validate user partition
		if userPartition > lwsReplicas {
			log.V(2).Info("User partition is greater than replicas, setting to replicas", "partition", userPartition, "replicas", lwsReplicas)
			userPartition = lwsReplicas
		}
		if userPartition < 0 {
			log.V(2).Info("User partition is negative, setting to 0", "partition", userPartition)
			userPartition = 0
		}

		states, err := r.getReplicaStates(ctx, lws, stsReplicas, revisionKey)
		if err != nil {
			return 0, 0, err
		}
		lwsUnreadyReplicas := calculateLWSUnreadyReplicas(states, lwsReplicas)

		return userPartition, wantReplicas(lwsUnreadyReplicas), nil
	}

```

xigang:
Should `MaxUnavailable` also be considered when adjusting the partition?

lixin010:
Hello @xigang. Thanks for your advise. At first, I also placed it in the `rollingUpdateParameters` function. In fact, the `rollingUpdateParameters` function calculates which groups can be updated based on the `MaxUnavailable` and `MaxSurge` parameters, and controls the step of updates by setting the `sts partition` field. The `lws partition` field only needs to restrict that the `sts partition` to be updated will not be smaller than the `lws partition`. In this way, if the serial number is smaller than the lws partition, it will not be updated.

xigang:
Got it. How about putting the logic `max(partition, *lws.Spec.RolloutStrategy.RollingUpdateConfiguration.Partition)` into `rollingUpdatePartition?` That way, the responsibilities would be clearer.😄

```
func rollingUpdatePartition(states []replicaState, stsReplicas int32, rollingStep int32, currentPartition int32, lwsPartition int32) int32 {
	//......

	// That means Partition moves in one direction to make it simple.
	// First apply unidirectional decrease constraint
	partition = min(partition, currentPartition)

	// Then apply lws partition constraint, but don't allow increasing partition
	// This respects user settings while maintaining unidirectional movement
	if lwsPartition > currentPartition {
		// If user-set partition is larger than current, it will take effect in next rolling update
		return partition
	} else {
		// Ensure partition is not lower than user-set partition
		return max(lwsPartition, partition)
	}
}
```

lixin010:
> Got it. How about putting the logic `max(partition, *lws.Spec.RolloutStrategy.RollingUpdateConfiguration.Partition)` into `rollingUpdatePartition?` That way, the responsibilities would be clearer.😄
> 
> ```
> func rollingUpdatePartition(states []replicaState, stsReplicas int32, rollingStep int32, currentPartition int32, lwsPartition int32) int32 {
> 	//......
> 
> 	// That means Partition moves in one direction to make it simple.
> 	// First apply unidirectional decrease constraint
> 	partition = min(partition, currentPartition)
> 
> 	// Then apply lws partition constraint, but don't allow increasing partition
> 	// This respects user settings while maintaining unidirectional movement
> 	if lwsPartition > currentPartition {
> 		// If user-set partition is larger than current, it will take effect in next rolling update
> 		return partition
> 	} else {
> 		// Ensure partition is not lower than user-set partition
> 		return max(lwsPartition, partition)
> 	}
> }
> ```

It may not be feasible to place it in the `rollingUpdatePartition` function. This is because the `rollingUpdateParameters` function considers multiple cases, and not all of them are returned from the `rollingUpdatePartition` function.

xigang:
Yeah, the logic looks fine. Let's wait for @kerthcet  review.

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -248,6 +248,16 @@ const (
 
 // RollingUpdateConfiguration defines the parameters to be used for RollingUpdateStrategyType.
 type RollingUpdateConfiguration struct {
+	// Partition indicates the ordinal at which the lws should be partitioned for updates.
+	// During a rolling update, all the groups from ordinal Partition to Replicas-1 will be updated.
+	// The groups from 0 to Partition-1 will not be updated.
+	// This is helpful in being able to do a canary based deployment.
```
kerthcet:
```suggestion
	// This is helpful in incremental rollout strategies like canary deployments or interactive rollout strategies for multiple replicas like xPyD deployments.
```

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -336,6 +336,9 @@ func (r *LeaderWorkerSetReconciler) rollingUpdateParameters(ctx context.Context,
 func (r *LeaderWorkerSetReconciler) SSAWithStatefulset(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, partition, replicas int32, revisionKey string) error {
 	log := ctrl.LoggerFrom(ctx)
 
+	// Limit the replicas with less than lwsPartition will not be updated.
+	partition = max(partition, *lws.Spec.RolloutStrategy.RollingUpdateConfiguration.Partition)
```
kerthcet: Can we move this to the `rollingUpdatePartition`, or the partition logic is scattered across the project.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -442,19 +449,20 @@ func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *l
 	}
 
 	var conditions []metav1.Condition
-	updateDone := false
-	if updatedNonBurstWorkerCount < currentNonBurstWorkerCount {
+	if lwsPartition < *lws.Spec.Replicas && partitionedUpdatedNonBurstCount < partitionedCurrentNonBurstCount {
```
kerthcet: Why `lwsPartition < *lws.Spec.Replicas` here? Once `partitionedUpdatedNonBurstCount < partitionedCurrentNonBurstCount`, doesn't it mean it's always rolling update now?

Please update the comment as well, thanks.

test/integration/controllers/leaderworkerset_test.go
```diff
@@ -1963,6 +1963,102 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
...
+						testing.ExpectLeaderWorkerSetStatusReplicas(ctx, k8sClient, lws, 6, 2)
+					},
+				},
+				{
+					// Set lws partition to 2, rolling update start
+					lwsUpdateFn: func(lws *leaderworkerset.LeaderWorkerSet) {
+						testing.SetLwsPartition(ctx, k8sClient, lws, 2)
+					},
+					checkLWSState: func(lws *leaderworkerset.LeaderWorkerSet) {
+						testing.ExpectLeaderWorkerSetProgressing(ctx, k8sClient, lws, "Replicas are progressing")
```
kerthcet, line 2071: Since we're in rolling update, why we'll hit the progressing condition?

kerthcet:
I tried to find you in slack, but I haven't found .. So the deadline of this release is the end of this month, we really hope to include this feature, ping me once you're ready for another review. Thanks @lixin010

test/integration/controllers/leaderworkerset_test.go
kerthcet: I guess it's because we will start a new rolling update progress then we'll scale first?

lixin010:
```go
if lwsPartition < *lws.Spec.Replicas && partitionedUpdatedNonBurstCount < partitionedCurrentNonBurstCount {
	// upgradeInProgress is true when the upgrade replicas is smaller than the expected
	// number of total replicas not including the burst replicas
	conditions = append(conditions, makeCondition(leaderworkerset.LeaderWorkerSetUpdateInProgress))
	conditions = append(conditions, makeCondition(leaderworkerset.LeaderWorkerSetProgressing))
}
```
I'm a little confused. In a rolling update, the progressing state is always true. Isn't it so?

xigang:
> I have a question here like once we update the Partition field? Will we trigger a new rolling update from the scratch?

@kerthcet Looking at the code, the hash is only calculated using the `ControllerRevision`'s `leaderWorkerTemplate` and `networkConfig` field, so updating the LWS `Partition` field shouldn’t trigger a rolling update.

https://github.com/kubernetes-sigs/lws/blob/4cd9e42f75b156a999847e3adb76c0a658ab3ac1/pkg/utils/revision/revision_utils.go#L289-L299

https://github.com/kubernetes-sigs/lws/blob/4cd9e42f75b156a999847e3adb76c0a658ab3ac1/pkg/utils/revision/revision_utils.go#L221-L253

https://github.com/kubernetes-sigs/lws/blob/4cd9e42f75b156a999847e3adb76c0a658ab3ac1/pkg/utils/revision/revision_utils.go#L67-L87

api/leaderworkerset/v1/leaderworkerset_types.go
lixin010: done

pkg/controllers/leaderworkerset_controller.go
lixin010: I move this to the defer function within the `rollingUpdateParameters` function, because the rollingUpdateParameters function has multiple return branches.

pkg/controllers/leaderworkerset_controller.go
lixin010: Yes, the condition of `lwsPartition < *lws.Spec.Replicas` is redundant. Removed.

lixin010:
@kerthcet Yeah, updating lws partition field wiil not compute a new revision. So it will not trigger a rolling update. I did my best to make time and fix some code according to the review. I also hope that this feature can be released in the next version.

kerthcet:
> > I have a question here like once we update the Partition field? Will we trigger a new rolling update from the scratch?
> 
> @kerthcet Looking at the code, the hash is only calculated using the `ControllerRevision`'s `leaderWorkerTemplate` and `networkConfig` field, so updating the LWS `Partition` field shouldn’t trigger a rolling update.

Yeah, that's right.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -405,8 +412,8 @@ func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *l
 			}
 		}
 
-		if index < int(*lws.Spec.Replicas) {
-			currentNonBurstWorkerCount++
+		if index < int(*lws.Spec.Replicas) && index >= int(lwsPartition) {
+			partitionedCurrentNonBurstCount++
```
kerthcet, line 416: one question, can't we get the result directly? Which is more accurate I think.

test/integration/controllers/leaderworkerset_test.go
kerthcet: Yes, this is a bit different than before, when each condition type is mutually exclusive, anyway, not related to this PR. Let's forget it for now.

test/integration/controllers/leaderworkerset_test.go
```diff
@@ -1963,6 +1963,102 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
 				},
 			},
 		}),
+		ginkgo.Entry("rolling update with the partition setting", &testCase{
+			makeLeaderWorkerSet: func(nsName string) *wrappers.LeaderWorkerSetWrapper {
```
kerthcet, line 2010: Can you add one more test with maxSurge set, and the partition field will update to 0 in the end.

kerthcet (commented):
One thing we should notice here is that: with partition field set, and rolling update finished, we'll have more replicas then `spec.replicas`. Is that as expected? I guess no but if we take this into account, it will leads to more complex calculation for the partition field, as well as the replicas. What's your opinion?

lixin010:
> One thing we should notice here is that: with partition field set, and rolling update finished, we'll have more replicas then `spec.replicas`. Is that as expected? I guess no but if we take this into account, it will leads to more complex calculation for the partition field, as well as the replicas. What's your opinion?

Yes, when the partition update complete, the surge replicas will not be deleted, which is what I expected. I believe that the lifecycle of surge replicas should belong to the entire update process, and a partition update is just one phase within the overall update. Designing it this way keeps the logic simpler and avoids introducing too many scaling operations. For example, when aligned releases across multiple roles, the interval between two partition updates is usually not very long. This approach reduces complexity and is friendlier to the upper-level scheduling logic.

pkg/controllers/leaderworkerset_controller.go
lixin010: I think using the index is a straightforward solution. I haven't found any cases where it might be inaccurate. The semantic of the partition field is also based on the replica index.

test/integration/controllers/leaderworkerset_test.go
lixin010: Will add this case later.

kerthcet:
> > One thing we should notice here is that: with partition field set, and rolling update finished, we'll have more replicas then `spec.replicas`. Is that as expected? I guess no but if we take this into account, it will leads to more complex calculation for the partition field, as well as the replicas. What's your opinion?
> 
> Yes, when the partition update complete, the surge replicas will not be deleted, which is what I expected. I believe that the lifecycle of surge replicas should belong to the entire update process, and a partition update is just one phase within the overall update. Designing it this way keeps the logic simpler and avoids introducing too many scaling operations. For example, when aligned releases across multiple roles, the interval between two partition updates is usually not very long. This approach reduces complexity and is friendlier to the upper-level scheduling logic.

Yeah, we can keep as it is and see what's going on in the future, let's add one comment to the api field, and I think it's fair enough.

pkg/controllers/leaderworkerset_controller.go
kerthcet: for example, the Pod is deleted then the number of list items is not right eventually, is that one?

lixin010: In the logic without the partition update feature, the UpdateInProgress state is true when currentNonBurstWorkerCount == updatedNonBurstCount. If the pod is deleted, the currentNonBurstWorkerCount < lws.Spec.Replicas and currentNonBurstWorkerCount  ==  updatedNonBurstCount. So the UpdateInProgress state is false and the Progressing state is true. I have kept this logic.

lixin010:
@kerthcet The test case has been added. cc.

wewang58:
@kerthcet Hi, I have a question,  if there's  a maximum quantity limit for replicas and size  in lws? thanks in advance.

kerthcet:
> @kerthcet Hi, I have a question, if there's a maximum quantity limit for replicas and size in lws? thanks in advance.

No I think, do you have specific usecases?

kerthcet:
/lgtm
/approve

Thanks @lixin010 This is great, I'll try to prototype a xPyD framework based on this feature.

lixin010:
@kerthcet Do you know the reason why the pull-lws-verify-main check failed ?

yankay:
The issue can be reproduced by running make verify.​​
​​It's better to use Git squashing (merging multiple commits into one).​

lixin010:
@kerthcet I've rebased onto the main branch and squashed the commits to fix the check. Please help tag `lgtm` again. Thanks.

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -263,6 +263,18 @@ const (
 
 // RollingUpdateConfiguration defines the parameters to be used for RollingUpdateStrategyType.
 type RollingUpdateConfiguration struct {
+	// Partition indicates the ordinal at which the lws should be partitioned for updates.
+	// During a rolling update, all the groups from ordinal Partition to Replicas-1 will be updated.
+	// The groups from 0 to Partition-1 will not be updated.
+	// This is helpful in incremental rollout strategies like canary deployments
+	// or interactive rollout strategies for multiple replicas like xPyD deployments.
+	// When the partition update complete, the surged replicas will keep running.
```
kerthcet:
```suggestion
Once partition field and maxSurge field both set, the bursted replicas will keep remaining until the rolling update is completely done and the partition field is reset to 0. This is as expected to reduce the reconciling complexity.
```

kerthcet (commented):
I just found one comment we can improve, would you like to address that, I'll put a LGTM then. Thanks for this. @lixin010

api/leaderworkerset/v1/leaderworkerset_types.go
lixin010: ok

lixin010:
@kerthcet  Done.

kerthcet:
Thanks @lixin010 for your efforts.
/lgtm

Let's go!

wewang58:
> > @kerthcet Hi, I have a question, if there's a maximum quantity limit for replicas and size in lws? thanks in advance.
> 
> No I think, do you have specific usecases?
Thanks response, our performance qe wants to know if there's maximum quantity limit for her performance test for lws.

kerthcet:
If you have any perf result,  plz share with us. thanks.

wewang58:
> If you have any perf result, plz share with us. thanks.

Ok, no problem.

PR #568, opened 2025-06-25
fix rolling update stuck that when old pods were not ready

### LINKED ISSUE #567: RollingUpdate stuck if all old pod are not ready
**What happened**:
Rolling update stuck if all old pod were not ready.

Rolling update successfully.

old.yaml
```yaml
kind: LeaderWorkerSet
metadata:
  name: leaderworkerset-rollout
spec:
  rolloutStrategy:
    type: RollingUpdate
    rollingUpdateConfiguration:
      maxUnavailable: 0
      maxSurge: 1
  replicas: 4
  leaderWorkerTemplate:
    size: 1
    workerTemplate:
      spec:
        containers:
          - name: nginx
            image: image-not-exist:v1
            ports:
              - containerPort: 8080
```
new.yaml
```yaml
kind: LeaderWorkerSet
metadata:
  name: leaderworkerset-rollout
spec:
  rolloutStrategy:
    type: RollingUpdate
    rollingUpdateConfiguration:
      maxUnavailable: 0
      maxSurge: 1
  replicas: 4
  leaderWorkerTemplate:
    size: 1
    workerTemplate:
      spec:
        containers:
          - name: nginx
            image: nginx:latest
            ports:
              - containerPort: 8080
```
 
```
kubectl get po

NAME                        READY   STATUS         RESTARTS   AGE
leaderworkerset-rollout-0   0/1     ErrImagePull   0          42s
leaderworkerset-rollout-1   0/1     ErrImagePull   0          42s
leaderworkerset-rollout-2   0/1     ErrImagePull   0          42s
leaderworkerset-rollout-3   0/1     ErrImagePull   0          42s
leaderworkerset-rollout-4   1/1     Running        0          22s

```

v0.6.2 works well,  but fails on the main branch .

**Environment**:

Client Version: v1.32.1
Kustomize Version: v5.5.0
Server Version: v1.33.1-aliyun.1

v0.6.1-29-gf9a6396-dirty

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

gujingit:
#### What type of PR is this?
/kind bug

fix rolling update stuck that when old pods were not ready

Fixes #567 

```release-note

```

### DIFF
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -611,7 +611,7 @@ func rollingUpdatePartition(states []replicaState, stsReplicas int32, rollingSte
 	// (for example, all replicas are not ready when rolling update is started).
 	// Note that we never drop the partition below rolliingStepPartition.
 	for idx := min(partition, stsReplicas-1); idx >= rolliingStepPartition; idx-- {
-		if !states[idx].ready {
+		if !states[idx].ready || states[idx].updated {
 			partition = idx
 		} else {
 			break
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -163,6 +163,23 @@ var _ = ginkgo.Describe("leaderWorkerSet e2e tests", func() {
 		testing.ExpectLeaderWorkerSetAvailable(ctx, k8sClient, lws, "All replicas are ready")
 	})
 
+	ginkgo.It("Can perform a rolling update even if old lws not ready", func() {
+		// Create lws with not exist image.
+		lws := wrappers.BuildLeaderWorkerSet(ns.Name).LeaderTemplate(nil).Size(1).Replica(2).MaxSurge(1).MaxUnavailable(0).Obj()
+		lws.Spec.LeaderWorkerTemplate.WorkerTemplate.Spec.Containers[0].Image = "not-exist-image:v1"
+		testing.MustCreateLws(ctx, k8sClient, lws)
+
+		//Update lws.
+		testing.UpdateWorkerTemplateImage(ctx, k8sClient, lws)
+
+		// Rolling update completes.
+		testing.ExpectValidLeaderStatefulSet(ctx, k8sClient, lws, 2)
+		testing.ExpectValidWorkerStatefulSets(ctx, lws, k8sClient, true)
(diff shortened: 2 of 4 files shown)


ahg-g (commented):
Thanks, any tests we can add to cover this bug?

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -611,9 +611,10 @@ func rollingUpdatePartition(states []replicaState, stsReplicas int32, rollingSte
 	// (for example, all replicas are not ready when rolling update is started).
 	// Note that we never drop the partition below rolliingStepPartition.
 	for idx := min(partition, stsReplicas-1); idx >= rolliingStepPartition; idx-- {
-		if !states[idx].ready {
-			partition = idx
+		if states[idx].ready && states[idx].updated {
```
kerthcet: Also involve @pierewoj in as the author of this.

pierewoj: This does not seem right to me. It means that we will ignore the readiness of the replicas that have not been updated and will move the partition potentially causing avail drop, or no?

We already have `rolling update with no ready replicas` Integ test that prevent a case that was reported in the https://github.com/kubernetes-sigs/lws/issues/567. However we don't test the update with maxSurge>0 and potentially that's where the issue comes from.

I am on vacation atm, but likely can find a bit of time in the evening to try to repro and see what could be a fix.

pierewoj: Spent a short while on this, I was able to reproduce this with an integ test. I agree the issue is that in this logic here we don't consider that it's actually OK to move the partition if the replica is already updated. But the proper condition seems to be:
```
if !states[idx].ready || states[idx].updated {
	partition = idx
} else {
	break
}
```

And with that the repro test (and other integ tests) succeed.

kerthcet: Maybe we should be more strict here like `if !states[idx].ready || (states[idx].updated && states[idx].ready)`, what do your guys think? @pierewoj @gujingit It follows the same pattern in lws now as we try to make sure the updated ones should be ready first.

pierewoj: Aren't these equivalent? `~a || b === ~a || (b && a)`

kerthcet: true.
/lgtm
/approve

Thanks all!

PR #569, opened 2025-06-26
improve ci: add auto sync crds to helm chart

nicole-lihui:
#### What type of PR is this?

/kind failing-test

#### Which issue(s) this PR fixes

Fixes #537

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #576, opened 2025-07-10
feat: allow updates to spec.leaderWorkerTemplate.size

### LINKED ISSUE #552: Allow to update `spec.leaderWorkerTemplate.size`

synthe102:
#### What type of PR is this?

/kind feature
/kind api-change

This PR implements KEP 552 by introducing a new field: `spec.leaderWorkerTemplate.resizePolicy`. This field defaults to `None` which keeps the current behavior of not allowing updates to the `size` field. If the value is set to `Recreate`, updates to `spec.leaderWorkerTemplate.size` will be allowed.

Fixes #552 

```release-note
Introduce `resizePolicy` to control updates to the `size` field.
```


api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -169,6 +176,14 @@ type LeaderWorkerTemplate struct {
 	// in each replica.
 	// +optional
 	SubGroupPolicy *SubGroupPolicy `json:"subGroupPolicy,omitempty"`
+
+	// ResizePolicy defines how to handle worker size updates.
+	// None doesn't allow to update spec.leaderWorkerTemplate.size.
+	// Recreate updates the size and restarts existing pods.
```
kerthcet:
```suggestion
	// ResizePolicy defines how to handle group size updates:
	// - None: indicates the `spec.leaderWorkerTemplate.size` is immutable.
	// - Recreate: indicates recreating the group Pods to reshape the size.
```

test/testutils/util.go
```diff
@@ -392,6 +392,21 @@ func HasLWSEnvVarsPopulated(pod corev1.Pod) bool {
 	return hasAllEnvVarPopulated(pod, []string{leaderworkerset.LwsLeaderAddress, leaderworkerset.LwsGroupSize, leaderworkerset.LwsWorkerIndex})
 }
 
+func CheckSizeAnnotation(pod corev1.Pod, size int) error {
```
kerthcet: Can we make this move general like`CheckAnnotation(pood, key, value)`

kerthcet (commented):
To make sure I understand the whole process: once we change the size, the leader sts will update the size value in the pod template annotation, which will trigger the rolling update of leader pods, then the worker pods, right?

Would you like to take another look? @Edwinhr716

Edwinhr716:
> once we change the size, the leader sts will update the size value in the pod template annotation, which will trigger the rolling update of leader pods, then the worker pods

Yes that would happen, but using this current implementation we don't trigger a rolling update at the LWS controller level.

Edwinhr716:
Actually, nevermind. Looking at the revision code closely, a rolling update is triggered whenever there is a change in the LeaderWorkerTemplate https://github.com/kubernetes-sigs/lws/blob/main/pkg/utils/revision/revision_utils.go#L247, which the size parameter is a part of. So from the LWS controller level, the change in the size parameter is the one that triggers the rolling update, and then the annotation triggers the change at the Leader Sts level

Edwinhr716:
LGTM overall. @synthe102 did you do manual testing to confirm that a rolling update is triggered at the LWS level? If you check the events of the LWS object during it and it shows a RollingUpdateInProgress event, then it that would confirm that it was at the LWS level

test/testutils/util.go
synthe102: Sure !

synthe102:
Thanks both of you for your reviews.
@Edwinhr716 I confirm that when updating the size field, I get this event:
`default   32s         Normal  GroupsUpdating     leaderworkerset/leaderworkerset-sample            leaderworkerset  Rolling Upgrade is in progress, with 4 groups ready of total 4 groups`

kerthcet:
> d it shows a RollingUpdateInProgress eve

Let's add a `ValidateEvent` then, then we can merge I think.

synthe102:
Just to be sure, you want add a new Event Reason `Validate` that will be emitted when a resize occurs ? I'm not sure I understand where this event will be used, could you clarify ?

config/crd/bases/leaderworkerset.x-k8s.io_leaderworkersets.yaml
```diff
@@ -8184,6 +8184,16 @@ spec:
...
+                    default: None
+                    description: |-
+                      ResizePolicy defines how to handle group size updates:
+                      - None: indicates the `spec.leaderWorkerTemplate.size` is immutable.
+                      - Recreate: indicates recreating the group Pods to reshape the size.
+                    enum:
+                    - None
+                    - Recreate
+                    type: string
                   restartPolicy:
```
yankay: Do the CRDs in the charts also need to be modified :-)

ref: https://github.com/kubernetes-sigs/lws/blob/main/charts/lws/templates/crds/leaderworkerset.x-k8s.io_leaderworkersets.yaml

synthe102: Good catch. It seems that its required. I added a copy step to `make manifests` that copies the CRDs from the kustomization directory to the chart.

kerthcet:
> Just to be sure, you want add a new Event Reason `Validate` that will be emitted when a resize occurs ? I'm not sure I understand where this event will be used, could you clarify ?

Sorry, I mean we can add a event validation like this: https://github.com/kubernetes-sigs/lws/blob/82b36869d3f774bd6b022353b434e813de84296a/test/integration/controllers/leaderworkerset_test.go#L205-L207

so solve Edwin's concern.

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -169,6 +176,14 @@ type LeaderWorkerTemplate struct {
...
 	// +optional
 	SubGroupPolicy *SubGroupPolicy `json:"subGroupPolicy,omitempty"`
+
+	// ResizePolicy defines how to handle group size updates:
+	// - None: indicates the `spec.leaderWorkerTemplate.size` is immutable.
+	// - Recreate: indicates recreating the group Pods to reshape the size.
+	// +kubebuilder:default=None
+	// +kubebuilder:validation:Enum={None,Recreate}
+	// +optional
+	ResizePolicy ResizePolicyType `json:"resizePolicy,omitempty"`
```
ahg-g: This is probably best to be a reference so that we can handle existing objects (basically treat nil and None as the same)

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -103,7 +103,9 @@ func (r *LeaderWorkerSetWebhook) ValidateUpdate(ctx context.Context, oldObj, new
 
 	oldLws := oldObj.(*v1.LeaderWorkerSet)
 	newLws := newObj.(*v1.LeaderWorkerSet)
-	allErrs = append(allErrs, apivalidation.ValidateImmutableField(*newLws.Spec.LeaderWorkerTemplate.Size, *oldLws.Spec.LeaderWorkerTemplate.Size, field.NewPath("spec", "leaderWorkerTemplate", "size"))...)
+	if newLws.Spec.LeaderWorkerTemplate.ResizePolicy == v1.ResizePolicyNone {
```
ahg-g: For existing objects, this field will not be set, and so this condition will evaluate to false, which breaks the existing behavior, pls see comment on the api file for how we could handle this.

Makefile
```diff
@@ -120,6 +120,7 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 		crd:generateEmbeddedObjectMeta=true output:crd:artifacts:config=config/crd/bases \
 		webhook output:webhook:artifacts:config=config/webhook \
 		paths="{./api/..., ./pkg/...}"
+	cp config/crd/bases/* charts/lws/templates/crds
```
kerthcet: Can we remove this? It's working on https://github.com/kubernetes-sigs/lws/pull/569 to sync the crds.

test/integration/webhooks/leaderworkerset_test.go
```diff
@@ -261,6 +261,24 @@ var _ = ginkgo.Describe("leaderworkerset defaulting, creation and update", func(
 			},
 			lwsCreationShouldFail: true,
 		}),
+		ginkgo.Entry("update to size when resizePolicy is None should fail", &testValidationCase{
```
ahg-g, line 264: This is only testing the validation logic. Can we add an integration test to verify that the controller actually applies this change to the statefulsets?

synthe102: Ok. 👍

api/leaderworkerset/v1/leaderworkerset_types.go
kerthcet: @synthe102 Can you follow this? Since it's optional, let's make it `*ResizePolicyType` also not break the users.

kerthcet (commented):
I think we still miss three parts right now:
- change the `resizePolicy ` field to a pointer
- add an integration test to verify that the size is indeed updated
- add the rolling update event validation

All of these three are very simple ones I think, so we're very close now.  :)

api/leaderworkerset/v1/leaderworkerset_types.go
synthe102: Done ✅ I'll wrap up the remaining tests before end of day.

synthe102:
> I think we still miss three parts right now:
> 
>     * change the `resizePolicy ` field to a pointer
> 
>     * add an integration test to verify that the size is indeed updated
> 
>     * add the rolling update event validation
> 
> 
> All of these three are very simple ones I think, so we're very close now. :)

I think we should be good. I changed the `ResizePolicy` to a pointer, added an integration test to verify that the size is updated, and check that we have a rolling update event.

kerthcet:
Will take a look later today, Thanks @synthe102

test/integration/controllers/leaderworkerset_test.go
```diff
@@ -1971,6 +1971,41 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
...
+						testing.ExpectLeaderWorkerSetStatusReplicas(ctx, k8sClient, lws, 2, 2)
+					},
+				},
+				{
+					// Update size to 3.
+					lwsUpdateFn: func(lws *leaderworkerset.LeaderWorkerSet) {
+						testing.UpdateSize(ctx, k8sClient, lws, 3)
+						testing.SetPodGroupsToReady(ctx, k8sClient, lws, 2)
+					},
+					checkLWSState: func(lws *leaderworkerset.LeaderWorkerSet) {
```
kerthcet, line 1998: Nit: let's verify that before the size is 2, after update the size increased to 3, I think we missed that part.

kerthcet (commented):
/approve

Generally LGTM. Only one nit from my side.

test/integration/controllers/leaderworkerset_test.go
synthe102: I would say that `ExpectValidWorkersStatefulSets` being called before and after the resize takes care of that ? [Here](https://github.com/kubernetes-sigs/lws/blob/8d65262f630745031013318c8861d22fb892e738/test/testutils/validators.go#L286-L288) for the replicas and [here](https://github.com/kubernetes-sigs/lws/blob/8d65262f630745031013318c8861d22fb892e738/test/testutils/validators.go#L286-L288) for the size annotation.

kerthcet: Although not that straightforward, but fair enough, since we assume that the update is succeeded.

kerthcet:
/lgtm
Let's merge now since we addressed all the comments and other PRs may hit code conflicts with this one, we're close to the deadline, then can save some time for them to rebase. 

Thanks @synthe102 for the patient and efforts, it's great to see this feature to be merged in v0.7.0, also glad to see MistralAI as one of the users, maybe you're happy to join the list: https://lws.sigs.k8s.io/docs/adoption/

kerthcet:
/hold

Just to make sure the crd update for the helm chart is right? Since we changed a lot.

synthe102:
> /hold
> 
> Just to make sure the crd update for the helm chart is right? Since we changed a lot.

I copy/pasted the whole CRD from the `config/crd/bases` folder into `charts/lws/templates/crds` because I couldn't minimize the diff by tweaking `yamlfmt` and manually patching the file.
I did a last copy/paste to check and there's no diff.

kerthcet:
/hold cancel

I'll add a fix to the ci auto-sync. Thanks @synthe102

JesseStutler:
image
@synthe102 @kerthcet Hi, I noticed that the `resize shoud update the size of the replicas` integration test failed in my environment. Is it a bug or I missed something to configure in my environment? But other 44 cases can pass

kerthcet:
Do you mean the main branch? I checked locally, all tests passed.

JesseStutler:
> Do you mean the main branch? I checked locally, all tests passed.

Yes main branch, maybe it's because my environment's fault, never mind 😄

ahg-g:
Looking at this again, why do we need an api to select between two modes? why not just always allow resize? this is a backward compatible change since this is simply relaxing validation.

kerthcet:
> Looking at this again, why do we need an api to select between two modes? why not just always allow resize? this is a backward compatible change since this is simply relaxing validation.

We used to have three modes, none, recreate, updateInPlace, and the third one is really complex, so we dropped it for now. And what if people don't want to change the size?

ahg-g:
> And what if people don't want to change the size?

Then they shouldn't update it :)

kerthcet:
> Then they shouldn't update it :)

Yes, what I mean here is this is an user-facing change, but from the API perspective, it's indeed backward compatible. cc @synthe102 any feedback? I'm ok with both ways since new API explicitly indicates that the size(topology) is changeable or not, it's more than what replicas represents.

ahg-g:
Relaxing validation is generally a backward compatible change.

kerthcet:
> but from the API perspective, it's indeed backward compatible

Yes, what I mean here I agree that relaxing the validation is backward compatible. So you strongly prefer to remove the resizePolicy? @ahg-g

ahg-g:
Right, if this is only to be backward compatible, then we are covered without it, so best to not extend the api surface

synthe102:
@kerthcet I'm ok with always allowing resize and removing the `ResizePolicy`. How do you want to proceed ? I can draft a PR to removed the `ResizePolicy` field and adjust tests accordingly.

kerthcet:
Yes, if you're available, you can fire a PR. we can catch up with this new release then.

synthe102:
Opened #601 to remove the `resizePolicy` field.

Issue #593, opened 2025-07-30
runtime error: invalid memory address or nil pointer dereference


drikster80:
Installed LWS using the helm chart:
```bash
CHART_VERSION=0.6.2 helm install lws oci://registry.k8s.io/lws/charts/lws   --version=$CHART_VERSION   --namespace lws-system   --create-namespace   --wait --timeout 300s
```

I swore this worked a couple of days ago. I have reinstalled the chart (and deleted the namespace) and seeing the same error.

**What happened**:

Getting `"error": "panic: runtime error: invalid memory address or nil pointer dereference"` when deploying.

```
2025-07-30T15:28:09Z	DEBUG	events	Creating revision with key 9fd4f9c4 for a newly created LeaderWorkerSet	{"type": "Normal", "object": {"kind":"LeaderWorkerSet","namespace":"default","name":"vllm","uid":"ea254ec8-176f-446a-95e8-5daceee25c3f","apiVersion":"leaderworkerset.x-k8s.io/v1","resourceVersion":"358280748"}, "reason": "CreatingRevision"}
[...]

PR #594, opened 2025-07-31
Default Partition in RollingUpdateConfiguration to 0 to align with spec

ardaguclu:
#### What type of PR is this?
/kind bug

With this PR https://github.com/kubernetes-sigs/lws/pull/563, new `Partition` field is defined in `RollingUpdateConfiguration`. However, controller assumes that this field is non-empty which consequently leads to a panic.

This PR defaults Partition field in RollingUpdateConfiguration to 0 as defined in the spec https://github.com/kubernetes-sigs/lws/blob/af5d4295972e4647acbd2da8c4b2be7080893d24/api/leaderworkerset/v1/leaderworkerset_types.go#L277

Fixes https://github.com/kubernetes-sigs/lws/issues/593

```release-note
None
```

### DIFF
--- a/pkg/webhooks/leaderworkerset_webhook.go
+++ b/pkg/webhooks/leaderworkerset_webhook.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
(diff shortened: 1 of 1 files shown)


kerthcet:
@ardaguclu you need to be part of the sig member, since you're already a kubernetes org member, you can simply apply for this I think. By the way, thanks for the fix.

ardaguclu:
> @ardaguclu you need to be part of the sig member, since you're already a kubernetes org member, you can simply apply for this I think. By the way, thanks for the fix.

I think, @lixin010 is not a kubernetes-sigs member. So that github warns us.

kerthcet:
Ah, yes. Forget it.

kerthcet (commented):
/lgtm
/approve

lixin010:
Why the filed is nil? Only the lws manager was updated, but the crd was not?

PR #611, opened 2025-08-07
Fix Volcano instaillation guide/ Add gang scheduling configuration in `/config` dir / Add unit test

JesseStutler:
#### What type of PR is this?
/kind documentation
/kind feature

1. Fix Volcano installation guide, The current README does not include the steps for configuring the Volcano clusterrole if users use kubectl to install lws.
2. Add gang scheduling configuration in /config dir, following the recommendation as @ardaguclu https://github.com/kubernetes-sigs/lws/pull/498#issuecomment-3160134320
3. Add unit test for CalculatePGMinResources, following the comment as @kerthcet https://github.com/kubernetes-sigs/lws/pull/498#discussion_r2251521649

NONE

```release-note
NONE
```

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -118,3 +118,13 @@ rules:
   - get
   - patch
   - update
+# If you want to enable gang scheduling with volcano, you need to uncomment the following section
+# - apiGroups:
+#   - scheduling.volcano.sh
+#   resources:
+#   - podgroups
+#   verbs:
+#   - create
+#   - get
+#   - list
+#   - watch
--- a/config/manager/controller_manager_config.yaml
+++ b/config/manager/controller_manager_config.yaml
@@ -4,3 +4,6 @@ leaderElection:
   leaderElect: true
 internalCertManagement:
   enable: true
+# If you want to enable gang scheduling, you need to uncomment the following section(take volcano as an example)
+# gangSchedulingManagement:
+#   schedulerProvider: volcano
--- a/docs/examples/sample/gang-scheduling/README.md
+++ b/docs/examples/sample/gang-scheduling/README.md
@@ -32,7 +32,22 @@ To enable gang scheduling, you must enable the feature flag and specify `volcano
 
   Refer to the [install-by-kubectl](https://lws.sigs.k8s.io/docs/installation/#install-by-kubectl). Gang scheduling is **disabled by default**. To enable gang scheduling capabilities, you need to:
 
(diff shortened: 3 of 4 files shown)


LuyuZhang00:
Gang-scheduling is an important feature. It would be better if you could update the examples in `site/content/en/docs/`.

JesseStutler:
> Gang-scheduling is an important feature. It would be better if you could update the examples in `site/content/en/docs/`.

@LuyuZhang00 Yes I will also add content to site in later PRs, otherwise this PR may contain too many seperate contents

config/rbac/role.yaml
```diff
@@ -118,3 +118,13 @@ rules:
   - get
   - patch
   - update
+# If you want to enable gang scheduling with volcano, you need to uncomment the following section
```
ardaguclu: In my opinion, this role can be located under its dedicated file in https://github.com/kubernetes-sigs/lws/tree/main/config/rbac. User just needs to enable it in here https://github.com/kubernetes-sigs/lws/blob/main/config/rbac/kustomization.yaml (by default it can be commented out). This would align more kustomize way.

JesseStutler: Sure, thanks for your suggestion. I have modified now, please check

config/rbac/volcano_rbac.yaml
```diff
@@ -0,0 +1,30 @@
+# This file provides RBAC permissions for volcano gang scheduling.
+# To enable volcano gang scheduling, uncomment this file in kustomization.yaml
+# and ensure gangSchedulingManagement.schedulerProvider is set to "volcano" in controller configuration.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: lws-volcano-role
```
yankay, line 7: HI @JesseStutler

If we can avoid creating a new ClusterRole lws-volcano-role here and instead reuse an existing ClusterRole, it would be more consistent with Helm.
Then we can use the patches mechanism in config/default/kustomization.yaml.

Would this be better?

JesseStutler: https://github.com/kubernetes-sigs/lws/pull/611#discussion_r2266067847, at first I add the volcano related cluster role in the config/rbac/role.yaml but @ardaguclu suggested that I can seperate a lws-volcano-role so I followed it. So what should I do now 😂

ardaguclu: If we update role.yaml, in any case this role will be added. But this functionality is optional. So that in my opinion it is better to have it separately.

PR #618, opened 2025-08-15
Validate partition field be non-negative

### LINKED ISSUE #615: The partition number should not bigger than the replicas setting
**What happened**:
When I set the partition bigger than the replicas, no warning , no prompt message , should we add validation to check the value of partition field.

  replicas: 6
  rolloutStrategy:
    rollingUpdateConfiguration:
      maxSurge: 0
      maxUnavailable: 2
      partition: 10
    type: RollingUpdate

Add validation to check the value of partition field.

always

**Environment**:

- Kubernetes version (use `kubectl version`): v1.33.3
- LWS version (use `git describe --tags --dirty --always`): 0.7
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

LuyuZhang00:
When  user sets partition while creating a lws, they will receive the following warnings/errors:
•​​partition < 0​​: Error; creation is rejected.
partition and maxsurge will be ignored during creation

Fixes #615 

NONE


LuyuZhang00:
cc @ardaguclu @kerthcet

pkg/webhooks/leaderworkerset_webhook_test.go
```diff
@@ -184,3 +188,124 @@ func TestIsNotMoreThan100Percent(t *testing.T) {
 		})
 	}
 }
+
+func TestPartitionValidation(t *testing.T) {
```
Edwinhr716: We use integration tests to test that validation is working as intended, see https://github.com/kubernetes-sigs/lws/blob/main/test/integration/webhooks/leaderworkerset_test.go#L184. Please write an test there instead.

Edwinhr716:
Thanks for the quick fix!

pkg/webhooks/leaderworkerset_webhook_test.go
LuyuZhang00: thanks, I have updated it! @Edwinhr716

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -151,6 +174,19 @@ func (r *LeaderWorkerSetWebhook) generalValidate(obj runtime.Object) field.Error
...
+		// Validate partition value
+		partition := lws.Spec.RolloutStrategy.RollingUpdateConfiguration.Partition
+		partitionPath := specPath.Child("rolloutStrategy", "rollingUpdateConfiguration", "partition")
+		if partition != nil {
+			if *partition < 0 {
+				allErrs = append(allErrs, field.Invalid(partitionPath, *partition, "partition must be greater than or equal to 0"))
+			}
+			if *partition > *lws.Spec.Replicas {
+				allErrs = append(allErrs, field.Invalid(partitionPath, *partition,
+					fmt.Sprintf("partition must be less than or equal to replicas %d", *lws.Spec.Replicas)))
```
yankay: HI @LuyuZhang00 

By analogy, in a StatefulSet, it is acceptable to set it this way: "If a StatefulSet's .spec.updateStrategy.rollingUpdate.partition is greater than its .spec.replicas, updates to its .spec.template will not be propagated to its Pods." (Reference: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#partitions)

Therefore, "partition greater than replicas" should not be considered a bug.
How do you think about that?

LuyuZhang00: @yankay thanks! To maintain consistency with the StatefulSet, I have refined it. Please check!

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -94,8 +94,33 @@ var _ webhook.CustomValidator = &LeaderWorkerSetWebhook{}
...
+	lws := obj.(*v1.LeaderWorkerSet)
+	warnings := admission.Warnings{}
+
+	if lws.Spec.RolloutStrategy.RollingUpdateConfiguration != nil {
+		partition := lws.Spec.RolloutStrategy.RollingUpdateConfiguration.Partition
+		maxSurge := lws.Spec.RolloutStrategy.RollingUpdateConfiguration.MaxSurge
+		replicas := *lws.Spec.Replicas
+
+		// Warn users that partition/maxSurge will be ignored during initial deployment
+		if partition != nil && *partition > 0 {
```
kerthcet: Should we add this to warning? I think warning is something 1) when we got unexpected behaviors, 2) or potential performance problems 3) or some unrecommended usages. Here this is working as expected.

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -228,7 +261,7 @@ func getPercentValue(intOrStringValue intstr.IntOrString) (int, bool) {
 func validateNonnegativeField(value int64, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 	if value < 0 {
-		allErrs = append(allErrs, field.Invalid(fldPath, value, "must be grater than or equal to 0"))
+		allErrs = append(allErrs, field.Invalid(fldPath, value, "must be greater than or equal to 0"))
```
kerthcet, line 239: Good catch.

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -159,6 +184,14 @@ func (r *LeaderWorkerSetWebhook) generalValidate(obj runtime.Object) field.Error
 		allErrs = append(allErrs, validatePositiveIntOrPercent(maxSurge, maxSurgePath)...)
 		allErrs = append(allErrs, isNotMoreThan100Percent(maxSurge, maxSurgePath)...)
 	}
+
+	// Validate partition value
+	partition := lws.Spec.RolloutStrategy.RollingUpdateConfiguration.Partition
+	partitionPath := specPath.Child("rolloutStrategy", "rollingUpdateConfiguration", "partition")
+	if lws.Spec.RolloutStrategy.RollingUpdateConfiguration != nil {
```
kerthcet, line 163: If this is nil, L189 will panic, move the code to this block please.

pkg/webhooks/leaderworkerset_webhook.go
kerthcet: Does sts or deployment has this warning? If not, let's remove it.

kerthcet (commented):
/lgtm
/approve

Thanks @LuyuZhang00

PR #623, opened 2025-08-25
fix: 🐛 Declare the variable CGO_ENABLED in Dockerfile

yafengio:
#### What type of PR is this?

/kind bug

Declare the variable `CGO_ENABLED` in Dockerfile.


Edwinhr716:
Why is this needed?

kerthcet (commented):
/approve
LGTM to me since we will visit the CGO_ENABLED here.

Issue #630, opened 2025-08-27
An unnecessary podgroup has been created


zhouzhifeisdcn:
I install lws with parameter：
	VERSION=v0.7.0
helm install lws https://github.com/kubernetes-sigs/lws/releases/download/$VERSION/lws-chart-$VERSION.tgz \
  --namespace lws-system \
  --create-namespace \ 
  --set gangSchedulingManagement.schedulerProvider=volcano  \
  --wait --timeout 300s

I create a lws.yaml but  did not specify the scheduler as Volcano；
```·yaml
kind: LeaderWorkerSet
metadata:
  name: leaderworkerset-multi-template
spec:
  replicas: 1
  leaderWorkerTemplate:
    leaderTemplate:
      spec:
        containers:
        - name: nginx2
          image: nginxinc/nginx-unprivileged:1.27
          resources:
            limits:
              cpu: "100m"
            requests:
              cpu: "50m"
          ports:
          - containerPort: 8080
    size: 4
    workerTemplate:
      spec:
        containers:
        - name: nginx
          image: nginxinc/nginx-unprivileged:1.27
          resources:
            limits:
              cpu: "100m"
            requests:
              cpu: "50m"
          ports:
          - containerPort: 8080

```
LWS created a podgroup, but the status of the podgroup remains Inqueue; This means that it has never been scheduled by Volcano

<img width="1108" height="111" alt="Image" src="https://github.com/user-attachments/assets/7b540284-b8a4-46d3-b03e-627ae62beee7" />

ahg-g:
Has this been released? if so, we should fix this and patch previous releases that included pod group support

ardaguclu:
Does this happen if you don't set `--set gangSchedulingManagement.schedulerProvider=volcano` in your helm chart?

yankay:
I'm trying to reproduce this issue. The PodGroup was also created successfully.

Hi @JesseStutler,

Could you please clarify:

When `gangSchedulingManagement.schedulerProvider=volcano` is set but `schedulerName: volcano` is not specified, is it expected behavior for the PodGroup to be created or not?

Is this a bug or an intentional design?

JesseStutler:
> I'm trying to reproduce this issue. The PodGroup was also created successfully.
> 
> Hi [@JesseStutler](https://github.com/JesseStutler),
> 
> Could you please clarify:
> 
[...]

PR #632, opened 2025-08-27
Do not automatically create PG,when pod not use schedule volcano 

zhouzhifeisdcn:
Do not automatically create PG,when pod not use schedule volcano 
fixed issue https://github.com/kubernetes-sigs/lws/issues/630

### DIFF
--- a/pkg/controllers/pod_controller.go
+++ b/pkg/controllers/pod_controller.go
@@ -126,7 +126,7 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 		return ctrl.Result{}, nil
 	}
 
-	if r.SchedulerProvider != nil {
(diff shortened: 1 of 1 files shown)


ahg-g:
can we please add tests to cover this?

pkg/controllers/pod_controller.go
```diff
@@ -126,7 +126,7 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 		return ctrl.Result{}, nil
 	}
 
-	if r.SchedulerProvider != nil {
+	if r.SchedulerProvider != nil && strings.Contains(pod.Spec.SchedulerName, string(schedulerprovider.Volcano)) {
```
JesseStutler, line 129: I don't think it's a good choice to hard-code it here, I think the main reason in your issue is that, volcano will automatically create podgroups for orphan pods, but you didn't specify `schedulerName: volcano` in your lws template. I think we'd better automatically inject `schedulerName: volcano` for lws needs to use volcano as the gang scheduler

PR #634, opened 2025-08-29
fix: add nodeSelector on Linux to avoid crash on Windows node

andyzhangx:
#### What type of PR is this?
/kind bug

fix: add nodeSelector on Linux to avoid crash on Windows node

Fixes #

```release-note
fix: add nodeSelector on Linux to avoid crash on Windows node
```

### DIFF
--- a/charts/lws/values.yaml
+++ b/charts/lws/values.yaml
@@ -45,7 +45,8 @@ resources:
     # requests:
     #   cpu: 100m
     #   memory: 128Mi
-nodeSelector: {}
+nodeSelector:
+  kubernetes.io/os: linux
 tolerations: []
 affinity: {}
 # gangSchedulingManagement is configuration for gang scheduling management.


charts/lws/values.yaml
```diff
@@ -45,7 +45,8 @@ resources:
     # requests:
     #   cpu: 100m
     #   memory: 128Mi
-nodeSelector: {}
+nodeSelector:
```
ardaguclu, line 48: Does kustomize include this selector?

andyzhangx: I have searched in this repo, it does not contain `kubernetes.io/os: linux` nodeSelector, I think we should use this nodeSelector value by default, and user could specify other values if necessary.

ahg-g: you can use tolerations (we already have that below), basically setting that by default this doesn't tolerate running on windows machines.

Issue #622, opened 2025-08-21
Support per-pod PV (similar to StatefulSet volumeClaimTemplates)


zhuangqh:
**What would you like to be added**:

We found LWS very suitable for deploying inference clusters. However, in our scenario, we place model data on local NVMe disks. These disks are exposed as Local PVs, and we typically rely on volumeClaimTemplates in a StatefulSet to manage per-pod PVs.

Currently, LWS does not support something similar to volumeClaimTemplates. Since LWS is built on top of StatefulSet as its underlying workload, would it be possible to consider supporting per-pod PV management in a similar way?

**Why is this needed**:

This feature would allow each pod in LWS to have its own dedicated local PV, making it much easier to run workloads that depend on fast local storage (e.g., large model inference).

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

Fei-Guo:
@ahg-g, would you please comment if it is possible to add API changes for this? It is currently a key blocker for us to onboard LWS. Thanks.

yankay:
Thank you, that's a great suggestion.  

May I ask, if this data is "cache" rather than "persistent," would using an emptyDir solve the problem?

zhuangqh:
> Thank you, that's a great suggestion.
> 
> May I ask, if this data is "cache" rather than "persistent," would using an emptyDir solve the problem?

The disk I/O performance of emptyDir is too poor, so we use a CSI plugin to leverage local NVMe disks. Here, “cache” mainly refers to placing the model files on the local disk.

andyzhangx:
if it's emptyDir, then data would be wiped when pod is deleted, while in our scenario, we download model data to the pv, and the model data is quite big, we don't want to download again every time when new pod is started.

yankay:
I can also think of a scenario involving KVCache Local Storage, https://docs.lmcache.ai/kv_cache/storage_backends/local_storage.html. I'm not sure if it's also applicable to this feature.

ahg-g:
I am supportive of this feature, please feel free to submit a concrete proposal. /cc @Edwinhr716 since we discussed this in the past.

kerthcet:
Thanks for reaching @zhuangqh Also +1 for this, glad to see this be part of release v0.8.0, some concerns here:
- should the leader has the same claim template as the worker, or they're dedicated
[...]

PR #636, opened 2025-09-01
KEP-622: VolumeClaimTemplates support for Leader and Worker StatefulSets

andyzhangx:
#### What type of PR is this?
/kind documentation

KEP-622: VolumeClaimTemplates support for Leader and Worker StatefulSets

KEP for the fix for https://github.com/kubernetes-sigs/lws/issues/622

Fixes #

```release-note
none
```


keps/622-volume-claim-templates/README.md
```diff
@@ -0,0 +1,295 @@
...
+	// Every claim in this list must have at least one matching (by name) volumeMount
+	// in one container in the template. A claim in this list takes precedence over
+	// any volumes in the template, with the same name.
+	// +optional
+	VolumeClaimTemplates []corev1.PersistentVolumeClaim `json:"volumeClaimTemplates,omitempty"`
+
+	// PersistentVolumeClaimRetentionPolicy describes the policy used for PVCs created from
+	// the StatefulSet VolumeClaimTemplates.
+	// +optional
+	PersistentVolumeClaimRetentionPolicy *appsv1.StatefulSetPersistentVolumeClaimRetentionPolicy `json:"persistentVolumeClaimRetentionPolicy,omitempty"`
```
Edwinhr716, line 170: Can you add somewhere what the default values are in the case where there is a `ValueClaimTemplates` but no value for `PersistentVolumeClaimRetentionPolicy`

andyzhangx: added in https://github.com/kubernetes-sigs/lws/pull/636/commits/002966c1102939a5737fe468d21e79b8a8a1d001:

> If the `PersistentVolumeClaimRetentionPolicy` is not set, its default value will be the same as the default values of the statefulset (as shown below).:
> ```
>   persistentVolumeClaimRetentionPolicy:
>     whenDeleted: Retain
>     whenScaled: Retain
> ```

andyzhangx: @Edwinhr716 any comment on this design doc PR? thx

Edwinhr716:
The implementation side makes sense to me, we can just rely on the StatefulSet controller. But I'm wondering if there is an API definition that makes it easier to add support for VCT isolation between leaders and workers in the future.

andyzhangx:
> VCT isolation

@Edwinhr716 what does `VCT isolation` mean? do you have more context? thanks.

Edwinhr716:
Sorry, what I meant was a VCT for only the leaders, and a VCT for the workers

andyzhangx:
> Sorry, what I meant was a VCT for only the leaders, and a VCT for the workers

we could add `VolumeClaimTemplatesForLeader`, `VolumeClaimTemplatesForWorker` if that's necessary in the future.

PR #633, opened 2025-08-28
[WIP] feat: add volumeClaimTemplates in LeaderWorkerSetSpec

andyzhangx:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it
this PR adds volumeClaimTemplates and PersistentVolumeClaimRetentionPolicy (both are optional) in LeaderWorkerSetSpec, since these two fields are optional, that would be compatible with existing LeaderWorkerSetSpec config in previous version

#### Which issue(s) this PR fixes

Fixes #622

#### Does this PR introduce a user-facing change?

```release-note
feat: add volumeClaimTemplates in LeaderWorkerSetSpec
```

PR #655, opened 2025-09-17
Run make generate and make manifests to update CRD

ardaguclu:
#### What type of PR is this?
/kind bug

It looks like `make manifests` and `make generate` were missed to call. So that these targets create unexpected diff. This PR updates the code base to synchronize manifests with the latest changes in the code.

Probably originated from https://github.com/kubernetes-sigs/lws/pull/633

```release-note
None
```


andyzhangx (approved):
/lgtm

PR #663, opened 2025-09-22
docs: add hpa docs

### LINKED ISSUE #652: [Documentation] Add HPA example to LWS site
Add a section to the lws site showcasing HPA. Example in codebase is here https://github.com/kubernetes-sigs/lws/blob/main/docs/examples/sample/horizontal-pod-autoscaler.yaml, we can potentially add it in the examples section https://lws.sigs.k8s.io/docs/examples/. Code side of that is here https://github.com/kubernetes-sigs/lws/tree/main/site/content/en/docs/examples

LuyuZhang00:
#### What type of PR is this?
/kind documentation

add hpa docs

Fixes #652 

No

```release-note

```

### DIFF
--- a/site/content/en/docs/examples/hpa.md
+++ b/site/content/en/docs/examples/hpa.md
@@ -1,0 +1,384 @@
+---
+title: "Horizontal Pod Autoscaler (HPA)"
+linkTitle: "HPA"
+weight: 5
+description: >
+  An example of using Horizontal Pod Autoscaler with LeaderWorkerSet
+---
+
+LeaderWorkerSet supports Horizontal Pod Autoscaler (HPA) through its scale subresource. This allows you to automatically scale the number of replica groups based on resource utilization metrics like CPU or memory.
+
+## How HPA Works with LeaderWorkerSet
+
+When using HPA with LeaderWorkerSet:
+
+- HPA monitors **leader pods** only (not worker pods)
+- The `hpaPodSelector` in LeaderWorkerSet status helps HPA identify which pods to monitor
+- Scaling affects the number of **replica groups**, not individual pods within a group
+- Each replica group (leader + workers) is scaled as a unit
+
+## Prerequisites
+
+Before setting up HPA, ensure you have:
+
+### 1. Install Metrics Server
+
+HPA requires metrics-server to be running in your cluster to collect CPU and memory metrics. Here are the detailed installation steps:
+
+#### Step 1: Download metrics-server configuration file
+
+```shell
+wget https://github.com/kubernetes-sigs/metrics-server/releases/latest/download/components.yaml -O metrics-server.yaml
+```
+
+#### Step 2: Modify configuration file
+
+metrics-server requires certificate validation to access kubelet, but Kind clusters use self-signed certificates. Edit the `metrics-server.yaml` file and add the `--kubelet-insecure-tls` parameter to the deployment's `args` section:
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    k8s-app: metrics-server
+  name: metrics-server
+  namespace: kube-system
+spec:
+  # ... other configurations remain unchanged
+  template:
+    spec:
+      containers:
+      - args:
+        - --cert-dir=/tmp
+        - --secure-port=4443
+        - --kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname
+        - --kubelet-use-node-status-port
+        - --metric-resolution=15s
+        - --kubelet-insecure-tls  # Add this parameter to skip certificate verification
+        image: registry.k8s.io/metrics-server/metrics-server:v0.8.0
+        # If you encounter image download issues, you can use domestic mirrors:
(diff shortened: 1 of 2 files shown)


site/content/en/docs/examples/hpa.md
```diff
@@ -0,0 +1,384 @@
...
+- HPA monitors **leader pods** only (not worker pods)
+- The `hpaPodSelector` in LeaderWorkerSet status helps HPA identify which pods to monitor
+- Scaling affects the number of **replica groups**, not individual pods within a group
+- Each replica group (leader + workers) is scaled as a unit
+
+## Prerequisites
+
+Before setting up HPA, ensure you have:
+
+### 1. Install Metrics Server
```
Edwinhr716, line 24: Is metrics server provided by most cloud providers? I've never had to do this step.

site/content/en/docs/examples/hpa.md
```diff
@@ -0,0 +1,384 @@
...
+  An example of using Horizontal Pod Autoscaler with LeaderWorkerSet
+---
+
+LeaderWorkerSet supports Horizontal Pod Autoscaler (HPA) through its scale subresource. This allows you to automatically scale the number of replica groups based on resource utilization metrics like CPU or memory.
+
+## How HPA Works with LeaderWorkerSet
+
+When using HPA with LeaderWorkerSet:
+
+- HPA monitors **leader pods** only (not worker pods)
```
Edwinhr716, line 15: Would be good to add that model server metrics often take into account the whole group

site/content/en/docs/examples/hpa.md
LuyuZhang00: This is a note, if user already have installed, they can safely ignore this step, I  included it for my initial test

​​For open-source Kubernetes:​
By default, installations like kubeadm, kind, or minikube don't include the Metrics Server. The quickest way to get started with kubectl topor HPA is to install it manually.

For managed Kubernetes services (like GKE, EKS, AKS, etc.):​​
Most cloud providers include Metrics Server out of the box in their managed offerings—so it’s usually already there and good to go for things like HPA and kubectl top.

site/content/en/docs/examples/hpa.md
LuyuZhang00: I tested adding CPU load to both the leader pod and the worker pods separately. In practice, the HPA only took effect when CPU load was increased on the leader pod. I noticed that in the updateStatus function within pkg/controllers/leaderworkerset_controller.go, it also monitors the leader pod. This seems like a potential area for improvement.

site/content/en/docs/examples/hpa.md
Edwinhr716: Makes sense, I'm fine with keeping this section. I think we should have a note saying: if using a managed Kubernetes service, skip this step

site/content/en/docs/examples/hpa.md
Edwinhr716: Yes, when using metrics exposed by the nodes themselves, it will only take the leader into account. My point was that we should add that model servers, such as vLLM, expose metrics that take into account the whole group.

> This seems like a potential area for improvement.

I agree, but it is a complex area. We would need a way to aggregate the metrics of a group of pods, which should be covered by the revised Gang Scheduling KEP https://docs.google.com/document/d/1ulO5eUnAsBWzqJdk_o5L-qdq5DIVwGcE7gWzCQ80SCM/edit?pli=1&tab=t.0. 

An idea I was thinking of earlier was to be able to change whether to target a leader pod or a worker pod for HPA, since there are deployment patterns where the leader only has a proxy server, so it isn't representative of when the accelerators being used have reached its limit.

site/content/en/docs/examples/hpa.md
```diff
@@ -0,0 +1,384 @@
...
+1. CPU usage decreases to 0%
+2. HPA waits for the stabilization window before starting to scale down
+3. LeaderWorkerSet replicas gradually decrease to 2
+
+## Advanced HPA Configuration
+
+### Memory-based Scaling
+
+You can also configure HPA to scale based on memory utilization:
+
```
yankay, line 292: HI @LuyuZhang00 

Considering other examples, their YAML files are placed in the https://github.com/kubernetes-sigs/lws/tree/main/docs/examples/sample directory rather than within the markdown files.

Should we create an HPA directory under docs/examples and place the YAML files there? This would make maintenance easier.

kerthcet:
I think this is good to merge, would you mind to rebase the main branch and push again, because I found we have something similar in the example page. @LuyuZhang00 Thanks a lot.

LuyuZhang00:
After rebase, it seems nothing changed. Do you mean only keeping the content from `site/content/en/docs/examples/hpa.md`?
@kerthcet

kerthcet:
The error seems like a flaky, not related. Are u happy with this @Edwinhr716  as well?

Edwinhr716:
> The error seems like a flaky, not related

Test is failing since it needs to include this PR #741

> Are u happy with this @Edwinhr716 as well?

If the comments are addressed, yes

kerthcet:
Pretty good to me, so
/lgtm

ahg-g:
Force merging

PR #705, opened 2025-11-19
[bugfix] fix StatefulsetReady check

LuyuZhang00:
#### What type of PR is this?
/kind bug

We noticed that when the pods in the worker StatefulSet (sts) have been created but are not yet ready, the sts is still considered "ready", which doesn't match the expected behavior. We found that the sts has three parameters used to determine its readiness status(/go/pkg/mod/k8s.io/api@v0.34.2/apps/v1/types.go) :
- `Replicas`: The number of Pods created by the StatefulSet controller. This includes pods in all states (Running, Pending, Unknown, etc.) and does not consider whether the pods are ready or available.
- `ReadyReplicas`: The number of pods created for this StatefulSet with a Ready Condition. This requires the pods to be in the Running state and for all their containers to have passed the readiness probe.
- `AvailableReplicas`: Total number of available pods (ready for at least minReadySeconds) targeted by this statefulset. In addition to being "ready" (as defined by ReadyReplicas), the pods must have been running for at least the duration specified by minReadySeconds. The minReadySeconds defaults to 0s, which makes it behave the same as ReadyReplicas. However, when a specific value is set, it provides the Pod with a sufficient stable running time

 Based on the current situation, using AvailableReplicas may be a better choice to determine the ready status of a StatefulSet
cc @Edwinhr716 @kerthcet

### DIFF
--- a/pkg/utils/statefulset/statefulset_utils.go
+++ b/pkg/utils/statefulset/statefulset_utils.go
@@ -46,6 +46,6 @@ func GetParentNameAndOrdinal(name string) (string, int) {
 
 // StatefulsetReady checks whether a sts is ready.
 func StatefulsetReady(sts appsv1.StatefulSet) bool {
-	return *sts.Spec.Replicas == sts.Status.Replicas &&
+	return *sts.Spec.Replicas == sts.Status.AvailableReplicas &&
 		sts.Status.CurrentRevision == sts.Status.UpdateRevision
 }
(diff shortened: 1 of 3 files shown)


kannon92:
Why would you not choose AvailableReplicas? That seems even better than ReadyReplicas if MinReady is set.

LuyuZhang00:
> Why would you not choose AvailableReplicas? That seems even better than ReadyReplicas if MinReady is set.

yes, AvailableReplicasis better. We've been using that internally too. Initially, I set it to ReadyReplicas thinking its conditions were more relaxed and a better fit for open-source scenarios. Now I've changed it to AvailableReplicas as well.

Edwinhr716:
Nice catch, makes sense to use AvailableReplicas instead of Replicas. 

/lgtm
/approve

Issue #695, opened 2025-11-07
LWS helm uninstall will delete CRDs


ifelseend:
**What happened**:
LWS helm uninstall will delete lws CRDs

**What you expected to happen**:
https://helm.sh/docs/chart_best_practices/custom_resource_definitions/
LWS helm uninstall will keep lws CRDs

**How to reproduce it (as minimally and precisely as possible)**:
LWS helm uninstall
**Anything else we need to know?**:

**Environment**:

- Kubernetes version (use `kubectl version`):
- LWS version (use `git describe --tags --dirty --always`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Mostafahassen1:
Hi @Edwinhr716 , @ifelseend   is this issue still open so I can work on it? And is there any documentation to guide me on how to clone the project? I am a new contributor. I already read some parts of the code and documentation, and now I understand a helicopter view of the project.

Edwinhr716:
Yes, the issue is still open, feel free to start working on it. To clone it, you should create a fork of the lws repo, then create a branch in the fork, and create PR to merge the branch into the LWS repo once the change is ready for review

Edwinhr716:
/close
PR for this has been merged, will close

PR #718, opened 2025-12-14
Try to fix Helm so it does not allow deleting CRDs

Mostafahassen1:
**Issue =>**  https://github.com/kubernetes-sigs/lws/issues/695

**The issue is not fixed yet. I am trying to fix it, but I may be going in the wrong direction. I need help to find the correct way to resolve this issue.**

**I used the following commands to run the application:**

```bash
eval $(minikube docker-env)
```

```bash
docker build -t my-lws:dev .
```

```bash
CHART_VERSION=0.6.1
```

```bash
helm upgrade --install lws oci://registry.k8s.io/lws/charts/lws \
  --version=$CHART_VERSION \
  --namespace lws-system \
  --create-namespace \
  --set manager.image.repository=my-lws \
  --set manager.image.tag=dev \
  --set manager.image.pullPolicy=Never \
  --wait --timeout 300s
```

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -392,4 +392,4 @@ hugo:
 crds: kustomize yq # update helm CRD files
 	$(KUSTOMIZE) build config/default \
 	| $(YQ) 'select(.kind == "CustomResourceDefinition")' \
-	> charts/lws/templates/crds/leaderworkerset.x-k8s.io_leaderworkersets.yaml
+	> charts/lws/crds/leaderworkerset.x-k8s.io_leaderworkersets.yaml


Mostafahassen1:
Hi @Edwinhr716, please review this when you have time.

Mostafahassen1:
> I think the problem is because crds is placed under the template folder, but in practice, we should make them in parallel.

Hi @kerthcet, I worked on the issue again by moving the crds out of the templates/ directory. I then tested the app to check whether Helm deletes the CRDs, but I found that it still deletes them.

image


image

kerthcet:
image
I tested locally, can you verify again? We're talking about CRD here.

Mostafahassen1:
Hi @kerthcet 

I pushed the code after makeing the crds and templates folders separately,
and removed `helm.sh/resource-policy: keep`.
I added this in commit number 1.

The latest work is in commit number 2.

kerthcet:
Thanks, will take a look later.

kerthcet:
/lgtm
/approve

Thanks @Mostafahassen1

kerthcet:
Seems we still need to change the Makefile, see command: `make crds`.

Mostafahassen1:
Hi @kerthcet, I fixed it

Issue #706, opened 2025-11-20
Inject TPU_PROCESS_ADDRESSES, TPU_PROCESS_PORT environment variable


Edwinhr716:
**What would you like to be added**:
 I would like LWS to inject the variable `TPU_PROCESS_ADDRESSES` and `TPU_PROCESS_PORT` just like it does `TPU_WORKER_HOSTNAMES`

**Why is this needed**:
New TPU machine types require these environment variables to be set.

**Completion requirements**:
This feature will be done in two stages

Stage 1: TPU_PROCESS_ADDRESSES will inject the same values that TPU_WORKER_HOSTNAMES, but with the added port 8746 appended to every hostname

```
hostnamesWithHost = append(hostnames, fmt.Sprintf("%s-%d.%s:8476", leaderName, i, pod.Spec.Subdomain))
```

And add TPU_PROCESS_PORT=8476

Stage 2: TPU_PROCESS_ADDRESSES will support a new port per each container running in the pod, using ports 8476-8479

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

adinilfeld:
A couple of use cases that we still don't support:
- Adjusting the TPU environment variables based on subgroups (https://github.com/kubernetes-sigs/lws/tree/main/keps/115-Subgroup-support#tpu-environment-variable-injection). #754 added the injection logic to [AddTPUVariables](https://github.com/adinilfeld/lws/blob/beb25cb7e9905f66e3d0208090b934e6db5e98c3/pkg/utils/accelerators/tpu.go#L201), but [addTPUVariablesSubGroup](https://github.com/adinilfeld/lws/blob/beb25cb7e9905f66e3d0208090b934e6db5e98c3/pkg/utils/accelerators/tpu.go#L99) wasn't updated.
- We currently assume that the leader and worker pods contain the same number of containers ([src](https://github.com/adinilfeld/lws/blob/beb25cb7e9905f66e3d0208090b934e6db5e98c3/pkg/utils/accelerators/tpu.go#L261-L267)). Implementing the logic for different numbers of containers would be non-trivial, since `AddTPUVariables` only has access to the spec of the pod being admitted (from which `numContainers` is determined), not both the leader and worker specs.

PR #722, opened 2025-12-18
Injecting environment variables for new TPU machine

Edwinhr716:
#### What type of PR is this?
/kind feature

Injects new environment variables that are needed to run workloads on new TPU machines, part of #706 

Fixes #

```release-note

```

### DIFF
--- a/pkg/utils/accelerators/tpu.go
+++ b/pkg/utils/accelerators/tpu.go
@@ -34,7 +34,7 @@ const (
 	TpuWorkerHostNames              string              = "TPU_WORKER_HOSTNAMES"
 	TpuProcessAddresses             string              = "TPU_PROCESS_ADDRESSES"
 	TpuProcessPortName              string              = "TPU_PROCESS_PORT"
-	TpuProcessDefaultPort                  string              = "8476"
+	TpuProcessDefaultPort           string              = "8476"
 	TpuWorkerId                     string              = "TPU_WORKER_ID"
 	TpuName                         string              = "TPU_NAME"
 	LeaderRequestsTPUsAnnotationKey string              = "leaderworkerset.sigs.k8s.io/leader-requests-tpus"
@@ -122,7 +122,7 @@ func addTPUVariablesSubGroup(pod *corev1.Pod) error {
 	tpuProcessPortInContainer, tpuProcessPort := podutils.GetEnvVarValueIfInContainer(container, TpuProcessPortName)
 	if !tpuProcessPortInContainer {
 		// If user doesn't specify tpuProcessPort, we fall back to hardcoded value
-		tpuProcessPort = TpuProcessPort
+		tpuProcessPort = TpuProcessDefaultPort
 	}
 
 	start := subGroupSize*subGroupIndex + 1
@@ -207,7 +207,7 @@ func AddTPUVariables(pod *corev1.Pod, size int) error {
 	tpuProcessPortInContainer, tpuProcessPort := podutils.GetEnvVarValueIfInContainer(container, TpuProcessPortName)
 	if !tpuProcessPortInContainer {
 		// If user doesn't specify tpuProcessPort, we fall back to hardcoded value
-		tpuProcessPort = TpuProcessPort
+		tpuProcessPort = TpuProcessDefaultPort
 	}
 
 	leaderName := pod.Name
--- a/pkg/utils/accelerators/tpu_test.go
+++ b/pkg/utils/accelerators/tpu_test.go
@@ -140,7 +140,7 @@ func TestAddTPUVariables(t *testing.T) {
 						if diff := cmp.Diff(envVar.Value, tc.expectedTpuProcessAddresses); diff != "" {
 							t.Errorf("unexpected add TPU worker ID operation: %s", diff)
 						}
-					case TpuProcessPort:
+					case TpuProcessPortName:
 						if diff := cmp.Diff(envVar.Value, tc.expectedTpuProcessPort); diff != "" {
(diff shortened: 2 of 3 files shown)


pkg/utils/accelerators/tpu.go
```diff
@@ -29,11 +29,17 @@ import (
...
-	TpuName                         string              = "TPU_NAME"
-	LeaderRequestsTPUsAnnotationKey string              = "leaderworkerset.sigs.k8s.io/leader-requests-tpus"
+	TpuResourceName     corev1.ResourceName = corev1.ResourceName("google.com/tpu")
+	TpuWorkerHostNames  string              = "TPU_WORKER_HOSTNAMES"
+	TpuProcessAddresses string              = "TPU_PROCESS_ADDRESSES"
+	TpuProcessPortName  string              = "TPU_PROCESS_PORT"
+	TpuProcessPort      string              = "8476"
+	TpuWorkerId         string              = "TPU_WORKER_ID"
+	TpuName             string              = "TPU_NAME"
+	Tpu
```
ahg-g: what is this?

pkg/utils/accelerators/tpu.go
```diff
@@ -29,11 +29,17 @@ import (
...
+	LeaderRequestsTPUsAnnotationKey string = "leaderworkerset.sigs.k8s.io/leader-requests-tpus"
+	TpuAcceleratorLabel             string = "cloud.google.com/gke-tpu-accelerator"
```
ahg-g: same here, hardcoding the label name is not future proof; but also do we really need to set it differently for different tpu models? can we just inject the new envvars for all?

pkg/utils/accelerators/tpu.go
```diff
@@ -29,11 +29,17 @@ import (
...
-	TpuResourceName                 corev1.ResourceName = corev1.ResourceName("google.com/tpu")
-	TpuWorkerHostNames              string              = "TPU_WORKER_HOSTNAMES"
-	TpuWorkerId                     string              = "TPU_WORKER_ID"
-	TpuName                         string              = "TPU_NAME"
-	LeaderRequestsTPUsAnnotationKey string              = "leaderworkerset.sigs.k8s.io/leader-requests-tpus"
+	TpuResourceName     corev1.ResourceName = corev1.ResourceName("google.com/tpu")
+	TpuWorkerHostNames  string              = "TPU_WORKER_HOSTNAMES"
+	TpuProcessAddresses string              = "TPU_PROCESS_ADDRESSES"
+	TpuProcessPortName  string              = "TPU_PROCESS_PORT"
+	TpuProcessPort      string              = "8476"
```
ahg-g: Hardcoding the port number is strange, what if one needs to set it to something different?

pkg/utils/accelerators/tpu.go
Edwinhr716: accident, removed it.

pkg/utils/accelerators/tpu.go
Edwinhr716: Previous generations should still work, main concern was adding a new environment variable breaks backwards compatibility. As in, two LWS deployments on a previous gen TPU that look the same won't be exactly the same depending on the LWS version.

pkg/utils/accelerators/tpu.go
Edwinhr716: The same thing is done on TPU7x deployments by GKE. 

For added context. The new generation supports running multiple containers on a single TPU node. To be able to run, containers are assigned ports from 8476 to 8479 depending on the index of the container. So for single container deployments (e.g LWS), it should be assigned 8476.

ahg-g: Hardcoding port numbers is an anti-pattern, I strongly recommend defining an api to capture this intent, perhaps cap, either as an annotation or find it in the env vars (assume the user will set the port as an env var with a specific name)

pkg/utils/accelerators/tpu.go
ahg-g: it should be a no-op as long as we correctly handle not triggering unintentional rolling update  when upgrading the operator

pkg/utils/accelerators/tpu.go
```diff
@@ -25,12 +25,16 @@ import (
...
+	podutils "sigs.k8s.io/lws/pkg/utils/pod"
 	statefulsetutils "sigs.k8s.io/lws/pkg/utils/statefulset"
 )
 
 const (
 	TpuResourceName                 corev1.ResourceName = corev1.ResourceName("google.com/tpu")
 	TpuWorkerHostNames              string              = "TPU_WORKER_HOSTNAMES"
+	TpuProcessAddresses             string              = "TPU_PROCESS_ADDRESSES"
+	TpuProcessPortName              string              = "TPU_PROCESS_PORT"
+	TpuProcessPort                  string              = "8476"
```
ahg-g:
```suggestion
	TpuProcessDefaultPort                  string              = "8476"
```

ahg-g: pls document that the port will be first looked up in the env vars and if not found will be defaulted to this

pkg/utils/pod/pod_utils.go
```diff
@@ -92,6 +92,15 @@ func getPodConditionFromList(conditions []corev1.PodCondition, conditionType cor
 	return -1, nil
 }
 
+func GetEnvVarValueIfInContainer(c *corev1.Container, envVarName string) (bool, string) {
```
ahg-g, line 95: add unit tests pls

Issue #688, opened 2025-10-28
LWS does not support zero-downtime updates when replicas = 1


bcfre:
**What happened**:
When I deploy a LeaderWorkerSet (LWS) with only one replica, I want to leverage maxSurge to perform a zero-downtime rollout. However, our current implementation seems to not support this scenario.

**What you expected to happen**:
Rolling update is vital to online services with zero downtime. 

**How to reproduce it (as minimally and precisely as possible)**:
1. Deploy LWS with maxSurge = 2 and replicas = 1:
```yaml
apiVersion: leaderworkerset.x-k8s.io/v1
kind: LeaderWorkerSet
metadata:
  name: leaderworkerset-rollout
spec:
  rolloutStrategy:
    type: RollingUpdate
    rollingUpdateConfiguration:
      maxUnavailable: 1
      maxSurge: 2
  replicas: 1
  leaderWorkerTemplate:
    size: 1
    workerTemplate:
      spec:
        containers:
        - name: nginx
          image: nginxinc/nginx-unprivileged:1.27
          resources:
            limits:
              cpu: "100m"
            requests:
              cpu: "50m"
          ports:
          - containerPort: 8080
```
2. Once the LWS is ready, redeploy with an invalid image:
```yaml
apiVersion: leaderworkerset.x-k8s.io/v1
kind: LeaderWorkerSet
metadata:
  name: leaderworkerset-rollout
spec:
  rolloutStrategy:
    type: RollingUpdate
    rollingUpdateConfiguration:
      maxUnavailable: 1
      maxSurge: 2
  replicas: 1
  leaderWorkerTemplate:
    size: 1
    workerTemplate:
      spec:
        containers:
        - name: nginx
          image: nginxinc/nginx-unprivileged:1.27-not-exist
          resources:
            limits:
              cpu: "100m"
            requests:
              cpu: "50m"
          ports:
          - containerPort: 8080
```
3. There only one Pod, and that Pod in ImagePullBackOff state.
```shell
k get po -l leaderworkerset.sigs.k8s.io/name=leaderworkerset-rollout
NAME                        READY   STATUS             RESTARTS   AGE
leaderworkerset-rollout-0   0/1     ImagePullBackOff   0          48s
```

**Anything else we need to know?**:
I suspect this issue is related to the following code snippet:
1. In this section, the code prematurely scales down the expected maxSurge in order to gradually remove surge replicas. This causes finalReplicas to go back to lws.replica = 1.
```go
[...]

PR #724, opened 2026-01-03
LWS does not support zero-downtime updates when replicas = 1

Mostafahassen1:
Issue > https://github.com/kubernetes-sigs/lws/issues/688#issuecomment-3706895557

I think LWS now supports zero-downtime updates when replicas = 1

This is the output when running the correct image:

Image

And this is the output when building the wrong image:

Image
  

Image


pkg/controllers/leaderworkerset_controller.go
```diff
@@ -295,8 +295,8 @@ func (r *LeaderWorkerSetReconciler) rollingUpdateParameters(ctx context.Context,
 	// Case 2:
 	// Indicates a new rolling update here.
 	if leaderWorkerSetUpdated {
-		// Processing scaling up/down first prior to rolling update.
-		return min(lwsReplicas, stsReplicas), wantReplicas(lwsReplicas), nil
+		protectedPartition := lwsReplicas
```
Edwinhr716, line 298: Why are we not processing scale up/down before triggering the rolling update? In case of a scale up, it means that the partition would be at n+2, even though the 2 will have the newer version.

Edwinhr716:
Can you add an integration test here https://github.com/kubernetes-sigs/lws/blob/main/test/integration/controllers/leaderworkerset_test.go to check that the fix actually works?

Edwinhr716:
@kerthcet, @ahg-g  do you see an issue with always scaling up to lws-replica + maxSurge at the beginning of the rolling update, and releasing one by one as needed?

ahg-g:
> @kerthcet, @ahg-g do you see an issue with always scaling up to lws-replica + maxSurge at the beginning of the rolling update, and releasing one by one as needed?

What is the semantics of the Deployment API? mirroring that would be ideal I think.

Mostafahassen1:
Hi @ahg-g ,@Edwinhr716,

I want to check if the current solution is okay. If it is okay, I will update the unit tests. and if there's another way, please tell me to update  code 
Thank you for your time.

PR #725, opened 2026-01-05
Add RecreateGroupAfterStart annotation

Edwinhr716:
#### What type of PR is this?
/kind feature

Adding new functionality where the group is only restarted if and only if there are no pending pods in the group.

Inference servers such as vLLM have images that take a while to download. When increasing the size of LWS, this means that some pods in the group can take longer to download the image and be ready. vLLM has a timeout for all workers to be up and running, so often this timeout can expire, triggering failure across the whole group, and getting stuck on a loop of restarts since all the workers cannot fully download the image.

We can address this by restarting the whole group only if all pods are not pending. This way, the workers with timeout errors can restart, while the pods waiting for the image can finish downloading the image, allowing all of the workers to sync up.

Part of #726 

```release-note

```

### DIFF
--- a/api/leaderworkerset/v1/leaderworkerset_types.go
+++ b/api/leaderworkerset/v1/leaderworkerset_types.go
@@ -93,9 +93,9 @@ const (
 	// will be injected. Corresponds to LeaderWorkerSet.Spec.NetworkConfig.SubdomainPolicy
 	SubdomainPolicyAnnotationKey string = "leaderworkerset.sigs.k8s.io/subdomainPolicy"
 
-	// Enables feature where the group will be restarted if and only if all pods in the
-	// group are not pending
-	RecreateGroupAfterStart string = "leaderworkerset.sigs.k8s.io/recreateGroupAfterStart"
+	// Enables feature where the group will be restarted after pod failure if and only if
+	// all pods in the group are not pending
+	RecreateGroupAfterStart string = "leaderworkerset.sigs.k8s.io/experimentalRecreateGroupAfterStart"
 )
 
 // One group consists of a single leader and M workers, and the total number of pods in a group is M+1.


api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -92,6 +92,10 @@ const (
 	// Leader pods will have an annotation that determines what type of domain
 	// will be injected. Corresponds to LeaderWorkerSet.Spec.NetworkConfig.SubdomainPolicy
 	SubdomainPolicyAnnotationKey string = "leaderworkerset.sigs.k8s.io/subdomainPolicy"
+
+	// Enables feature where the group will be restarted if and only if all pods in the
+	// group are not pending
+	RecreateGroupAfterStart string = "leaderworkerset.sigs.k8s.io/recreateGroupAfterStart"
```
ahg-g: experimental-recreate-group-after-start

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -92,6 +92,10 @@ const (
 	// Leader pods will have an annotation that determines what type of domain
 	// will be injected. Corresponds to LeaderWorkerSet.Spec.NetworkConfig.SubdomainPolicy
 	SubdomainPolicyAnnotationKey string = "leaderworkerset.sigs.k8s.io/subdomainPolicy"
+
+	// Enables feature where the group will be restarted if and only if all pods in the
+	// group are not pending
```
ahg-g: clarify that this is in the context of failure handling

api/leaderworkerset/v1/leaderworkerset_types.go
Edwinhr716: Done

api/leaderworkerset/v1/leaderworkerset_types.go
Edwinhr716: Added

Edwinhr716:
Created issue, and linked it to the PR

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -92,6 +92,10 @@ const (
 	// Leader pods will have an annotation that determines what type of domain
 	// will be injected. Corresponds to LeaderWorkerSet.Spec.NetworkConfig.SubdomainPolicy
 	SubdomainPolicyAnnotationKey string = "leaderworkerset.sigs.k8s.io/subdomainPolicy"
+
+	// Enables feature where the group will be restarted after pod failure if and only if
```
kannon92, line 96: Why make this an annotation?

Would the eventual goal be to make this an option in the API once its no longer experimental?

Edwinhr716: Correct, we want to get feedback first, and if it is positive graduate it to an API

kannon92:
Will you update the site with this annotation?

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -92,6 +92,10 @@ const (
...
+	// all pods in the group are not pending
+	RecreateGroupAfterStart string = "leaderworkerset.sigs.k8s.io/experimentalRecreateGroupAfterStart"
```
ahg-g, line 98: we have some inconsistency in naming, sometimes we use camelcase and sometimes dashes, I strongly prefer the latter.

Edwinhr716:
Since it is experimental, not sure if we should add the annotation to the docs yet

kannon92:
> Since it is experimental, not sure if we should add the annotation to the docs yet

How will you get feedback without documenting it?

api/leaderworkerset/v1/leaderworkerset_types.go
Edwinhr716: Created https://github.com/kubernetes-sigs/lws/pull/728 to change it to use dashes

Edwinhr716:
Good point, will add the option here https://lws.sigs.k8s.io/docs/concepts/failure-handling/ labeling it as experimental after #728 is merged

PR #733, opened 2026-01-15
Returning if the leader statefulset is being deleted

### LINKED ISSUE #732: LWS controller will create a revision connected to a StatefulSet that is being deleted
**What happened**:
When quickly deleting an LWS instance and creating it again, the controller will create a revision for a StatefulSet that is being deleted. This is due to the statefulset taking a while to be deleted, so when the controller checks the leader Sts here https://github.com/kubernetes-sigs/lws/blob/main/pkg/controllers/leaderworkerset_controller.go#L111, it will find the statefulset that corresponds to the previews LWS object. 

Wait for the leader Sts to be deleted before attempting to create a new controller revision.

Create an LWS instance, delete it and immediately create it again.

**Environment**:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Edwinhr716:
#### What type of PR is this?
/kind bug

Fixes #732 

```release-note

```

### DIFF
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -115,7 +116,7 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	}
 
 	if leaderSts != nil && leaderSts.DeletionTimestamp != nil {
-		return ctrl.Result{Requeue: true}, nil
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 	}
 
 	// Handles two cases:


pkg/controllers/leaderworkerset_controller.go
```diff
@@ -114,6 +114,10 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, err
 	}
 
+	if leaderSts.DeletionTimestamp != nil {
+		return ctrl.Result{}, nil
+	}
+
```
ahg-g, line 121: What will trigger a reconcile again?

Edwinhr716: Added a requeue request

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -114,6 +114,10 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, err
 	}
 
+	if leaderSts != nil && leaderSts.DeletionTimestamp != nil {
+		return ctrl.Result{Requeue: true}, nil
```
ahg-g: I would use requeueAfter and give it like 5 seconds

Edwinhr716: Changed

Issue #734, opened 2026-01-15
Deploying lws using kustomize in the same namespace with other controllers like jobset creates conflicts


akshaychitneni:
**What happened**:
Deploying lws via kustomize only adds `control-plane: controller-manager` label and selector causing conflcits when multiple controller-managers like jobset are deployed in the same namesapce as described in this [jobset issue ](https://github.com/kubernetes-sigs/jobset/issues/1084)

**What you expected to happen**:
Lws should work with multiple controller-managers in the same namespace as with [kubeflow data cache](https://www.kubeflow.org/docs/components/trainer/user-guides/data-cache/) usecase

**How to reproduce it (as minimally and precisely as possible)**:
Deploy lws and jobset in the same namespace with Kustomize.
Kustomize.yaml
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namespace: jobset-lws-system
patches:
  - patch: |-
      apiVersion: v1
      kind: Namespace
      metadata:
        name: not-used
      $patch: delete
    target:
      kind: Namespace
resources:
  - https://github.com/kubernetes-sigs/jobset/releases/download/v0.10.1/manifests.yaml
  - https://github.com/kubernetes-sigs/lws/releases/download/v0.7.0/manifests.yaml
```
 
```
 kubectl get services -n jobset-lws-system -o custom-columns=NAME:.metadata.name,SELECTOR:.spec.selector
NAME                                        SELECTOR
jobset-controller-manager-metrics-service   map[control-plane:controller-manager]
jobset-webhook-service                      map[control-plane:controller-manager]
lws-controller-manager-metrics-service      map[control-plane:controller-manager]
lws-webhook-service                         map[control-plane:controller-manager]
```

**Anything else we need to know?**:

**Environment**:

- Kubernetes version (use `kubectl version`): v1.33.4
- LWS version (use `git describe --tags --dirty --always`): v0.7.0
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

andreyvelich:
This should be fixed by: https://github.com/kubernetes-sigs/lws/pull/739
/close

PR #739, opened 2026-01-22
fix(manifests): Fix common labels for Kustomize deployment

andreyvelich:
#### What type of PR is this?

/kind bug

Adding the common labels to LWS Kustomize deployment to follow the JobSet change: https://github.com/kubernetes-sigs/jobset/pull/1120
That should allow to deploy JobSet and LWS in the same namespace.

/cc @ahg-g @Edwinhr716 @akshaychitneni

Fixes # https://github.com/kubernetes-sigs/lws/issues/734

```release-note

```

### DIFF
--- a/charts/lws/crds/leaderworkerset.x-k8s.io_leaderworkersets.yaml
+++ b/charts/lws/crds/leaderworkerset.x-k8s.io_leaderworkersets.yaml
@@ -3,6 +3,9 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.17.2
+  labels:
+    app.kubernetes.io/instance: lws
+    app.kubernetes.io/name: lws
   name: leaderworkersets.leaderworkerset.x-k8s.io
 spec:
   conversion:
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -8,11 +8,10 @@ namespace: lws-system
 # field above.
 namePrefix: lws-
 
-# Labels to add to all resources and selectors.
-#labels:
-#- includeSelectors: true
-#  pairs:
-#    someName: someValue
+# Common labels to add to all resources and selectors (matches Helm behavior).
+commonLabels:
+  app.kubernetes.io/name: lws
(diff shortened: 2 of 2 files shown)


charts/lws/crds/leaderworkerset.x-k8s.io_leaderworkersets.yaml
```diff
@@ -3,6 +3,9 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.17.2
+  labels:
```
Edwinhr716, line 6: Are these necessary? I don't see them in the sister PR https://github.com/kubernetes-sigs/jobset/pull/1120

andreyvelich: This is due to auto-copy CRDs into Helm Charts after Kustomize Build: https://github.com/kubernetes-sigs/lws/blob/main/Makefile#L393-L395

Do we want to update this script to copy CRDs from the manifests directly, like we do in JobSet: https://github.com/kubernetes-sigs/jobset/blob/main/Makefile#L113 ?

Edwinhr716: Yes, we can do that in a separate PR though, just create an issue for it. I'll approve this one

andreyvelich: Actually, LWS has custom patches for LWS CRD in the Kustomize manifests: https://github.com/kubernetes-sigs/lws/blob/main/config/crd/kustomization.yaml#L11
It looks like we are loosing them in JobSet while just copying CRDs from [the `base/components` folder](https://github.com/kubernetes-sigs/jobset/tree/main/config/components/crd) in JobSet
(cc @kannon92 @GiuseppeTT )

We might want to see how Kueue manages Kustomize + Helm via processing plan in the future: https://github.com/kubernetes-sigs/kueue/blob/main/hack/processing-plan.yaml

PR #757, opened 2026-02-13
Graduate RecreateGroupAfterStart to an API

### LINKED ISSUE #726: Support only failing the group if all pods are not pending
A failure mechanism where the group only fails if all pods are not pending

**Why is this needed**:
Inference servers such as vLLM have images that take a while to download. When increasing the size of LWS, this means that some pods in the group can take longer to download the image and be ready. vLLM has a timeout for all workers to be up and running, so often this timeout can expire, triggering failure across the whole group, and getting stuck on a loop of restarts since all the workers cannot fully download the image.

We can address this by restarting the whole group only if all pods are not pending. This way, the workers with timeout errors can restart, while the pods waiting for the image can finish downloading the image, allowing all of the workers to sync up.

**Completion requirements**:
First enable the feature via an annotation. Once it has been soaked, add a new field to the API called `recreateGroupAfterStart` under `RestartPolicy`

- [X] Design doc
- [X] API change
- [X] Docs update

adinilfeld:
#### What type of PR is this?

/feature

The `RecreateGroupAfterStart` restart policy causes the whole group to restart upon pod failure only when all pods are Ready. This policy addresses the issue of pods timing out due to other pods still pulling an image, leading to the whole group restarting and never becoming ready. PR #725 added an annotation to enable this behavior; the current PR graduates that annotation to an API. We also maintain the annotation's functionality for backwards compatibility.

Fixes #726

- I tested manually on B200s; see http://b/482102724 (Google-internal access only).
- I updated the `Size()` method for integration tests to represent the total size of the LWS, including the leader pod. Previously it referred to the number of worker pods, which contradicted the use of "size" in the LWS spec.

```release-note
NONE
```

### DIFF
--- a/site/content/en/docs/concepts/failure-handling/_index.md
+++ b/site/content/en/docs/concepts/failure-handling/_index.md
@@ -36,10 +36,23 @@ spec:
 
 When any pod in a group fails, the entire group is recreated if and only if there are no pods currently pending. This allows for large image pulls to complete without interruption.
 
+On version 0.9+, this feature can be enabled via the `restartPolicy` field:
+
 ```yaml
 spec:
   leaderWorkerTemplate:
     restartPolicy: RecreateGroupAfterStart
+```
+
+On version 0.8, this feature is experimental and can be enabled via the following annotation:
+
+```yaml
+apiVersion: leaderworkerset.x-k8s.io/v1
+kind: LeaderWorkerSet
+metadata:
+  name: leaderworkerset-sample
+  annotations:
+    leaderworkerset.sigs.k8s.io/experimental-recreate-group-after-start: true
 ```
 
 ## Node Failure Handling


site/content/en/docs/concepts/failure-handling/_index.md
```diff
@@ -32,18 +32,14 @@ spec:
     restartPolicy: None
 ```
 
-### RecreateGroupAfterStart (Experimental)
+### RecreateGroupAfterStart
 
-When any pod in a group fails, the entire group is recreated if and only if there are no pods currently pending.
-This feature is currently experimental on versions 0.8+.
+When any pod in a group fails, the entire group is recreated if and only if there are no pods currently pending. This allows for large image pulls to complete without interruption.
```
Edwinhr716, line 37: We should mention that the feature is in experimental in 0.8, keep the annotation doc for it, and also mention that on LWS 0.9+ it is an API field

pkg/controllers/pod_controller.go
```diff
@@ -248,7 +249,7 @@ func (r *PodReconciler) handleRestartPolicy(ctx context.Context, pod corev1.Pod,
 	}); err != nil {
 		return false, err
 	}
-	r.Record.Eventf(&leaderWorkerSet, corev1.EventTypeNormal, "RecreateGroupOnPodRestart", fmt.Sprintf("Worker pod %s failed, deleted leader pod %s to recreate group %s", pod.Name, leader.Name, leader.Labels[leaderworkerset.GroupIndexLabelKey]))
+	r.Record.Eventf(&leaderWorkerSet, corev1.EventTypeNormal, "GroupRecreation", fmt.Sprintf("Worker pod %s failed, deleted leader pod %s to recreate group %s", pod.Name, leader.Name, leader.Labels[leaderworkerset.GroupIndexLabelKey]))
```
Edwinhr716: nit: I think RecreateGroup makes more sense, since both `RecreateGroupOnPodRestart` and `RecreateGroupAfterPodStart` have that in the name.

test/integration/controllers/leaderworkerset_test.go
```diff
@@ -407,7 +407,7 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
 		}),
 		ginkgo.Entry("Pod restart will not recreate the pod group when restart policy is None", &testCase{
 			makeLeaderWorkerSet: func(nsName string) *wrappers.LeaderWorkerSetWrapper {
-				return wrappers.BuildLeaderWorkerSet(nsName).RestartPolicy(leaderworkerset.NoneRestartPolicy).Replica(1).Size(3)
+				return wrappers.BuildLeaderWorkerSet(nsName).RestartPolicy(leaderworkerset.NoneRestartPolicy).Replica(1).Size(4)
```
Edwinhr716, line 410: Why are we changing from 3 to 4 here? I don't see another change in the test.

test/integration/controllers/leaderworkerset_test.go
```diff
@@ -2004,7 +2076,7 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
 		}),
 		ginkgo.Entry("resize should update the size of the replicas", &testCase{
 			makeLeaderWorkerSet: func(nsName string) *wrappers.LeaderWorkerSetWrapper {
-				return wrappers.BuildLeaderWorkerSet(nsName).Replica(2).Size(2)
```
Edwinhr716, line 2007: Similar, why are we going from 2 to 3, and 3 to 4 later?

test/testutils/util.go
```diff
@@ -57,7 +57,7 @@ func MustCreateLws(ctx context.Context, k8sClient client.Client, lws *leaderwork
 
 func CreateWorkerPodsForLeaderPod(ctx context.Context, leaderPod corev1.Pod, k8sClient client.Client, lws leaderworkerset.LeaderWorkerSet) {
 	gomega.Eventually(func() error {
-		for i := 1; i <= int(*lws.Spec.LeaderWorkerTemplate.Size); i++ {
+		for i := 1; i < int(*lws.Spec.LeaderWorkerTemplate.Size); i++ {
```
Edwinhr716, line 60: Oh I'm guessing this is why we bump the size. But why is this change needed?

adinilfeld: I mentioned briefly in the PR description - It's not strictly necessary, but previously `Size` referred to the number of worker pods, which contradicted the use of "size" in the LWS spec (which includes the leader). I found that confusing when updating the tests, so I think it makes sense to standardize it.

test/integration/controllers/leaderworkerset_test.go
adinilfeld: See below - just to match the use of `size` in the actual LWS spec.

test/integration/controllers/leaderworkerset_test.go
adinilfeld: See below - just to match the use of `size` in the actual LWS spec.

pkg/controllers/pod_controller.go
adinilfeld: Agreed, fixed.

site/content/en/docs/concepts/failure-handling/_index.md
adinilfeld: Done.

adinilfeld:
> Forgot to mention, can you run `make crds`? The command updates the helm charts

This didn't lead to any changes - `charts/lws/crds/leaderworkerset.x-k8s.io_leaderworkersets.yaml` was already updated in the original commit. Are there any other changes you were expecting to see?

Edwinhr716:
> his didn't lead to any changes - charts/lws/crds/leaderworkerset.x-k8s.io_leaderworkersets.yaml was already updated in the original commit. Are there any other changes you were expecting to see?

Nothing specific, just wanted to make sure all needed files in the chart were updated.

Issue #710, opened 2025-12-01
Support GroupIndex to be random string to accelerate rollout


hasB4K:
**What would you like to be added**:
The possibility to have GroupIndex to not be integers for deployment that don't need it. Something like that:

```
NAME       READY   STATUS    RESTARTS   AGE
vllm-8hc4x     1/1     Running   0          2s
vllm-8hc4x-1   1/1     Running   0          2s
vllm-mk9w6     1/1     Running   0          2s
vllm-mk9w6-1   1/1     Running   0          2s
```

**Why is this needed**:

[Example from the doc](https://lws.sigs.k8s.io/docs/examples/vllm/#deploy-leaderworkerset-of-vllm):
```
NAME       READY   STATUS    RESTARTS   AGE
vllm-0     1/1     Running   0          2s
vllm-0-1   1/1     Running   0          2s
vllm-1     1/1     Running   0          2s
vllm-1-1   1/1     Running   0          2s
```
Group index are the first 0 and 1 so (in bold) vllm-**0**-1 and vllm-**1**-1. (it's obvious, but just in case: `GroupIndex == ReplicaIndex`).

The issue with that is when we do rollout, it always tries to keep consecutive group indices since LWS reconciles statefulSets and those statefulsets have consistent naming. So if I redeploy with maxSurge at 2, we will have this:
```
NAME       READY   STATUS    RESTARTS   AGE
vllm-0     1/1     Running   0          20h
vllm-0-1   1/1     Running   0          20h
vllm-1     1/1     Running   0          20h
vllm-1-1   1/1     Running   0          20h
vllm-2     0/1     Pending   0          2s
vllm-2-1   0/1     Pending   0          2s
vllm-3     0/1     Pending   0          2s
vllm-3-1   0/1     Pending   0          2s
```

Then this:
```
NAME       READY   STATUS    RESTARTS   AGE
vllm-0     0/1     Pending   0          2s
vllm-0-1   0/1     Pending   0          2s
vllm-1     0/1     Pending   0          2s
vllm-1-1   0/1     Pending   0          2s
vllm-2     1/1     Running   0          2m
vllm-2-1   1/1     Running   0          2m
vllm-3     1/1     Running   0          2m
vllm-3-1   1/1     Running   0          2m
```

and finally this again:
```
NAME       READY   STATUS    RESTARTS   AGE
vllm-0     1/1     Running   0          20s
vllm-0-1   1/1     Running   0          20s
[...]

Issue #665, opened 2025-10-01
A guide for using Kueue as a TAS scheduler for LWS


ahg-g:
**What would you like to be added**:

A user guide that shows how to deploy Kueue as a topology-aware scheduler for LWS. 

Ideally we come up with a helm chart in Kueue that allows deploying Kueue as a TAS scheduler only, abstracting the concepts of ClusterQueue, Flavor and LocalQueue; It would take a couple of parameters, the topology labels and the accelerator type.

We may need to automate creating LocalQueues, perhaps we should have Kueue create a default LocalQueue in each namespace.

Edwinhr716:
Current state of Kueue + LWS https://kueue.sigs.k8s.io/docs/tasks/run/leaderworkerset/

Edwinhr716:
> Ideally we come up with a helm chart in Kueue that allows deploying Kueue as a TAS scheduler only, abstracting the concepts of ClusterQueue, Flavor and LocalQueue; It would take a couple of parameters, the topology labels and the accelerator type.
>
>We may need to automate creating LocalQueues, perhaps we should have Kueue create a default LocalQueue in each namespace.

Should we open an issue on the Kueue side as well for this ask?

ahg-g:
I think it is good to start the discussion on this, yes

kannon92:
The other relevant documentation would be https://kueue.sigs.k8s.io/docs/concepts/topology_aware_scheduling/

kannon92:
hmm. 

cc @mimowo @tenzen-y 

We can discuss here or in Kueue but this is an interesting feature request..

kannon92:
I think this is a valid use case for LWS but in general it seems there is a lot of interest in collocating Training and Serving workloads.

In your request, it sounds like you want to only serve a single tenant designed for LWS (maybe llm-d) and you are not really focused on multi tenant?

kannon92:
> We may need to automate creating LocalQueues, perhaps we should have Kueue create a default LocalQueue in each namespace.

Starting in 0.12, Kueue has a [localQueueDefaulting](https://kueue.sigs.k8s.io/docs/tasks/manage/enforce_job_management/setup_default_local_queue/) feature where it will use the "default" localQueue if it exists.

I'm not entirely sure how you automate that with a helm chart as this has to be created for each namespace you support.

ahg-g:
Not via helm, I want Kueue to create the default localQueue if it doesn't exist in each namespace that points to a default CQ.

[...]

PR #767, opened 2026-03-05
KEP-766: Add DisaggregatedSet proposal

### LINKED ISSUE #766: Add DisaggDeployment CRD for coordinated prefill/decode workload management

  A new `DisaggDeployment` CRD that manages two LeaderWorkerSets (prefill + decode) as a single logical unit, with coordinated rolling updates and
   service orchestration.

  **Why is this needed**:

  Disaggregated LLM inference (prefill/decode separation) requires coordinating two LeaderWorkerSets. Currently users must manually manage both
  resources and handle rolling updates across them, which is error-prone and complex.

  **Completion requirements**:

  * `DisaggDeployment` CRD with prefill and decode side configurations
  * Two-dimensional coordinated rolling updates across both sides
  * Automatic Service creation when both sides are ready

  * [x] KEP
  * [ ] Implementation
  * [ ] Docs update

hasB4K:
This KEP proposes adding DisaggDeployment as a higher-level API for managing disaggregated inference workloads. DisaggDeployment orchestrates multiple LeaderWorkerSets (prefill and decode) with coordinated lifecycle management, including two-dimensional rolling updates and service orchestration.

#### What type of PR is this?
 
/kind feature
/kind api-change

This KEP proposes adding DisaggDeployment as a higher-level API for managing disaggregated inference workloads. DisaggDeployment orchestrates multiple LeaderWorkerSets (prefill and decode) with coordinated lifecycle management, including two-dimensional rolling updates and service orchestration.

Fixes #766

As discussed offline, the implementation of this KEP already exists, and will be pushed tomorrow. I will change my implementation based on the review of this KEP and the review of the coming implementation.

Yes and no. It's a new API.

### DIFF
--- a/keps/766-DisaggregatedSet/README.md
+++ b/keps/766-DisaggregatedSet/README.md
@@ -116,7 +116,6 @@ type DisaggregatedSetSpec struct {
 }
 
 // DisaggregatedRoleSpec defines the configuration for a disaggregated role.
-// This structure embeds LeaderWorkerSetSpec from sigs.k8s.io/lws.
 type DisaggregatedRoleSpec struct {
     // Name is the unique identifier for this role.
     // +kubebuilder:validation:MinLength=1
@@ -125,19 +124,26 @@ type DisaggregatedRoleSpec struct {
     // +required
     Name string `json:"name"`
 
-    // LeaderWorkerSetSpec is embedded inline to inherit all LWS configuration fields.
+    // LeaderWorkerSetTemplateSpec is embedded inline to inherit LWS template fields.
     // Note: RolloutStrategy.Type must be RollingUpdate (or empty) and
     // RolloutStrategy.RollingUpdateConfiguration.Partition must not be set.
     // DisaggregatedSet handles rollouts across roles and does not propagate
     // RolloutStrategy to the underlying LWS resources.
-    leaderworkerset.LeaderWorkerSetSpec `json:",inline"`
+    leaderworkerset.LeaderWorkerSetTemplateSpec `json:",inline"`
+}
 
-    // Metadata allows setting labels and annotations on the LWS CR's ObjectMeta.
-    // This is useful for integrations like Kueue (queue assignment via
-    // kueue.x-k8s.io/queue-name label) and LWS exclusive-topology scheduling
-    // (leaderworkerset.sigs.k8s.io/exclusive-topology label).
+// LeaderWorkerSetTemplateSpec describes the data a LeaderWorkerSet should have when created
+// from a template. This type needs to be added to the LWS API (similar to PodTemplateSpec).
+type LeaderWorkerSetTemplateSpec struct {
+    // Metadata for the LWS CR. Labels and annotations are propagated to the LWS ObjectMeta.
+    // Useful for Kueue integration (kueue.x-k8s.io/queue-name) and exclusive-topology
+    // scheduling (leaderworkerset.sigs.k8s.io/exclusive-topology).
     // +optional
-    Metadata *metav1.ObjectMeta `json:"metadata,omitempty"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    // Spec defines the LeaderWorkerSet configuration.
+    // +optional
+    Spec leaderworkerset.LeaderWorkerSetSpec `json:"spec,omitempty"`
 }
 ```
 


Edwinhr716:
Can you run this script https://github.com/kubernetes-sigs/lws/blob/main/hack/update-toc.sh to fix the formatting in the KEP?

kannon92:
And please sign the CLA.

keps/766-DisaggDeployment/README.md
```diff
@@ -0,0 +1,390 @@
...
+    AutoPopulateSelector *bool `json:"autoPopulateSelector,omitempty"`
+}
+```
+
+**Naming Convention**: LeaderWorkerSets are named `{disaggdeployment-name}-{revision}-{side}` where:
+- `revision` is a truncated hash of both pod templates (ensures coordinated updates)
+- `side` is either `prefill` or `decode`
+
+**Labels**: The following labels are applied to managed LeaderWorkerSets:
+- `leaderworkerset.x-k8s.io/disagg-name`: DisaggDeployment name
```
Edwinhr716: wouldn't it be better for these labels to have a `dissaggdeployment.x-k8s.io` prefix instead so that we know they were added by this controller?

keps/766-DisaggDeployment/README.md
```diff
@@ -0,0 +1,390 @@
...
+### Graduation Criteria
+
+**Alpha (v0.1)**:
+- DisaggDeployment CRD with validation
+- LeaderWorkerSet creation and ownership
+- Two-dimensional rolling update algorithm
+- Service orchestration
+- Comprehensive test coverage (>80%)
+- Documentation and examples
+- Feature gate: `DisaggDeployment`
```
Edwinhr716: I don't think we need a feature gate here since it is a separate controller

keps/766-DisaggDeployment/kep.yaml
```diff
@@ -0,0 +1,13 @@
+title: DisaggDeployment
+kep-number: 766
+authors:
+  - "Mathis Felardos"
+status: implementable
+creation-date: 2026-03-05
+reviewers:
+  - TBD
```
Edwinhr716: you can add @ahg-g and I as reviewers / approvers. 

cc @kerthcet

keps/766-DisaggDeployment/README.md
hasB4K: Agreed! But I think @ahg-g wanted this controller to be part of LWS entirely at some point?

Edwinhr716: My understanding was that we will keep them as separate APIs, but package them together under the same binary. Not that LWS will absorb the logic of this controller.

hasB4K: Okay, makes sense. I will change the labels then.

keps/766-DisaggDeployment/kep.yaml
kerthcet: I will take a look since I'm been waiting for this feature for a while.

keps/766-DisaggDeployment/README.md
```diff
@@ -0,0 +1,389 @@
...
+
+2. **Coordinated Rolling Updates**: Implement a two-dimensional rolling update algorithm that updates both sides in lockstep, respecting per-side surge constraints.
+
+3. **Service Orchestration**: Automatically manage Services based on workload readiness, ensuring traffic only flows when both sides are operational.
+
+4. **Stateless Controller**: Design the controller to derive all state from observed resources, enabling safe restarts at any point.
+
+### Non-Goals
+
+1. **Auto-scaling support**: HPA/VPA integration or automatic scaling based on inference load metrics is out of scope.
```
Edwinhr716: You can still autoscale at the prefill / decode level right?

keps/766-DisaggDeployment/README.md
```diff
@@ -0,0 +1,389 @@
...
+// fields like ServiceTemplate.
+type DisaggSideConfig struct {
+    // Replicas is the number of leader-worker groups.
+    // +kubebuilder:validation:Minimum=0
+    // +kubebuilder:default=1
+    Replicas *int32 `json:"replicas,omitempty"`
+
+    // LeaderWorkerTemplate defines the template for leader/worker pods.
+    // +required
+    LeaderWorkerTemplate LeaderWorkerTemplate `json:"leaderWorkerTemplate"`
```
Edwinhr716: Is there a way to pass annotations that will be injected to the LWS object? The exclusive-placement feature relies on that to work https://lws.sigs.k8s.io/docs/concepts/#exclusive-lws-to-topology-placement

keps/766-DisaggDeployment/README.md
hasB4K: Yes, I think it should work. I haven't tested it in my implementation, but I don't see any reason why it couldn't work. I put it as Non-Goal since it's not the focus of the operator for now, and I actually think that scaling Prefill and Decode independently makes more sense than scaling the entire DisaggDeployment directly.

keps/766-DisaggDeployment/README.md
Edwinhr716: An idea would be to add a field here called `ExclusivePlacement` and if set, have the controller inject the annotation. Though would need some sort of config to set `subgroup-exclusive-placement` as well.

hasB4K: if it doesn't work out of the box in my implementation, we could add it. I'll see if we can add annotations without changing anything tomorrow morning.

keps/766-DisaggDeployment/README.md
Edwinhr716: > I actually think that scaling Prefill and Decode independently

Yeah that's the one that I was referring to, still be able to scale independently while treating it as a unit

keps/766-DisaggDeployment/README.md
Edwinhr716: Yeah it is a future work type of thing, just thinking out loud. We can help with the implementation as well.

keps/766-DisaggDeployment/README.md
```diff
@@ -0,0 +1,389 @@
...
+| 2    | 2           | 2          | 1           | 1          |
+| 3    | 2           | 2          | 2           | 2          |
+| 4    | 1           | 1          | 2           | 2          |
+| 5    | 1           | 1          | 3           | 3          |
+| 6    | 0           | 0          | 3           | 3          |
+
+**Example 2: Asymmetric Replicas** (9 prefill, 3 decode → 9 prefill, 3 decode, maxSurge=1):
+
+When prefill and decode have different replica counts, the algorithm computes steps based on the larger dimension (9 in this case). The smaller side (decode) progresses proportionally slower:
+
```
Edwinhr716: If I'm understanding correctly, this means that during updates we are keeping track of four different LWS instances correct?

Can you expand more on why you opted for this implementation instead of for example using the partition field in LWS similar to how we use the partition field in Sts at the LWS level? Is the only gap https://github.com/kubernetes-sigs/lws/issues/710?

You can add it to the alternatives considered section, and we can discuss later if there are any improvements we can make at the LWS level to simplify this algorithm, or if it is worth it at all.

keps/766-DisaggDeployment/README.md
```diff
@@ -0,0 +1,389 @@
...
+1. **Operational complexity**: Users must manually ensure both sides are updated together and handle failure scenarios across two resources.
+
+2. **Rolling update coordination**: There is no built-in mechanism to coordinate rolling updates across prefill and decode sides, risking service disruption if one side is updated without the other.
+
+3. **Service lifecycle**: Users must manually manage Services and ensure they only route traffic when both sides are ready.
+
+4. **Configuration drift**: Without a unified resource, prefill and decode configurations can drift apart, leading to subtle incompatibilities.
+
+### Goals
```
kannon92: Has there been any thoughts on how this would work with Kueue?

What I am wondering is if this DisaggDeployment is quota limited do you want to avoid creating the LWS's?

keps/766-DisaggDeployment/README.md
hasB4K: Okay, I was not aware of the partition field. So just read about it, and maybe I'm wrong here, but I don't think it would work:

- **Revision-aware traffic routing**: DisaggDeployment is designed for disaggregated inference, where a load balancer (e.g., a modified LLM-d Endpoint Picker) must route prefill requests to backends whose decode counterparts are on the **same revision**. With separate LWS (and service) per revision, each pod's revision is explicit via labels (`disaggdeployment.x-k8s.io/revision`). The LB can count backends per revision across both pools and distribute traffic proportionally. With partition-based updates, pods within the same LWS have different templates based on ordinal, making revision-aware routing significantly more complex. The goal is to avoid having a prefill talk to an incompatible decode. Both sides must be treated as totally incompatible. This special routing requires work on the LLM-d side.

- **LWS as a Read-Only resource**: Treating LWS as a Read-Only resource (like a ReplicaSet) makes more sense, as you want to update Prefill and Decode at different paces depending on the step you are at. It's a tied update. I don't think you can achieve this with partition.

- **Ops observability**: It's simpler for ops observability. You can see directly at which stage your update is, since you can see the version right away during updates. (see scrennshot)

image

keps/766-DisaggDeployment/README.md
hasB4K: In my understanding Kueue is working thanks to annotations right? I would say that it works out of the box, and if adding annotation doesn't work directly, we could add support:

see:
https://github.com/kubernetes-sigs/lws/pull/767#discussion_r2892561661

kannon92: Well Kueue works with annotations for LWS because it has a internal controller that works with it. https://github.com/kubernetes-sigs/kueue/tree/main/pkg/controller/jobs/leaderworkerset

But I guess I'm mostly curious on things like Kueue's TopologyAwareScheduling (ie #665 ) with dissaggreatedserving

keps/766-DisaggDeployment/README.md
hasB4K: I added a section about this

keps/766-DisaggDeployment/README.md
hasB4K: Well, I'm not sure 🙂 - I guess we will need to experiment

keps/766-DisaggDeployment/README.md
Edwinhr716: Thanks for adding the section in the KEP. A few follow up questions that I have: 

> With partition-based updates, pods within the same LWS have different templates based on ordinal, making revision-aware routing significantly more complex

I'm assuming that the labels are injected at the LWS level, but can't we achieve the same logic if the disagg revision label is injected to the leader pod instead? Then, those above the partition will have the newer revision label, those below will have the old label, so the router can still do the mapping between old version and newer version. This might depend on the router's implementation, but I don't think it make a difference at what level the label is injected. 


> as you want to update Prefill and Decode at different paces depending on the step you are at. It's a tied update. I don't think you can achieve this with partition.

Why not? I don't see the difference between scaling prefill & decode vs changing the partition value independently. 

> Ops observability

This is I think would be the main benefit. But I wonder how much it matters to have this be easily readable. Neither StatefulSet nor LWS have this.

Just to make it clear, this is not a blocker for the merger of the KEP or the code implementation. I just think that it is worth discussing for the future if there's a simpler path here. The algorithm needed for rolling updates is complex already, and would be easier to understand if we stick to existing Kubernetes concepts like partition instead of trying to maintain four LWS replicas.

keps/766-DisaggDeployment/README.md
```diff
@@ -0,0 +1,416 @@
...
+// DisaggSideConfig defines the configuration for the prefill/decode side.
+// This structure embeds LeaderWorkerSetSpec fields and adds disagg-specific
+// fields like ServiceTemplate.
+type DisaggSideConfig struct {
+    // Replicas is the number of leader-worker groups.
+    // +kubebuilder:validation:Minimum=0
+    // +kubebuilder:default=1
+    Replicas *int32 `json:"replicas,omitempty"`
+
+    // LeaderWorkerTemplate defines the template for leader/worker pods.
```
Edwinhr716: Have you tested if setting size = 1 for both the prefill and the decode server works?

keps/766-DisaggDeployment/README.md
kannon92: Yea np. Its not blocking.

keps/766-DisaggDeployment/README.md
```diff
@@ -0,0 +1,416 @@
...
+## Summary
+
+This KEP proposes adding `DisaggDeployment` as a new Custom Resource Definition (CRD) to the LeaderWorkerSet (LWS) project. DisaggDeployment is a higher-level abstraction that orchestrates multiple LeaderWorkerSets with coordinated lifecycle management, specifically designed for disaggregated inference architectures where "prefill" and "decode" workloads run on separate infrastructure.
+
+Disaggregated serving is an optimization for LLM inference workloads that takes advantage of the fact that the two phases of inference (prefill and decode) have different computational characteristics. State-of-the-art LLM serving frameworks such as [vLLM](https://github.com/vllm-project/vllm) and [SGLang](https://github.com/sgl-project/sglang) support this optimization.
+
+DisaggDeployment simplifies the deployment of disaggregated LLM inference by:
+- Managing two LeaderWorkerSets (prefill and decode) as a single logical unit
+- Providing coordinated two-dimensional rolling updates across both sides
+- Automatically creating Services when both sides are ready
```
ahg-g: We have generally avoided bundling/instantiating easily composable resources. For example, in upstream, you don't see the workload apis (like Deployment, StatefulSet etc.) automatically create a Service.

keps/766-DisaggDeployment/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+1. **Auto-scaling support**: HPA/VPA integration or automatic scaling based on inference load metrics is out of scope.
+
+2. **Multi-cluster federation**: Managing DisaggDeployments across multiple Kubernetes clusters is not addressed.
+
+3. **Custom workload backends**: Supporting backends other than LeaderWorkerSet (e.g., StatefulSet, Deployment) is not planned.
+
+4. **Traffic management / routing**: Integration with service meshes or ingress controllers for traffic splitting during rollouts is out of scope.
+
+5. **Generic group support**: Supporting arbitrary groups beyond prefill/decode (e.g., `decode-long-context`, `encode`) is deferred to a future enhancement.
```
ahg-g: I think we should at least make the api future proof to support that.

keps/766-DisaggDeployment/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+## Proposal
+
+We propose adding a new CRD called `DisaggDeployment` that acts as a higher-level controller over LeaderWorkerSet resources. The DisaggDeployment controller will:
```
ahg-g: In general abbreviations are avoided in APIs.

A couple of suggestions: 
- `DisaggregatedDeployment` 
- `DisaggregatedSet` 

I prefer the latter since is aligns with `LeaderWorkerSet` API name.

keps/766-DisaggDeployment/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+As a platform engineer, I want to deploy a disaggregated LLM inference workload by creating a single DisaggDeployment resource that specifies both prefill and decode configurations, so that I don't have to manage two separate LeaderWorkerSets manually.
+
+#### Story 2: Rolling Update with Zero Downtime
+
+As a DevOps engineer, I want to update my model version across both prefill and decode sides with zero downtime, with the system automatically coordinating the rollout to ensure capacity is maintained throughout the update.
+
+#### Story 3: Coordinated Service Discovery
+
+As an application developer, I want Services to be automatically created for my prefill and decode endpoints, but only when both sides are ready to serve traffic, so that my application doesn't route requests to partially-ready infrastructure.
```
ahg-g: I do not think relying on the existence of the Service resource is the proper approach to address this problem. The P/D coordinator should handle this case, not the workload api.

keps/766-DisaggDeployment/README.md
```diff
@@ -0,0 +1,416 @@
...
+   - Scale-down uses `max(prefillStep, decodeStep)` to ensure both drain together
+
+3. **Coordinated Drain**: If either side reaches 0 replicas, both sides are forced to 0. This prevents orphaned single-side workloads from interrupted rollouts.
+
+4. **Surge Constraints**: Per-side `maxSurge` and `maxUnavailable` are respected:
+   ```
+   old + new <= target + maxSurge
+   ```
+
+5. **Scale-Up Priority**: New replicas are scaled up before old replicas are scaled down, ensuring capacity is never below the minimum.
```
ahg-g: Are we not concerned about limited hardware capacity? does this mean the cluster must have hardware capacity for at least one extra replica to conduct and update (vs doing in-place)? Can we do surge =0 and maxUnavailable = 1 to trigger an in-place rollout?

keps/766-DisaggDeployment/README.md
```diff
@@ -0,0 +1,416 @@
...
+|------|-------------|------------|-------------|------------|
+| 0    | 3           | 3          | 0           | 0          |
+| 1    | 3           | 3          | 1           | 1          |
+| 2    | 2           | 2          | 1           | 1          |
+| 3    | 2           | 2          | 2           | 2          |
+| 4    | 1           | 1          | 2           | 2          |
+| 5    | 1           | 1          | 3           | 3          |
+| 6    | 0           | 0          | 3           | 3          |
+
+**Example 2: Asymmetric Replicas** (9 prefill, 3 decode → 9 prefill, 3 decode, maxSurge=1):
```
ahg-g: Can we spill out the maxSurge semantics? is it for both prefill and decode or for each?

keps/766-DisaggDeployment/README.md
hasB4K: Renamed 😉 !

keps/766-DisaggDeployment/README.md
hasB4K: Yes, this works, and it's part of the test suite. Simply we favor maxSurge over maxUnavailable. We could change this behavior if you guys want.

keps/766-DisaggDeployment/README.md
hasB4K: It's for each side.

keps/766-DisaggDeployment/README.md
hasB4K: I think it's already future proof since:
```
apiVersion: disaggregatedset.x-k8s.io/v1alpha1
  kind: DisaggregatedSet
  metadata:
    name: my-llm
  spec:
    prefill:
      replicas: 3                      # Number of LWS groups
      rolloutStrategy:
        maxSurge: 1                    # Extra replicas during update
        maxUnavailable: 0              # Don't remove until new is ready
      serviceTemplate:                 # Optional: create Service when ready
        spec:
          ports:
          - port: 8080
      leaderWorkerTemplate:
        size: 2                        # 1 leader + 1 worker per group
        workerTemplate:
          spec:
            containers:
            - name: prefill
              image: vllm:latest
              resources:
                limits:
                  nvidia.com/gpu: 1

    decode:
      replicas: 6                      # Often more decode than prefill
      rolloutStrategy:
        maxSurge: 2
        maxUnavailable: 1
      serviceTemplate:
        spec:
          ports:
          - port: 8080
      leaderWorkerTemplate:
        size: 1                        # Single pod per group
        workerTemplate:
          spec:
            containers:
            - name: decode
              image: vllm:latest
              resources:
                limits:
                  nvidia.com/gpu: 1

    decode-long-context:  # <== adding a special group should work out of the box in the config.
        [ ... ]
```

What I mean is prefill and decode are currently hardcoded, but they could simply be treated as a standalone "side" with a name (either prefill or decode).

So except if I miss something (which is totally possible), I think it's already future proof?

keps/766-DisaggDeployment/README.md
hasB4K: The service is optional, and I thought it was okay to be able to create one since LWS already does this.

But anyway, we can discuss about this. it was easier to handle our scheduler on our side, but I understand your concerns.

keps/766-DisaggDeployment/README.md
hasB4K: The service is optional, and I thought it was okay to be able to create one since LWS already does this.

But anyway, we can discuss about this. it was easier to handle our scheduler on our side, but I understand your concerns.

keps/766-DisaggDeployment/README.md
hasB4K: You mean replicas==1 for both side? Yes I did, and it works.

hasB4K:
The implementation has been pushed here 🙂:
https://github.com/kubernetes-sigs/lws/pull/773

I wanted this PR to be on top of that branch, but since my branch is on my fork, I couldn't do it 😬 (it created a PR on my forked repo instead 🫤)

keps/766-DisaggDeployment/README.md
ahg-g: LWS only creates a **headless** service to trigger programming DNS entries for the individual workers, this is useful to standardize discovery among the leader-workers of a group (e.g., typically workers need to know the network address of their leader, for other the frameworks it could be the opposite, the leader needs to know the network addresses of its workers).

keps/766-DisaggDeployment/README.md
ahg-g: This is still hardcoding the prefill and decode as part of the api, we could instead have a list with named entries to make the api more generic. But we can follow up on that possibility when we merge it to run with the lws controller.

keps/766-DisaggDeployment/README.md
ahg-g: No, I think preferring maxSurge seems like the better semantics.

keps/766-DisaggDeployment/README.md
hasB4K: I guess you are suggesting this solution:
```
      Sides []DisaggSideConfig `json:"sides"`
  }

  type DisaggSideConfig struct {
      Name string `json:"name"` // "prefill", "decode", etc.
      // ... rest of fields
  }
  ```

  YAML:
  ```
  spec:
    sides:
      - name: prefill
        replicas: 10
        ...
      - name: decode
        replicas: 2
        ...
      - name: custom_stage
        replicas: 5
```

I need to say that this is an easy change, and I could do that (while only supporting prefill and decode for now).

-----------

What I had in mind however was this:

```
  Option 2: Pure Map (Breaking Change)

      Sides map[string]*DisaggSideConfig `json:"sides"`
  }
```

  YAML:
```
  spec:
    sides:
      prefill: <lws1>
      decode: <lws2>
      custom_stage: <lws3>
 ```

This would work with older YAML configs too.

Comment (Edwinhr716) [reply keps/766-DisaggDeployment/README.md]: > we could instead have a list with named entries to make the api more generic

If we go this route, we should make it clear somewhere that we only support synced rolling updates between the first two items in the list, not all of them.

File: keps/766-DisaggDeployment/README.md (hunk 4 above)
Comment (kerthcet) [reply keps/766-DisaggDeployment/README.md]: I think we support this naturally today since we support autoscaling for lws but we just need one example. Would you mind to remove it.

File: keps/766-DisaggregatedSet/README.md [hunk 15]
```diff
@@ -0,0 +1,318 @@
...
+
+3. **Stateless Controller**: Design the controller to derive all state from observed resources, enabling safe restarts at any point.
+
+### Non-Goals
+
+1. **Auto-scaling support**: HPA/VPA integration or automatic scaling based on inference load metrics is out of scope.
+
+2. **Multi-cluster federation**: Managing DisaggregatedSets across multiple Kubernetes clusters is not addressed.
+
+3. **Custom workload backends**: Supporting backends other than LeaderWorkerSet (e.g., StatefulSet, Deployment) is not planned.
```
kerthcet, line 70: Is there any plan to support this, because today the orchestration is becoming really complex, in some cases, we not only need the prefill pool, decode pool, we may also need cache pool, which is usually standard kubernetes workloads. 

Before, I was thinking of working on a new orchestration layer to coordinate different workloads, similar to argocd etc. but for long-running workloads to meet this requirements.

keps/766-DisaggregatedSet/README.md
```diff
@@ -0,0 +1,318 @@
...
+
+4. **Traffic management / routing**: Integration with service meshes or ingress controllers for traffic splitting during rollouts is out of scope.
+
+## Proposal
+
+We propose adding a new CRD called `DisaggregatedSet` that acts as a higher-level controller over LeaderWorkerSet resources. The DisaggregatedSet controller will:
+
+1. Create and manage multiple LeaderWorkerSets (one per phase, 2-10 phases supported)
```
kerthcet: Just curious to know how many phases your own cluster have.

keps/766-DisaggregatedSet/README.md
```diff
@@ -0,0 +1,318 @@
...
+2. Coordinate rolling updates using an N-dimensional algorithm
+3. Automatically create headless Services for each phase per revision
+
+### Risks and Mitigations
+
+**Risk**: The N-dimensional rolling update algorithm adds complexity that could lead to stuck rollouts.
```
kerthcet, line 84: Another risk might be A old prefill may sending traffic to a new decode instance, vice versa, is that possible, I have no idea how you handle the request dispatching.

keps/766-DisaggregatedSet/README.md
```diff
@@ -0,0 +1,318 @@
...
+
+    Spec   DisaggregatedSetSpec   `json:"spec"`
+    Status DisaggregatedSetStatus `json:"status,omitempty"`
+}
+
+// DisaggregatedSetSpec defines the desired state of DisaggregatedSet
+// +kubebuilder:validation:XValidation:rule="self.phases.all(p, self.phases.filter(q, q.name == p.name).size() == 1)",message="phase names must be unique"
+// +kubebuilder:validation:XValidation:rule="self.phases.all(p, p.replicas == 0) || self.phases.all(p, p.replicas > 0)",message="replicas must be zero for all phases or non-zero for all phases"
+type DisaggregatedSetSpec struct {
+    // Phases defines the list of phases (at least 2 required, maximum 10).
```
kerthcet: How to define the maximum number here? If it's randomed, maybe 2^n seems more convincing, like 8 or 16 here. Just from previous experience when defining some magic numbers.

keps/766-DisaggregatedSet/README.md
```diff
@@ -0,0 +1,318 @@
...
+    // Each phase has a unique name and its own configuration.
+    // +kubebuilder:validation:MinItems=2
+    // +kubebuilder:validation:MaxItems=10
+    // +required
+    Phases []DisaggregatedPhaseSpec `json:"phases"`
```
kerthcet: Phase usually means status, especially in Kubernetes Pod, maybe role is a more appropriate word here? Have we discussed this before?

keps/766-DisaggregatedSet/README.md
```diff
@@ -0,0 +1,318 @@
...
+    // +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
+    // +required
+    Name string `json:"name"`
+
+    // LeaderWorkerSetSpec is embedded inline to inherit all LWS configuration fields.
+    // Note: RolloutStrategy.Type must be RollingUpdate (or empty) and
+    // RolloutStrategy.RollingUpdateConfiguration.Partition must not be set.
+    // DisaggregatedSet handles rollouts across phases and does not propagate
+    // RolloutStrategy to the underlying LWS resources.
+    leaderworkerset.LeaderWorkerSetSpec `json:",inline"`
```
kerthcet: I'm thinking should we add a LeaderWorkerSetTemplateSpec here just like the PodTemplateSpec cc @ahg-g @Edwinhr716  which includes the metadata and spec both.

keps/766-DisaggregatedSet/README.md
hasB4K: The standard setup is Prefill / Decode. But Encode / Prefill / Decode is another thing you can have.
You can also have Encode / Prefill / Decode / Decode-long-context.

keps/766-DisaggregatedSet/README.md
hasB4K: Yes, we have something that handle that at Mistral. We want to integrates that in llm-d, but it's out of scope of this Operator.

keps/766-DisaggregatedSet/README.md
hasB4K: I think we could increase the number. 10 was indeed a bit arbitrary, 16 seems to high though. The maximum number of phases I can think of is 5 so far:
* encode
* prefill
* prefill-long-context
* decode
* decode-long-context

keps/766-DisaggregatedSet/README.md
Edwinhr716: So rename this field to `LeaderWorkerSetTemplateSpec`?

hasB4K: LeaderWorkerSetTemplateSpec doesn't exists yet, we could switch to it when we merge the implementation in the LWS binary though.

keps/766-DisaggregatedSet/README.md
hasB4K: Abduallah and I agrees with you, I made the change 😉.

FYI, it used to be `Side` then we discussed about switching to `Phase` in the implem PR: https://github.com/kubernetes-sigs/lws/pull/773#discussion_r2918748253

keps/766-DisaggregatedSet/README.md
hasB4K: Honestly, I'm not against it. But that would requires a future KEP and a deeper reflection I think.

keps/766-DisaggregatedSet/README.md
ahg-g: Yes, but this is the KEP, so I suggest we mention it here that creating this type and using it here is the plan of record.

keps/766-DisaggDeployment/README.md
hasB4K: Closing since we already had a discussion online on this.

keps/766-DisaggregatedSet/README.md
hasB4K: yeah makes sense, I added that in the last commit

https://github.com/kubernetes-sigs/lws/pull/767/changes/42a5501bcbb359e22c6288382e484a7116a03549

PR #773, opened 2026-03-09
KEP-766: DisaggregatedSet implementation

### LINKED ISSUE #766: Add DisaggDeployment CRD for coordinated prefill/decode workload management

hasB4K:
#### What type of PR is this?

/kind feature
/kind api-change

  This PR implements KEP-766 by introducing the DisaggregatedSet controller, a Kubernetes operator for managing disaggregated inference deployments. Disaggregated serving separates the prefill and decode phases
  of LLM inference onto different infrastructure, and this controller orchestrates multiple LeaderWorkerSets with coordinated lifecycle management.

#### Key features:
  - Unified Management: Manage prefill and decode LeaderWorkerSets as a single resource
  - Two-Dimensional Rolling Updates: Linear interpolation algorithm updates both sides in lockstep while preserving the prefill-to-decode ratio
  - Service Orchestration: Automatically create Services when both sides are ready
  - Stateless Operator: Safe to restart at any point during operations

Fixes #766

  The two-dimensional rollout algorithm uses linear interpolation to maintain the prefill/decode ratio throughout updates. A plan-steps CLI tool is included to visualize rollout plans:

```
go run ./hack/plan-steps --source '{"prefill": 5, "decode": 2}' --target '{"prefill": 5, "decode": 2}' --surge '{"prefill": 2, "decode": 2}'
Phases: [decode prefill]
Source: decode=2, prefill=5
Target: decode=2, prefill=5
Config: decode(surge=2, unavail=0), prefill(surge=2, unavail=0)

┌──────┬────────────┬─────────────┬────────────┬─────────────┬───────┬───────────────────────────────┐
│ Step │ Old decode │ Old prefill │ New decode │ New prefill │ Total │            Action             │
├──────┼────────────┼─────────────┼────────────┼─────────────┼───────┼───────────────────────────────┤
│ 0    │ 2          │ 5           │ 0          │ 0           │ 7     │ initial                       │
│ 1    │ 2          │ 5           │ 1          │ 2           │ 10    │ new decode +1, new prefill +2 │
│ 2    │ 2          │ 4           │ 1          │ 2           │ 9     │ old prefill -1                │
│ 3    │ 2          │ 3           │ 1          │ 2           │ 8     │ old prefill -1                │
│ 4    │ 2          │ 3           │ 2          │ 4           │ 11    │ new decode +1, new prefill +2 │
│ 5    │ 1          │ 2           │ 2          │ 4           │ 9     │ old decode -1, old prefill -1 │
│ 6    │ 1          │ 2           │ 2          │ 5           │ 10    │ new prefill +1                │
│ 7    │ 0          │ 0           │ 2          │ 5           │ 7     │ old decode -1, old prefill -2 │
└──────┴────────────┴─────────────┴────────────┴─────────────┴───────┴───────────────────────────────┘
```

  Introduce `DisaggregatedSet` API for managing disaggregated LLM inference deployments with coordinated prefill/decode rolling updates.
  
  
#### Current TODO:

- [x] List based instead of map based phases in the YAML config
- [x] (that was optional, but realized it was easy with the previous point done): **support multiple phases (ie. more than 2) to be able to support encode  phase (among other)**
- [x] Remove the Service in the YAML API. We now have a Headless Service that exposes a EndpointSlice (for llm-d EndpointPicker in the future)
- [x] Massive refactor of the e2e tests.
- [x] Propagate the annotations to the LWS level to support Kueue topology
- [x] See if I can use LeaderWorkerSetSpec in the YAML instead of LeaderWorkerTemplate as I do now.
- [x] Add the CI
- [x] Update the KEP and the doc.

### DIFF
--- a/disaggregatedset/test/e2e/e2e_test.go
+++ b/disaggregatedset/test/e2e/e2e_test.go
@@ -102,6 +102,14 @@ var _ = Describe("DisaggregatedSet E2E Tests", Ordered, func() {
 
 	Context("Operator Deployment", func() {
 		It("should have the controller-manager running", func() {
+			By("checking if controller-manager is already deployed")
+			cmd := exec.Command("kubectl", "get", "deployment",
+				"disaggregatedset-controller-manager", "-n", namespace, "-o", "name")
+			output, err := utils.Run(cmd)
+			if err == nil && strings.Contains(output, "deployment") {
+				_, _ = fmt.Fprintf(GinkgoWriter, "Controller-manager already deployed, skipping deployment\n")
+			}
+
 			By("validating that the controller-manager pod is running")
 			verifyControllerUp := func(g Gomega) {
 				cmd := exec.Command("kubectl", "get", "pods", "-l", "control-plane=controller-manager",
@@ -126,6 +134,29 @@ var _ = Describe("DisaggregatedSet E2E Tests", Ordered, func() {
 				g.Expect(output).To(Equal("Running"))
 			}
 			Eventually(verifyControllerUp).Should(Succeed())
+
+			By("waiting for webhook to be ready")
+			// The webhook needs time for certificate provisioning and endpoint registration.
+			// We verify readiness by attempting a dry-run create that triggers the webhook.
+			verifyWebhookReady := func(g Gomega) {
+				yaml := fixtures.PrefillDecode("webhook-test",
+					fixtures.Role{Replicas: 1},
+					fixtures.Role{Replicas: 1},
+				).YAML()
+				cmd := exec.Command("kubectl", "apply", "--dry-run=server", "-f", "-")
+				cmd.Stdin = strings.NewReader(yaml)
+				_, err := utils.Run(cmd)
+				// We just need the webhook to respond, not necessarily succeed.
+				// Connection errors indicate webhook not ready, validation errors are fine.
+				if err != nil {
+					errStr := err.Error()
+					// These errors mean webhook is not ready yet
+					g.Expect(errStr).NotTo(ContainSubstring("connection refused"), "webhook not ready")
+					g.Expect(errStr).NotTo(ContainSubstring("no endpoints available"), "webhook not ready")
+					g.Expect(errStr).NotTo(ContainSubstring("connection reset"), "webhook not ready")
+				}
+			}
+			Eventually(verifyWebhookReady, 2*time.Minute, 2*time.Second).Should(Succeed())
 		})
 	})
 


disaggregatedset/internal/controller/workload_manager.go
Edwinhr716: Labels for Kueue and exclusive-topology need to be added at this level. We need a way to add them.

disaggregatedset/api/v1alpha1/disaggregatedset_types.go
```diff
@@ -0,0 +1,186 @@
...
+	StartupPolicy leaderworkerset.StartupPolicyType `json:"startupPolicy,omitempty"`
+
+	// NetworkConfig defines the network configuration of the group.
+	// +optional
+	NetworkConfig *leaderworkerset.NetworkConfig `json:"networkConfig,omitempty"`
+
+	// ServiceTemplate defines an optional Service to create alongside the LWS.
+	// Services are only created when both prefill and decode sides are ready.
+	// +optional
+	ServiceTemplate *ServiceTemplate `json:"serviceTemplate,omitempty"`
```
ahg-g: As I mentioned on the KEP, this is an anti-pattern for workload apis even as an optional feature. A workload api shouldn't be concerned with how its pods are exposed for serving beyond applying the labels. Also, note that while Service is the most common API currently, it is not really the only one (InferencePool is another example).

disaggregatedset/api/v1alpha1/disaggregatedset_types.go
```diff
@@ -0,0 +1,186 @@
...
+
+// DisaggSideConfig defines the configuration for the prefill/decode side.
+// This structure embeds LeaderWorkerSetSpec from sigs.k8s.io/lws and adds
+// disagg-specific fields like ServiceTemplate.
+type DisaggSideConfig struct {
+	// Replicas is the number of leader-worker groups.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=1
+	// +optional
+	Replicas *int32 `json:"replicas,omitempty"`
```
ahg-g: I am sure you thought you about this, but what is the reason for not using a LeaderWorkerSetSpec instead of explicitly listing the individual parameters?

disaggregatedset/api/v1alpha1/disaggregatedset_types.go
```diff
@@ -0,0 +1,186 @@
...
+	// +kubebuilder:validation:XIntOrString
+	// +kubebuilder:default=1
+	// +optional
+	MaxSurge *intstr.IntOrString `json:"maxSurge,omitempty"`
+}
+
+// DisaggSideConfig defines the configuration for the prefill/decode side.
+// This structure embeds LeaderWorkerSetSpec from sigs.k8s.io/lws and adds
+// disagg-specific fields like ServiceTemplate.
+type DisaggSideConfig struct {
```
ahg-g: Perhaps `DisaggregatedPhaseSpec`?

hasB4K: Changed.

disaggregatedset/api/v1alpha1/disaggregatedset_types.go
hasB4K: As discussed offline (and I'll need to explain the why on the KEP): 
I removed the Service in the YAML API. We now have a Headless Service that exposes a EndpointSlice (for llm-d EndpointPicker in the future)

hasB4K:
Here what has been achieved since our last sync:
- [x] List based instead of map based phases in the YAML config
- [x] (that was optional, but realized it was easy with the previous point done): **support multiple phases (ie. more than 2) to be able to support encode  phase (among other)**
- [x] Remove the Service in the YAML API. We now have a Headless Service that exposes a EndpointSlice (for llm-d EndpointPicker in the future)
- [x] Massive refactor of the e2e tests.
- [x] Patched a bug in the planner.

Here what i still need to do:
- [x] Propagate the annotations to the LWS level to support Kueue topology
- [x] See if I can use LeaderWorkerSetSpec in the YAML instead of LeaderWorkerTemplate as I do now.
- [x] Add the CI
- [x] Update the KEP and the doc.

(I added a global TODO in the head message)

cc @ahg-g, @Edwinhr716

disaggregatedset/api/v1alpha1/disaggregatedset_types.go
```diff
@@ -0,0 +1,195 @@
...
+	// PhasePolicyStrict requires that the set of phase names remains constant.
+	// Adding, removing, or renaming phases is not allowed.
+	// This is the default policy.
+	PhasePolicyStrict PhasePolicy = "Strict"
+
+	// PhasePolicyFlexible allows adding, removing, or renaming phases.
+	// Removed phases are drained immediately (scaled to 0).
+	// New phases start at 0 and scale up.
+	PhasePolicyFlexible PhasePolicy = "Flexible"
+)
```
ahg-g: At first look, this doesn't feel right, validation should be deterministic, not conditional. We either allow to change the list of phases or we don't, and my preference is to start with making the list immutable.

disaggregatedset/hack/plan-steps/main.go
```diff
@@ -0,0 +1,230 @@
...
+//
+// Usage:
+//
+//	go run ./cmd/plan-steps \
+//	  --source '{"prefill": 6, "decode": 2}' \
+//	  --target '{"prefill": 6, "decode": 2}' \
+//	  --surge '{"prefill": 2, "decode": 2}'
+//
+// This helps understand what will happen during a specific rollout in advance.
+// Supports arbitrary phase names and will support N phases when the planner does.
```
ahg-g: So this is a tool that one runs locally, not a controller?

hasB4K: > So this is a tool that one runs locally, not a controller?

`cmd/plan-steps/main.go` is just a util script that I used for debug. it can also be used to know what will happen before one does a deployment. BUT the rest of the code in `disaggregatedset/internal/` is the controller/operator. 😉

disaggregatedset/internal/controller/workload_manager.go
hasB4K: @Edwinhr716 I checked, labels and annotations are propagated, I added a check in the e2e tests:
https://github.com/kubernetes-sigs/lws/pull/773/changes/7b1a2d8818fd7ac68192b2a80cdd2245f340eb9f 

Are you sure this is not enough?

Here is an example:
EDIT: removed

disaggregatedset/api/v1alpha1/disaggregatedset_types.go
hasB4K: Regarding the PhasePolicy = "Flexible" - I actually wanted to make it by default at Strict. Here is my rationale to add the support of Flexible mode though:

* It works natively with the planner.go.
* We need to have this, one of my team mate for instance wants to be able to change a deploy from: `["prefill", "decode"]` to `["prefill", "decode", "decode-long-context"]`. There is already some people within the community that could benefit from this. And I would prefer if the rolling update already handled within the DisaggregatedSet operator.


The rationale to have a Strict mode, is to be extra safe within a Deployment. But that was mostly the case before I handled the Flexible case. Now the renaming technically already works within this operator. So I would simply remove the PhasePolicy altogether and let it keep the behavior at "Flexible".

Would you be okay with that?

disaggregatedset/api/v1alpha1/disaggregatedset_types.go
hasB4K: @ahg-g, @Edwinhr716 :

I remember why I took the decision of not using LeaderWorkerSetSpec.
DisaggregatedPhaseSpec does not include rolloutStrategy from LeaderWorkerSetSpec because the DisaggregatedSet operator handles rolling updates itself:

Rollout type: We don't want users to pick a rollout strategy (RollingUpdate, OnDelete, etc.). The operator has its own logic to roll all phases together.
Partition: LWS uses partition to roll pods one by one, but this only works inside a single LWS. We can't sync partition across prefill and decode, so pods would end up on different versions. (this I didn't know)

Instead, the operator creates new LWS resources for each revision and scales them up/down together. Services only point to revisions where all phases are ready.
I think the API should be opinionated, and this is typically one of this case when I don't want to expose types (like RollouType or partition) that doesn't makes sense in the DisaggregatedSet operator

disaggregatedset/internal/controller/workload_manager.go
hasB4K: Okay. Since the labels/annotations should be on the LWS level, here the commit that fix this: https://github.com/kubernetes-sigs/lws/commit/5ef69d48a6dd27bb73498807e99f698e0f72a40d#diff-85187aef20f2f22a40fb063cb51097e33c38e0e508f4dbe4b1d9fa2c1797f694

This works (I added some tests):
```
apiVersion: disaggregatedset.x-k8s.io/v1alpha1
kind: DisaggregatedSet
metadata:
  name: disagg-test
spec:
  phases:
    - name: prefill
      replicas: 1
      # PROBLEM: No field to set LWS CR metadata labels!
      # These labels go to pods, NOT to the LWS CR:
      leaderWorkerTemplate:
        size: 1
        workerTemplate:
          metadata:
            labels:
              # This goes to pods, but Kueue needs it on LWS CR
              kueue.x-k8s.io/queue-name: my-queue
          spec:
            containers:
              - name: worker
                image: busybox:latest
                command: ["sleep", "infinity"]

    - name: decode
      replicas: 1
      leaderWorkerTemplate:
        size: 1
        workerTemplate:
          spec:
            containers:
              - name: worker
                image: busybox:latest
                command: ["sleep", "infinity"]
```

Edwinhr716: Great, this addresses it, thanks! Will close the ocmment.

disaggregatedset/api/v1alpha1/disaggregatedset_types.go
hasB4K: Since I truly believe we need flexible phases, I removed the Strict support here:
https://github.com/kubernetes-sigs/lws/commit/c76946426ff12aa52a991b9de97cb889eee1cd0c

But we can discuss about this if you want 😉

disaggregatedset/api/v1alpha1/disaggregatedset_types.go
ahg-g: Here is one option: we can use the LeaderWorkerSetSpec and validate that only the supported fields are set. The advantage is that this way the apis are always in sync (we don't need to copy every field), but at the same time we control which fields the new api should support. What do you think?

hasB4K: Okay fair. I can try to do that, I'll let you know.

disaggregatedset/hack/plan-steps/main.go
hasB4K: I moved this to hack/plan-steps/ - I think it's clearer that way.

disaggregatedset/api/v1alpha1/disaggregatedset_types.go
hasB4K: it's done!

disaggregatedset/internal/controller/disaggregatedset_controller.go
```diff
@@ -0,0 +1,317 @@
...
+	if err := reconciler.Get(ctx, req.NamespacedName, disaggregatedSet); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	log.Info("Reconciling DisaggregatedSet", "name", disaggregatedSet.Name, "namespace", disaggregatedSet.Namespace)
+
+	// Validate phases are configured (API validation ensures at least 2)
```
Edwinhr716: Can we move all validation to the disaggregatedset_webhook?

Edwinhr716:
I took a look at the code, nothing major sticks out. Will approve once the last comment is addressed.

/lgtm

disaggregatedset/internal/controller/disaggregatedset_controller.go
ahg-g: +1

hasB4K: Okay, this was dead code. I removed this in this [commit](https://github.com/hasB4K/lws/commit/078de81).

The validation is done at the CRD level:
https://github.com/hasB4K/lws/blob/078de81/disaggregatedset/api/v1alpha1/disaggregatedset_types.go#L55-L64 

Is this okay? Or do you want me to implement this logic in the webhook code?

Edwinhr716: That's fine, just wanted to make sure that it is either happening in the disaggregatedset_webhook or the kubebuilder validation.

Edwinhr716:
Thanks for addressing, will wait for tests to be fixed to merge

disaggregatedset/api/v1alpha1/disaggregatedset_types.go
```diff
@@ -0,0 +1,144 @@
...
+	leaderworkerset "sigs.k8s.io/lws/api/leaderworkerset/v1"
+)
+
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// DisaggregatedPhaseSpec defines the configuration for a disaggregated phase.
+// This structure embeds LeaderWorkerSetSpec from sigs.k8s.io/lws, with validation
+// to reject unsupported fields (RolloutStrategy.Type must be RollingUpdate,
+// RolloutStrategy.RollingUpdateConfiguration.Partition must not be set).
+type DisaggregatedPhaseSpec struct {
```
ahg-g: We discussed changing `Phase` to `Role`, I think Role is better.

disaggregatedset/api/v1alpha1/disaggregatedset_types.go
```diff
@@ -0,0 +1,144 @@
...
+	leaderworkerset.LeaderWorkerSetSpec `json:",inline"`
+
+	// Metadata allows setting labels and annotations on the LWS CR's ObjectMeta.
+	// This is useful for integrations like Kueue (queue assignment via
+	// kueue.x-k8s.io/queue-name label) and LWS exclusive-topology scheduling
+	// (leaderworkerset.sigs.k8s.io/exclusive-topology label).
+	// Only labels and annotations are used; other ObjectMeta fields are ignored.
+	// System labels (disaggregatedset.x-k8s.io/*, app) take precedence over user labels.
+	// +optional
+	Metadata *metav1.ObjectMeta `json:"metadata,omitempty"`
```
ahg-g, line 52: As discussed on the KEP, it would be better if we create a `LeaderWorkerSetTemplateSpec` type in https://github.com/kubernetes-sigs/lws/blob/main/api/leaderworkerset/v1/leaderworkerset_types.go as follows

```
type LeaderWorkerSetTemplateSpec struct {
	// Metadata of the LeaderWorkerSets created from this template.
	// +optional
	metav1.ObjectMeta

	// +optional
	Spec leaderworkerset.LeaderWorkerSetSpec
}
```

and use it here instead of explicitly setting two parameters

hasB4K: Agreed, but for now it doesn't exists. There is the `LeaderWorkerSet` type though, but there is the LeaderWorkerSetStatus in it. 

https://github.com/kubernetes-sigs/lws/blob/d4d6b172cf767386d48f7fb695cb79206f8c022c/api/leaderworkerset/v1/leaderworkerset_types.go#L420-L429

ahg-g: Sounds good

hasB4K: Okay, let's add `LeaderWorkerSetTemplateSpec` once we will merge it to the LWS binary (except if I misunderstood you).

(I said that in the KEP PR too btw: https://github.com/kubernetes-sigs/lws/pull/767#discussion_r2987289192)

PR #777, opened 2026-03-12
Speed up surge replica garbage collection during rollouts by correctly computing wantReplicas

adinilfeld:
This PR refactors the `wantReplicas` garbage collection logic to evaluate the explicit `maxUnavailable` value from the rollout configuration instead of simply using a hardcoded `1`. 

Previously, surge replicas were unnecessarily kept alive when `maxUnavailable > 1`. By dynamically reading the user's configured `maxUnavailable` instead of `- 1`, we allow rollouts to safely maintain necessary parallelism while efficiently cleaning up extra surge capacity as soon as it's mathematically safe.

### Example Scenario
Given a `LeaderWorkerSet` with the following configuration:
* `replicas: 4`
* `maxSurge: 2`
* `maxUnavailable: 2`

**Before (Buggy Logic):**
1. Rollout starts: 4 old replicas. `wantReplicas` = 6. Restart 2 old, create 2 new surge replicas.
2. 2 old replicas update and become ready. 2 old are unready. 2 surge are ready. `unreadyReplicas` (2) &lt;= `maxSurge` (2).
3. `finalReplicas = 4 + (2 - 1) = 5`.
4. The controller scales down the StatefulSet to 5. The result is that it unnecessarily holds onto 1 surge replica (costing cluster resources) even though the user was completely fine with `maxUnavailable: 2` (needing only 2 ready pods).

**After (New Logic):**
1. Rollout starts: 4 old replicas. `wantReplicas` = 6. Restart 2 old, create 2 new surge replicas.
2. 2 old replicas update and become ready. 2 old are unready. 2 surge are ready. 
3. `finalReplicas = 4 + max(0, 2 - 2) = 4`.
4. The controller correctly identifies that it can scale down to 4 immediately (killing *both* surge replicas), keeping exactly 2 ready pods alive and respecting the user's `maxUnavailable: 2` constraint without wasting any surge pods.

Added an integration test directly verifying this exact scenario to prevent regression. I confirmed that the test fails without the updates to the controller.

### DIFF
--- a/site/content/en/docs/concepts/rollout-strategy/_index.md
+++ b/site/content/en/docs/concepts/rollout-strategy/_index.md
@@ -7,10 +7,8 @@ description: >
 
 Rolling update is vital to online services with zero downtime. For LLM inference services, this is particularly important, which helps to mitigate stockout. Two different configurations are supported in LWS, `maxUnavailable` and `maxSurge`:
 
-- `MaxUnavailable`: Indicates how many replicas are allowed to be unavailable during the update, the unavailable number is based on the spec.replicas. Defaults to 1.
+- `MaxUnavailable`: Indicates how many replicas are allowed to be unavailable during the update, the unavailable number is based on the spec.replicas. Defaults to 1. Note that only values >= 1 are supported.
 - `MaxSurge`: Indicates how many extra replicas can be deployed during the update. Defaults to 0.
-
-Note that maxSurge and maxUnavailable can not both be zero at the same time.
 
 Here's a leaderWorkerSet configured with rollout strategy, you can find the example [here](https://github.com/kubernetes-sigs/lws/blob/main/docs/examples/sample/lws-rollout-strategy.yaml):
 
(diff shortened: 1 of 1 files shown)


pkg/controllers/leaderworkerset_controller.go
```diff
@@ -288,13 +288,18 @@ func (r *LeaderWorkerSetReconciler) rollingUpdateParameters(ctx context.Context,
...
+
 	// wantReplicas calculates the final replicas if needed.
 	wantReplicas := func(unreadyReplicas int32) int32 {
 		if unreadyReplicas <= int32(maxSurge) {
 			// When we have n unready replicas and n bursted replicas, we should
 			// start to release the burst replica gradually for the accommodation of
 			// the unready ones.
-			finalReplicas := lwsReplicas + utils.NonZeroValue(int32(unreadyReplicas)-1)
+			finalReplicas := lwsReplicas + utils.NonZeroValue(unreadyReplicas-int32(maxUnavailableInt))
 			r.Record.Eventf(lws, nil, corev1.EventTypeNormal, GroupsProgressing, Delete, fmt.Sprintf("deleting surge replica %s-%d", lws.Name, finalReplicas))
```
Edwinhr716: This log line is no longer accurate since we are now deleting more than one surge replica at the same time. We should include the range of replicas instead.

adinilfeld: Done, thanks for the catch.

adinilfeld:
> We need to enable the featuregate MaxUnavailable at the Sts level

I'm not sure this is required, as the feature has been enabled by default since 1.32. At least, it worked locally for me. I'll push an update if the remote e2e tests fail.

Edwinhr716:
> I'm not sure this is required, as the feature has been enabled by default since 1.32. 

It is enabled by default from 1.35 onward https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#maximum-unavailable-pods. You can still set it to value that isn't 1, but in practice it will only do `maxUnavailable = 1`

Edwinhr716:
Looking at it again, I don't think we can actually validate anything extra by setting maxUnavailable to 2 in an e2e test, so we can take it out (sorry!). 

One more thing, we should update the site to showcase this logic instead https://github.com/kubernetes-sigs/lws/blob/main/site/content/en/docs/concepts/rollout-strategy/_index.md

adinilfeld:
> One more thing, we should update the site to showcase this logic instead https://github.com/kubernetes-sigs/lws/blob/main/site/content/en/docs/concepts/rollout-strategy/_index.md

Done, thanks for calling that out.

> Looking at it again, I don't think we can actually validate anything extra by setting maxUnavailable to 2 in an e2e test, so we can take it out

I think it makes sense to keep the test anyway - even if we can't verify the behavior of maxUnavailable directly, at least it shows that it can be set together with maxSurge, and doesn't interfere with the overall rollout process. Happy to remove the test case entirely though if you think that's preferable.

adinilfeld:
Converted to a draft because PR #778 overlaps with this implementation, and we should decide how to consolidate before merging either. See https://github.com/kubernetes-sigs/lws/pull/778#issuecomment-4056690088

adinilfeld:
Closing since #778 was merged, and includes this logic.

Issue #776, opened 2026-03-11
Support maxUnavailable=0


adinilfeld:
**What would you like to be added**:

The ability to set `maxUnavailable=0` for LWS RollingUpdate. Currently, the value of `maxUnavailable` is passed to the underlying "leader StatefulSet"; and as StatefulSet requires `maxUnavailable >= 1`, setting it to 0 causes an error:
```
2026-03-11T17:41:30Z	ERROR	Reconciler error	{"controller": "leaderworkerset", "controllerGroup": "leaderworkerset.x-k8s.io", "controllerKind": "LeaderWorkerSet", "LeaderWorkerSet": {"name":"leaderworkerset-rollout","namespace":"default"}, "namespace": "default", "name": "leaderworkerset-rollout", "reconcileID": "24960973-5a89-4b80-8a35-bdd349e59f39", "error": "StatefulSet.apps \"leaderworkerset-rollout\" is invalid: spec.updateStrategy.rollingUpdate.maxUnavailable: Invalid value: 0: cannot be 0"}
```

(As an aside: setting `maxUnavailable=0` causes the LWS pods to silently fail to be created; the above error is present in the logs, but not Events. If we continue to disallow 0, we should ideally fail fast by validating it upon LWS creation.)

**Why is this needed**:

Not necessarily needed, but would bring the LWS RollingUpdate API in line with Deployments and DaemonSets. Both of these support `maxSurge`, and allow for `maxUnavailable=0`. (By contrast, StatefulSets only support in-place update (no surge), so `maxUnavailable=0` is not possible.)

It would also allow for zero-downtime updates for single-replica deployments, as in Issue #688.

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

PR #778, opened 2026-03-13
Fix LWS surge handling during scale-up and zero-unavailable updates

ZRJ026:
#### What type of PR is this?

/kind bug
/kind regression

This PR fixes two inconsistencies in LeaderWorkerSet rolling update behavior when `maxSurge` is enabled.

- It prevents surge replicas from being reclaimed too early when `maxUnavailable=0`. Previously, a temporary surge group could become ready and then be deleted before the desired non-burst ordinal had actually rolled to the updated revision.
- It fixes scale-up behavior so the controller converges directly to the requested replica count instead of temporarily creating one extra burst replica and then deleting it.
- It adds regression tests for zero-unavailable surge reclamation and scale-up paths that previously created an unnecessary temporary replica.

None

This change is intentionally scoped to rolling update replica calculation and its controller unit tests.

The new regression tests use small replica counts to keep the scenarios minimal while still reproducing the previous incorrect behavior.

```release-note
Fix LeaderWorkerSet rolling update replica calculation so surge replicas are not reclaimed too early when maxUnavailable=0, and scale-up does not create an extra temporary replica beyond the requested desired count.
```

### DIFF
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -211,14 +211,29 @@ func (r *LeaderWorkerSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.Service{}).
 		Watches(&appsv1.StatefulSet{},
 			handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, a client.Object) []reconcile.Request {
-				return []reconcile.Request{
-					{NamespacedName: types.NamespacedName{
-						Name:      a.GetLabels()[leaderworkerset.SetNameLabelKey],
-						Namespace: a.GetNamespace(),
-					}},
+				return enqueueLWSRequests(a)
+			})).
+		Watches(&corev1.Pod{},
+			handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, a client.Object) []reconcile.Request {
+				if a.GetLabels()[leaderworkerset.WorkerIndexLabelKey] != "0" {
+					return nil
 				}
+				return enqueueLWSRequests(a)
 			})).
 		Complete(r)
+}
+
+func enqueueLWSRequests(a client.Object) []reconcile.Request {
+	name := a.GetLabels()[leaderworkerset.SetNameLabelKey]
+	if name == "" {
+		return nil
+	}
+	return []reconcile.Request{
+		{NamespacedName: types.NamespacedName{
+			Name:      name,
+			Namespace: a.GetNamespace(),
+		}},
+	}
 }
 
 func SetupIndexes(indexer client.FieldIndexer) error {


adinilfeld:
This PR overlaps with the logic I implemented in PR #777 - we should consolidate.

> It prevents surge replicas from being reclaimed too early when maxUnavailable=0.

FYI currently `maxUnavailable=0` doesn't actually work; see Issue #776. I could see it still making sense to fix the logic, in case we implement that in the future. What do you think?

adinilfeld:
> This PR overlaps with the logic I implemented in PR https://github.com/kubernetes-sigs/lws/pull/777 - we should consolidate.

I think your implementation is more comprehensive + testable, so maybe we just bring over the testing + documentation updates from my change.

adinilfeld:
Oh, we should also include the updated logging from my change, for when multiple surge pods are being deleted.

ZRJ026:
> Oh, we should also include the updated logging from my change, for when multiple surge pods are being deleted.

Thanks, I consolidated the relevant follow-ups from #777 into this PR.

I kept the controller logic in this PR as the base implementation, and also brought over:
  - the updated logging for deleting multiple surge replicas
  - the additional integration/e2e coverage for the maxSurge + maxUnavailable path
  - the docs update from #777

So this should now reflect the combined direction we discussed here, rather than two overlapping changes. If you’d prefer, I can still narrow the scope further, but I wanted to first align this PR with the testing/logging/docs follow-ups you called out.

adinilfeld:
> Thanks, I consolidated the relevant follow-ups from #777 into this PR.

LGTM, thanks @ZRJ026!

test/e2e/e2e_test.go
```diff
@@ -164,6 +164,25 @@ var _ = ginkgo.Describe("leaderWorkerSet e2e tests", func() {
 		testing.ExpectLeaderWorkerSetAvailable(ctx, k8sClient, lws, "All replicas are ready")
 	})
 
+	ginkgo.It("Can perform a rolling update with maxSurge and maxUnavailable set", func() {
```
Edwinhr716: I mentioned this on the other PR as well, we are not testing anything new here, so we should remove this test. While yes, it technically shows that MaxSurge + MaxUnavailable works, we can't do any checks in an e2e test to check that it is working as intended, just that an update is completed. We should only keep necessary e2e tests here. cc @adinilfeld

adinilfeld:
Hi @ZRJ026, I just wanted to check if you're planning on finishing up this change? My PR #781 depends on this one being submitted.
Thanks!

test/e2e/e2e_test.go
ZRJ026: Removed this e2e test. Agreed that it was not asserting enough new rollout behavior to justify keeping it in the e2e suite. @Edwinhr716

ZRJ026:
> Hi @ZRJ026, I just wanted to check if you're planning on finishing up this change? My PR #781 depends on this one being submitted. Thanks!

The redundant e2e test was removed per review

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -211,16 +211,31 @@ func (r *LeaderWorkerSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		Watches(&appsv1.StatefulSet{},
 			handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, a client.Object) []reconcile.Request {
-				return []reconcile.Request{
-					{NamespacedName: types.NamespacedName{
-						Name:      a.GetLabels()[leaderworkerset.SetNameLabelKey],
-						Namespace: a.GetNamespace(),
-					}},
+				return enqueueLWSRequests(a)
+			})).
+		Watches(&corev1.Pod{},
```
Edwinhr716, line 216: Why are we adding that the manager watches leader pods here?

ZRJ026: Because the surge reclaim decision depends on leader pod readiness in `getReplicaStates()`.

The specific failure I was addressing was `pull-lws-test-e2e-main-1-32`, in the existing e2e test `Can perform a rolling update with maxSurge set`, which was  timing out with `leader StatefulSet replicas should match with specified replicas, want 4, got 5`.

In that path, the controller could remain at 5 replicas until a leader pod flipped to Ready, but that readiness transition was not triggering a reconcile to  recompute the desired replica count back down to 4. I limited the watch to leader pods (`worker-index=0`) for that reason.

Edwinhr716: weird that this hasn't caused an issue before, makes sense to watch the leader pod as well

PR #779, opened 2026-03-16
fix: resolve deadlock when maxSurge>0 rolling update on single-replica LWS

### LINKED ISSUE #688: LWS does not support zero-downtime updates when replicas = 1

veast:
#### What type of PR is this?

/kind bug

Fixes a deadlock where a `LeaderWorkerSet` with `replicas=1` and
`maxSurge=1` could never complete a rolling update.

**Symptom**: immediately after triggering an update the controller emits
a `"deleting surge replica <name>-1"` event and the `StatefulSet` stalls
at `partition=1, replicas=1` forever. The surge pod is never actually
created.

**Root cause**: in Case 2 of `rollingUpdateParameters` (a new rolling
update is detected) the code called `wantReplicas(lwsReplicas)`. With
`replicas=1` and `maxSurge=1` the condition inside `wantReplicas` was:

```
unreadyReplicas(1) <= maxSurge(1)  →  true
```

which jumped straight into the "release surge" branch and returned
`replicas=1`. No surge replica was ever created, so the StatefulSet
`partition` could never advance.

**Fix**: two interrelated changes:

1. **Case 2 returns `burstReplicas` directly** instead of going through
   `wantReplicas`. At the moment a new update is detected all existing
   replicas are still running the old template (none are unready due to
   the update yet), so the correct action is to expand to
   `lwsReplicas + maxSurge` first. `wantReplicas` is only meaningful
   once `stsReplicas == burstReplicas` and the surge pods are being
   replaced.

2. **`wantReplicas` condition tightened from `<=` to `<`**. When
   `unreadyReplicas == maxSurge` exactly (e.g. `replicas=2, maxSurge=1`,
   one replica still unready) the surge pods should be kept alive until
   that last replica becomes ready. Using strict-less-than ensures we
   only enter the shrink path when there is genuine headroom.

Fixes #688

The logic change is small (two lines in `leaderworkerset_controller.go`)
but the comments have been expanded to make the invariants clearer.

A new integration test `"rolling update with maxSurge=1 and single
replica creates surge before rolling"` directly exercises the regression:
it verifies that the leader StatefulSet expands to `replicas=2`
immediately after the update is triggered (i.e. the surge is actually
created), then converges back to `replicas=1` once all groups are ready.

Yes — `maxSurge`-based zero-downtime rolling updates now work correctly
for single-replica `LeaderWorkerSet` objects.

### DIFF
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -288,12 +288,27 @@ func (r *LeaderWorkerSetReconciler) rollingUpdateParameters(ctx context.Context,
 	}
 	burstReplicas := lwsReplicas + int32(maxSurge)
 
-	// wantReplicas calculates the final replicas if needed.
+	// wantReplicas calculates the desired total replicas (including any surge)
+	// during an in-progress rolling update based on the number of currently
+	// unready (not-yet-updated) replicas. It is only called after the initial
+	// surge expansion has already happened (i.e. stsReplicas == burstReplicas).
+	//
+	// We enter the shrink path only when both conditions hold:
+	//   1. unreadyReplicas < lwsReplicas: at least one original replica has
+	//      already been successfully updated, so the surge pod has served its
+	//      purpose and can start being released.
+	//   2. unreadyReplicas <= maxSurge: the remaining unready replicas fit
+	//      within the surge budget, meaning we can safely reduce replicas.
+	//
+	// Without condition (1), a single-replica LWS (lwsReplicas=1, maxSurge=1)
+	// would satisfy (2) immediately on the first reconcile after Case 2 expands
+	// to burstReplicas, causing the surge to be released before any replica has
+	// actually been updated.
 	wantReplicas := func(unreadyReplicas int32) int32 {
-		if unreadyReplicas <= int32(maxSurge) {
-			// When we have n unready replicas and n bursted replicas, we should
-			// start to release the burst replica gradually for the accommodation of
-			// the unready ones.
+		if unreadyReplicas < lwsReplicas && unreadyReplicas <= int32(maxSurge) {
+			// At least one replica has been updated and the remaining unready
+			// count fits within the surge budget. Release one surge slot per
+			// newly-ready replica so we converge back on lwsReplicas.
 			finalReplicas := lwsReplicas + utils.NonZeroValue(int32(unreadyReplicas)-1)
 			r.Record.Eventf(lws, nil, corev1.EventTypeNormal, GroupsProgressing, Delete, fmt.Sprintf("deleting surge replica %s-%d", lws.Name, finalReplicas))
 			return finalReplicas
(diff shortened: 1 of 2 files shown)


Edwinhr716:
#688, or rather, the feature request for it,  is being fixed in #779. Can you check that PR to see if addresses what you are trying to fix here?

Issue #723, opened 2025-12-19
Provide the ability to configure TLS settings like TLSMinVersion and Ciphersuites


kannon92:
**What would you like to be added**:

LWS controller, metrics and webhooks should allow one to set TLSMinVersion and CipherSuites.

**Why is this needed**:

Some organizations like to enforce custom TLS profiles or even would like to only support TLS 1.3.

Kubernetes by default is 1.2.

**Ecosystem asks**

This is already supported in Kubernetes (tlsMinVersion and CiperSuite are cli arguments for kubernetes components).

Kueue issue: https://github.com/kubernetes-sigs/kueue/issues/8190
JobSet issue: https://github.com/kubernetes-sigs/jobset/issues/1107

This enhancement requires the following artifacts:

- [ ] Design doc
- [x] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

kannon92:
cc @Edwinhr716 @ardaguclu @kerthcet

ardaguclu:
Core components in k8s already support this (i.e. [kube-apiserver](https://github.com/kubernetes/kubernetes/blob/1c24a05ce16db3f68ccfa707f49c8ff664afa728/staging/src/k8s.io/apiserver/pkg/server/options/serving.go#L184-L195)). In my opinion, this is a crucial functionality for improving the security posture and can be a nice highlight for 0.9.0 release.

EDITED: If this is accepted, I can try to find some time for this (no guarantees though)

Mostafahassen1:
Hi @kannon92 , @kerthcet and @ardaguclu, I will start working on this feature. Is there any advice I should know? This is my first time implementing a feature

kannon92:
https://github.com/kubernetes-sigs/kueue/pull/8246 has an example KEP but there is still some discussion on it.

ardaguclu:
> [kubernetes-sigs/kueue#8246](https://github.com/kubernetes-sigs/kueue/pull/8246) has an example KEP but there is still some discussion on it.

Thank you @Mostafahassen1. I think best path is to follow the discussions in kueue and adopt the similar behavior here.

Edwinhr716:
Feature has been added. Will close this

PR #783, opened 2026-03-20
support TLS configuration for LWS

Mostafahassen1:
#### What type of PR is this?
/kind feature
/kind api-change

This PR introduces TLS support for the LWS webhook server, allowing secure communication by specifying minimum TLS versions and cipher suites.

Changes included in this PR:
1. Added `TLSOptions` to `api/config/v1alpha1/configuration_types.go` and regenerated deepcopy methods.
2. Created `pkg/config/tls.go` to handle parsing and created `validateTLSOptions` in `pkg/config/validation.go`.
3. Refactored the webhook initialization logic by moving it out of `addTo()` and into a new exported function `AddWebhookSettingsTo` in `pkg/config/config.go` to prevent configuration conflicts.
4. Updated `cmd/main.go` to parse the TLS options and properly wire them to the new `AddWebhookSettingsTo` function.
5. Added the `tls:` stanza to the controller manager config.

Fixes #[#723]

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package config
 
 import (
+	"crypto/tls"
 	"errors"
 	"io/fs"
 	"net"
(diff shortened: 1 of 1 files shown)


config/manager/controller_manager_config.yaml
```diff
@@ -2,5 +2,3 @@ apiVersion: config.lws.x-k8s.io/v1alpha1
 kind: Configuration
 leaderElection:
   leaderElect: true
-internalCertManagement:
```
kannon92, line 5: This seems like it's not needed.

config/manager/kustomization.yaml
```diff
@@ -13,5 +13,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: us-central1-docker.pkg.dev/k8s-staging-images/lws/lws
+  newName: us-central1-docker.pkg.dev/k8s-staging-images/lws
```
kannon92: Can you revert this?

config/manager/controller_manager_config.yaml
Mostafahassen1: Done

config/manager/kustomization.yaml
Mostafahassen1: Done

Mostafahassen1:
Hi @Edwinhr716 @kannon92 

Thank you for triggering the tests! I noticed that the `pull-lws-test-unit-main` job is failing. 

I looked into the test logs, and the failures are occurring in `pkg/config/config_test.go` (specifically the `TestLoad` cases). This is happening because my refactor moves the webhook initialization out of the `Load` path and into the new `AddWebhookSettingsTo` function. As a result, the `TestLoad` assertions are failing because they expect the `WebhookServer` to be populated, but it is now returning `nil`.

I want to make sure I follow the best practices for this project. Could you advise on your preferred approach to fix this?

pkg/config/config.go
```diff
@@ -65,19 +66,31 @@ func addTo(o *ctrl.Options, cfg *configapi.Configuration) {
...
-		o.WebhookServer = webhook.NewServer(wo)
+	// Webhook server is NOT set here anymore.
+	// It is set in AddWebhookSettingsTo() after TLS is parsed.
+}
+
+// AddWebhookSettingsTo builds and sets the WebhookServer on the options,
+// applying TLS configuration if provided.
+// This is separated from addTo() because TLS must be parsed first,
+// and TLS parsing happens after config is loaded.
+func AddWebhookSettingsTo(o *ctrl.Options, cfg *configapi.Configuration, tlsOpts []func(*cryptotls.Config)) {
```
Edwinhr716, line 77: I'm confused why this isn't an issue for JobSet. They didn't have to split this function. @kannon92 do you have more context here?

cmd/main.go
```diff
@@ -295,6 +305,14 @@ func apply(configFile string,
 		c.NextProtos = []string{"http/1.1"}
 	}
 
+	serverTLSOpts := []func(*tls.Config){disableHTTP2}
+
+	// 3. LWS Pattern: Build and apply parsed TLS configuration from config
+	if parsedTLSConfig != nil {
```
Edwinhr716, line 311: I see that this logic is behind a feature gate in JobSet. Why is that @kannon92? https://github.com/kubernetes-sigs/jobset/blob/main/main.go#L164

kannon92: I was following this from Kueue where the ask was to have it under a feature gate that was on by default.

pkg/config/config.go
kannon92: This idea may have come from Kueue. For Kueue I had to move some stuff around because the setting of these settings was done in three different areas of code so I had to refactor to get DRY.

pkg/config/config_test.go
```diff
@@ -180,12 +182,6 @@ webhook:
 		LeaseDuration:              ptr.To(defaultLeaderElectionLeaseDuration),
 		RenewDeadline:              ptr.To(defaultLeaderElectionRenewDeadline),
 		RetryPeriod:                ptr.To(defaultLeaderElectionRetryPeriod),
-		WebhookServer: &webhook.DefaultServer{
```
Edwinhr716, line 183: Can we test anywhere that the final ctrl options include the webhook server?

Mostafahassen1: Done

PR #790, opened 2026-03-23
Fix flaky "Test default progressing state" integration test

adinilfeld:
#### What type of PR is this?

/kind bug
/kind failing-test

Integration test case was failing consistently after #778 was merged, with error:
```
Operation cannot be fulfilled on leaderworkersets.leaderworkerset.x-k8s.io "test-sample": the object has been modified; please apply your changes to the latest version and try again
```
([Example failure](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_lws/781/pull-lws-test-integration-main/2035152984828022784).) The likely cause is that the additional LWS controller reconciliation loops (now also triggered by leader pod updates) were causing the LWS `resourceVersion` to update more quickly than the integration test's cache could keep up with. 

To fix, we use `Patch` instead of `Update`, so that the `resourceVersion` doesn't matter.

```release-note

```

### DIFF
--- a/test/integration/controllers/leaderworkerset_test.go
+++ b/test/integration/controllers/leaderworkerset_test.go
@@ -600,8 +600,9 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
 			updates: []*update{
 				{
 					lwsUpdateFn: func(lws *leaderworkerset.LeaderWorkerSet) {
+						patch := client.MergeFrom(lws.DeepCopy())
 						lws.Status.Conditions = []metav1.Condition{}
-						gomega.Eventually(k8sClient.Status().Update(ctx, lws), testing.Timeout, testing.Interval).Should(gomega.Succeed())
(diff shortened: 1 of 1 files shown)


test/integration/controllers/leaderworkerset_test.go
```diff
@@ -600,8 +600,9 @@ var _ = ginkgo.Describe("LeaderWorkerSet controller", func() {
 			updates: []*update{
 				{
 					lwsUpdateFn: func(lws *leaderworkerset.LeaderWorkerSet) {
+						patch := client.MergeFrom(lws.DeepCopy())
```
Edwinhr716, line 603: Shouldn't the deep copy happen after updating the conditions?

adinilfeld: My understanding is that it has to come before, so that there's a diff generated which can be sent to the API server. If the deep copy happened after, then the diff between `lws` and `patch` would be empty, so no change would be generated.

adinilfeld:
Closing, as #791 accomplishes the same thing.

PR #800, opened 2026-03-30
fix(volcano): add missing PodGroup RBAC for kubectl-based installs

janghyukjin:
#### What type of PR is this?
/kind bug
/kind documentation

When LWS is installed via `kubectl` with Volcano gang scheduling enabled, the LWS controller automatically creates a `scheduling.volcano.sh/podgroups` resource for each LWS replica before pod scheduling. However, `config/rbac/role.yaml` is missing the RBAC permission for this resource, causing PodGroup creation to silently fail.

As a result, Volcano's admission webhook (`validatepod.volcano.sh`) denies pod creation because the expected PodGroup does not exist:

```
Warning  FailedCreate  statefulset-controller
  create Pod <name> in StatefulSet <name> failed:
  admission webhook "validatepod.volcano.sh" denied the request:
  failed to get PodGroup for pod <namespace/pod-name>:
  podgroups.scheduling.volcano.sh "<podgroup-name>" not found
```

Root cause confirmed via:
```sh
kubectl auth can-i create podgroups.scheduling.volcano.sh \
  --as=system:serviceaccount:lws-system:lws-controller-manager
# → no
```

The Helm chart already handles this correctly (see [`charts/lws/templates/rbac/clusterrole.yaml#L132-L142`](https://github.com/kubernetes-sigs/lws/blob/bcd37f46553449c14c69c772a586f7db05a8300d/charts/lws/templates/rbac/clusterrole.yaml#L132-L142)) with a conditional rule:
```yaml
{{- if eq .Values.gangSchedulingManagement.schedulerProvider "volcano" }}
- apiGroups: ["scheduling.volcano.sh"]
  resources: ["podgroups"]
  verbs: ["create", "get", "list", "watch"]
{{- end }}
```

This PR brings parity for `kubectl`-based installs by:
- Adding `config/components/volcano/` kustomize component that patches the existing `manager-role` with the required Volcano PodGroup rules (following the existing `config/components/prometheus/` pattern)
- Documenting the required RBAC step in the gang-scheduling README for `kubectl` users

N/A

This RBAC is intentionally scoped to `scheduling.volcano.sh/podgroups` only — it is Volcano-specific and not needed for other gang scheduler integrations (e.g. Coscheduler uses `scheduling.sigs.k8s.io/podgroup`). Placing it under `config/components/` rather than `config/rbac/` keeps it opt-in, consistent with how the Helm chart conditionally includes it.

```release-note
Add `config/components/volcano` kustomize component with Volcano PodGroup RBAC for kubectl-based LWS installations with gang scheduling enabled.
```

### DIFF
--- a/config/components/volcano/kustomization.yaml
+++ b/config/components/volcano/kustomization.yaml
@@ -4,5 +4,7 @@ kind: Component
 patches:
 - path: manager_role_patch.yaml
   target:
+    group: rbac.authorization.k8s.io
+    version: v1
     kind: ClusterRole
     name: manager-role
--- a/config/components/volcano/manager_role_patch.yaml
+++ b/config/components/volcano/manager_role_patch.yaml
@@ -1,14 +1,12 @@
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: manager-role
-rules:
-- apiGroups:
-  - scheduling.volcano.sh
-  resources:
-  - podgroups
-  verbs:
-  - create
-  - get
-  - list
-  - watch
+- op: add
+  path: /rules/-
+  value:
+    apiGroups:
+    - scheduling.volcano.sh
+    resources:
+    - podgroups
+    verbs:
+    - create
+    - get
+    - list
+    - watch
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -28,6 +28,10 @@ resources:
 # [METRICS] Expose the controller manager metrics service.
 - manager_metrics_service.yaml
 
+# [VOLCANO] To enable Volcano gang scheduling RBAC, uncomment the following.
+# components:
+# - ../components/volcano
+
 patches:
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
 # crd/kustomization.yaml
--- a/docs/examples/sample/gang-scheduling/README.md
+++ b/docs/examples/sample/gang-scheduling/README.md
@@ -52,9 +52,15 @@ To enable gang scheduling, you must enable the feature flag and specify `volcano
            schedulerProvider: volcano
      ```
 
-  2. **Apply the Volcano RBAC component** to grant the LWS controller permission to manage Volcano PodGroups:
+  2. **Enable the Volcano RBAC component** to grant the LWS controller permission to manage Volcano PodGroups. In `config/default/kustomization.yaml`, uncomment the `components` section:
+     ```yaml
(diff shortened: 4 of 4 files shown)


Edwinhr716:
Please sign the CLA authorization

Edwinhr716:
@JesseStutler is this needed?

janghyukjin:
@Edwinhr716 Thanks for the reminder! I’ve completed the CLA signing.
Please let me know if there’s anything else I should address 🙂

yankay:
Thanks @janghyukjin for this contribution! The bug analysis is solid and the need is clear — kubectl installs are missing Volcano PodGroup RBAC while Helm already handles it.

Two issues to address:

**1. `kubectl apply -k config/components/volcano` doesn't work**

`kind: Component` produces no resources on its own — `kustomize build` outputs 0 bytes, and `kubectl apply` fails with "no objects passed to apply". Components must be included via the `components:` field from a parent Kustomization.

Suggest: add a commented-out entry in `config/default/kustomization.yaml` (similar to the prometheus pattern) and update the README accordingly:

```yaml
# [VOLCANO] To enable Volcano gang scheduling RBAC, uncomment the following.
#components:
#- ../components/volcano
```

**2. Strategic merge patch replaces all rules instead of appending**

`ClusterRole.rules` is an atomic list (no merge key), so the SMP patch replaces the entire rules array — the controller loses all other RBAC permissions. See https://github.com/kubernetes-sigs/kustomize/issues/1523 for the same scenario.

Suggest: use a JSON 6902 patch to append:

```yaml
# manager_role_patch.yaml
- op: add
  path: /rules/-
  value:
    apiGroups:
    - scheduling.volcano.sh
    resources:
    - podgroups
    verbs:
    - create
    - get
    - list
    - watch
```

RBAC verbs (create/get/list/watch) are correct and match the Helm chart. Nice work keeping this opt-in under `config/components/`!

janghyukjin:
Thanks for the thorough review, @yankay!

Both issues are now fixed:

**1. `kubectl apply -k config/components/volcano` → parent kustomization approach**
Added a commented `[VOLCANO]` entry in `config/default/kustomization.yaml` following the same pattern as prometheus. Updated the README to instruct users to uncomment the `components:` section and run `kubectl apply -k config/default`.

**2. SMP replacing all rules → JSON 6902 patch**
Switched `manager_role_patch.yaml` to use a JSON 6902 `op: add, path: /rules/-` patch so the new rule is appended instead of replacing the entire rules array.

I also verified locally with `kustomize build` that the existing rules are preserved and the volcano podgroup rule is appended at the end.

JesseStutler:
> @JesseStutler is this needed?

Yes Thanks @janghyukjin for contributing, I didn't use `kubectl -k` before 👍

PR #801, opened 2026-03-31
perf(pod): optimize env var lookup

ErikJiang:
#### What type of PR is this?

/kind cleanup

Optimize the `addEnvVarsIfNotExists` function to use map-based lookup instead of nested loops.

Fixes #

```release-note
NONE
```


yankay:
Nice cleanup! Order semantics preserved and CI is green.

/lgtm

PR #805, opened 2026-04-06
fix: prevent leader pod from being recreated twice on background deletion

### LINKED ISSUE #804: Leader pod recreate twice when the restart policy is RecreateGroupOnPodRestart
**What happened**:
1、Create LWS resource
2、Deleted leader pod vllm-0 with background propagation
3、The leader pod vllm-0 was recreated twice and the worker pod vllm-0-1 was recreated once
```sh
$ kubectl get pod -w
NAME       READY   STATUS    RESTARTS   AGE
vllm-0     1/1     Running   0          7m26s
vllm-0-1   1/1     Running   0          7m26s

vllm-0     1/1     Terminating   0          8m9s
vllm-0     0/1     Completed     0          8m9s
vllm-0     0/1     Completed     0          8m10s
vllm-0     0/1     Completed     0          8m10s
vllm-0-1   1/1     Terminating   0          8m10s
vllm-0     0/1     Pending       0          0s
vllm-0     0/1     Pending       0          0s
vllm-0     0/1     ContainerCreating   0          0s
vllm-0-1   0/1     Completed           0          8m10s
vllm-0     0/1     Terminating         0          0s
vllm-0     0/1     Terminating         0          0s
vllm-0-1   0/1     Completed           0          8m11s
vllm-0-1   0/1     Completed           0          8m11s
vllm-0     1/1     Terminating         0          2s
vllm-0     0/1     Completed           0          2s
vllm-0     0/1     Completed           0          3s
vllm-0     0/1     Completed           0          3s
vllm-0     0/1     Pending             0          0s
vllm-0     0/1     Pending             0          0s
vllm-0     0/1     ContainerCreating   0          0s
vllm-0-1   0/1     Pending             0          0s
vllm-0-1   0/1     Pending             0          0s
vllm-0-1   0/1     ContainerCreating   0          0s
vllm-0-1   1/1     Running             0          2s
vllm-0     1/1     Running             0          2s
```
4、Attempted to delete the worker pod with background propagation and the leader pod with foreground propagation, but neither of these actions caused pod vllm-0 to be recreated twice.

When background deletion propagation is used to delete the leader pod, the leader pod should be rebuilt only once.

1、Create LWS with YAML above
```yaml
kind: LeaderWorkerSet
metadata:
  name: vllm
spec:
  replicas: 1
  leaderWorkerTemplate:    
    size: 2
    restartPolicy: RecreateGroupOnPodRestart
    workerTemplate:
      spec:
        imagePullSecrets:
          - name: default-secret
        containers:
          - name: vllm-worker
            image: hello-world
            resources:
              limits:
                cpu: 30m
                memory: 100Mi
              requests:
                cpu: 30m
                memory: 100Mi
```
2、Delete the leader pod vllm-0 using kubectl
```sh
$ kubectl delete pod vllm-0
```

I tested this with both release-0.7.0 and release-0.8.0, and the issue occurred in both versions.

**Environment**:

- Kubernetes version (use `kubectl version`): v1.34.0
- LWS version (use `git describe --tags --dirty --always`): release-0.8.0
- Cloud provider or hardware configuration: Kind
- OS (e.g: `cat /etc/os-release`): Ubuntu 20.04
- Kernel (e.g. `uname -a`): Linux Y30063157-hEQN4 5.15.167.4-microsoft-standard-WSL2 #1 SMP Tue Nov 5 00:21:55 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
- Install tools:
- Others:

yankay:

When a leader pod is deleted with background propagation (the default for `kubectl delete pod`) and the restart policy is `RecreateGroupOnPodRestart`, the leader pod gets recreated **twice**. This happens because during garbage collection of the old group, the terminating worker pods trigger `handleRestartPolicy` which looks up the leader pod by name — finding the **newly recreated** leader instead of the old one. Since only the revision key was checked (not the ownership chain), the stale worker pod was mistakenly treated as belonging to the new leader, causing the new leader to be deleted and recreated a second time.

This PR adds `workerPodBelongsToLeader` to verify that the worker pod's owner reference chain (either direct Pod owner or StatefulSet → Pod) actually matches the current leader's UID before triggering group recreation.

Fixes https://github.com/kubernetes-sigs/lws/issues/804

## Special notes for your reviewer

The race condition only manifests with **background** deletion propagation. Foreground propagation blocks until all dependents are gone, so the StatefulSet does not recreate the leader until the old worker pods are fully cleaned up.

```release-note
Fix leader pod being recreated twice when deleted with background propagation and RecreateGroupOnPodRestart restart policy.
```

/kind bug

### DIFF
--- a/pkg/controllers/pod_controller_test.go
+++ b/pkg/controllers/pod_controller_test.go
@@ -23,13 +23,16 @@ import (
 	"github.com/google/go-cmp/cmp"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	appsapplyv1 "k8s.io/client-go/applyconfigurations/apps/v1"
 	coreapplyv1 "k8s.io/client-go/applyconfigurations/core/v1"
 	metaapplyv1 "k8s.io/client-go/applyconfigurations/meta/v1"
 	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	leaderworkerset "sigs.k8s.io/lws/api/leaderworkerset/v1"
 	revisionutils "sigs.k8s.io/lws/pkg/utils/revision"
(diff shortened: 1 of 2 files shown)


pkg/controllers/pod_controller_test.go
```diff
@@ -420,3 +423,154 @@ func TestConstructWorkerStatefulSetApplyConfiguration(t *testing.T) {
 		})
 	}
 }
+
+func TestHandleRestartPolicyUsesCurrentWorkerOwnership(t *testing.T) {
+	lws := wrappers.BuildLeaderWorkerSet("default").Replica(1).Size(2).RestartPolicy(leaderworkerset.RecreateGroupOnPodRestart).Obj()
+	revisionKey := "revision-1"
+
+	makeLeader := func(uid types.UID) *corev1.Pod {
```
Edwinhr716: nit: makeLeaderPod

pkg/controllers/pod_controller_test.go
```diff
@@ -420,3 +423,154 @@ func TestConstructWorkerStatefulSetApplyConfiguration(t *testing.T) {
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:            leader.Name,
+				Namespace:       leader.Namespace,
+				UID:             uid,
+				OwnerReferences: []metav1.OwnerReference{*metav1.NewControllerRef(leader, corev1.SchemeGroupVersion.WithKind("Pod"))},
+			},
+		}
+	}
+
+	makeWorker := func(owner metav1.OwnerReference) *corev1.Pod {
```
Edwinhr716: nit: makeWorkerPod

pkg/controllers/pod_controller_test.go
yankay: Done, renamed to `makeLeaderPod`. Thanks!

pkg/controllers/pod_controller_test.go
yankay: Done, renamed to `makeWorkerPod`. Thanks!

Issue #789, opened 2026-03-23
Merge the LWS and DisaggregatedSet binaries


Edwinhr716:
**What would you like to be added**: Similar to how `kube-controller-manager` works, we should combine the lws and disaggregatedSet binaries so that users only have to deploy one binary to use both.

**Why is this needed**: Simplify the usage of LWS and DisaggregatedSet

**Completion requirements**: 
- Move the disaggregatedSet code to be under the corresponding lws folders (e.g api/ will contain both the LWS API definition and the DisaggregatedSet definition).

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

Edwinhr716:
cc @ahg-g @hasB4K

hasB4K:
Here are the steps to achieve this:
- [x] A PR that moves the API, etc. as an empty controller
  - https://github.com/kubernetes-sigs/lws/pull/815
- [x] A PR with the core logic only and the unit tests. This one will be the most important one to review, and the separation with the other one is to avoid the clutter during the review process
- [x] A PR that adds back the e2e test (i made a few choices that I like, but maybe you won't, I'm not sure)
- [x] A PR that that combines the disaggregatedset and lws controllers into one (It would be two independent controller loops, just bundled in the same binary)
- [x] A PR that add the helm chart 
- [x] A PR that update the doc
- [x] A PR that remove the disaggregatedset folder

PR #829, opened 2026-04-29
chore: reduce dependabot PR noise via grouping

yankay:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

Reduce weekly dependabot PR noise by adding `sigs.k8s.io/*` to the existing `kubernetes` group, modeled after [jobset's config](https://github.com/kubernetes-sigs/jobset/blob/main/.github/dependabot.yml). Packages like `controller-runtime` and `structured-merge-diff` will land in the existing weekly kubernetes group PR instead of one PR each.

#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for your reviewer:

An earlier revision also added `/disaggregatedset` via `directories:`. Dropped per [reviewer feedback](https://github.com/kubernetes-sigs/lws/pull/829#discussion_r3162389430) — that directory will go away once #789 lands, so the extra config isn't needed.

```release-note
NONE
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,7 @@ updates:
       kubernetes:
         patterns:
           - "k8s.io/*"
+          - "sigs.k8s.io/*"
     ignore:
       # Ignore major and minor versions for dependencies updates
       # Allow patches and security updates.


.github/dependabot.yml
```diff
@@ -2,7 +2,9 @@ version: 2
 updates:
   # Maintain dependencies for go
   - package-ecosystem: "gomod"
-    directory: "/"
+    directories:
+      - "/"
+      - "/disaggregatedset"
```
Edwinhr716: We will remove this dir once #789 is completed, so this isn't needed.

yankay: Good point, thanks! Dropped the `/disaggregatedset` part — kept only the `sigs.k8s.io/*` grouping change. PTAL.

PR #836, opened 2026-04-30
[DisaggregatedSet][2/N] Add controller implementation with rolling update support

hasB4K:

  This PR adds the full controller implementation for DisaggregatedSet, following the API types and empty controller stub merged in #815.

#### What's included:

* Controller reconciler with stateless rolling update detection, it creates and manages LeaderWorkerSet resources for each role
* Includes two small patch that were not in the `disaggregatedset` folder (see https://github.com/kubernetes-sigs/lws/pull/833):
    * Rolling update executor with coordinated drain (prevents orphan workloads across roles)
    * newest-first drain ordering for multi-revision scenarios (A→B→C)
* Stateless planner for N-role rolling updates with surge/unavailable constraints
* LeaderWorkerSet manager for workload CRUD, initial-replicas annotation tracking, and revision grouping
* Service manager for headless portless Services enabling EndpointSlice-based discovery with cross-role readiness coordination
* Comprehensive unit tests (~4100 lines) covering planner step computation, executor drain ordering, A→B→C scenarios, workload/service managers
* E2E tests (18 tests on Kind) covering basic deployment, rolling updates, coordinated drain, scaling, service creation, labels/annotations propagation, N-role support, role add/remove/rename, and mid-rollout A→B→C newest-first
  drain verification

#### What type of PR is this?

/kind feature

Linked to #789

* I decided to *not break* this PR, because *we need to have unit tests and e2e tests* when changing/debugging the behavior of `disaggregatedset` (cc @Edwinhr716, @ahg-g and @yankay)
* Those tests represent some battle testing we faced recently for some of our Deployment
* I'm wondering if everything is missing to combines the disaggregatedset and lws controllers into one (or if everything is there)
* The next PRs will be:
  * documentation update.
  * helm chart update.
  * clean the old disaggregatedset folder.

* Users can now create DisaggregatedSet resources that manage multiple LeaderWorkerSet roles with coordinated rolling updates.


config/manager/controller_manager_config.yaml
```diff
@@ -2,5 +2,3 @@ apiVersion: config.lws.x-k8s.io/v1alpha1
 kind: Configuration
 leaderElection:
   leaderElect: true
```
yankay, line 4: Looks like a rebase leftover unrelated to this PR — removing `internalCertManagement.enable: true` will break all LWS webhooks since no certs will be provisioned; could you restore it?

config/manager/kustomization.yaml
```diff
@@ -13,5 +13,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: us-central1-docker.pkg.dev/k8s-staging-images/lws/lws
+  newName: us-central1-docker.pkg.dev/k8s-staging-images/lws
```
yankay: Looks like a rebase leftover unrelated to this PR — the image path should stay as `…/k8s-staging-images/lws/lws` per the release convention (`cloudbuild.yaml`, `NEW_RELEASE.md`); could you revert this?

yankay (commented):
A couple of likely rebase leftovers — otherwise looks great, will do a deeper pass next.

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -0,0 +1,425 @@
...
+	roleIndex := make(map[string]int, len(allRoleNames))
+	for i, name := range allRoleNames {
+		roleIndex[name] = i
+	}
+
+	for _, role := range ds.Spec.Roles {
+		if rc := role.Spec.RolloutStrategy.RollingUpdateConfiguration; rc != nil {
+			i := roleIndex[role.Name]
+			surge := rc.MaxSurge.IntValue()
+			unavail := rc.MaxUnavailable.IntValue()
```
yankay: `IntValue()` returns `0` for percentage strings like `"25%"` (it falls back to `strconv.Atoi` and swallows the error). Since `MaxSurge` / `MaxUnavailable` are `intstr.IntOrString` and percentages aren't rejected by the webhook, a user-supplied `maxUnavailable: 25%` is silently parsed as `0`, the `if unavail > 0 ... else if surge > 0` branch is skipped, and the role falls back to the default `{MaxSurge: 1, MaxUnavailable: 0}` — the user's intent is lost without any warning.

The rest of the codebase uses `intstr.GetScaledValueFromIntOrPercent` for this — see `leaderworkerset_controller.go:295,299` and `leaderworkerset_webhook.go:163,167`. Same pattern in k/k Deployment's `ResolveFenceposts`.

```go
total := getTargetReplicas(ds, role.Name)
surge, err := intstr.GetScaledValueFromIntOrPercent(&rc.MaxSurge, total, true)
unavail, err := intstr.GetScaledValueFromIntOrPercent(&rc.MaxUnavailable, total, false)
```

Worth adding percentage cases to the planner/executor tests — current tests only cover integers, which is why CI missed this.

yankay:
CI `pull-lws-verify-main` failed because `charts/lws/crds/leaderworkerset.x-k8s.io_leaderworkersets.yaml` is out of date — running `make generate` locally regenerates it (the `crds` sub-target rebuilds the chart CRD via `kustomize` + `yq`). Please run `make generate` and commit the regenerated file.

config/manager/controller_manager_config.yaml
hasB4K: Fixed :)

config/manager/kustomization.yaml
hasB4K: Fixed :)

pkg/controllers/disaggregatedset/executor.go
hasB4K: Okay, it's an easy fix, but I can't do it now. I will do it on Monday (and I'll add a test)

hasB4K: And added!

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -18,22 +18,29 @@ package disaggregatedset
...
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/events"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/log"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	leaderworkerset "sigs.k8s.io/lws/api/leaderworkerset/v1"
 
 	disaggregatedsetv1 "sigs.k8s.io/lws/api/disaggregatedset/v1"
```
Edwinhr716, line 34: Can this be changed to `disaggregatedset`? That way it is consistent with the `leaderworkerset` import

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -44,20 +51,226 @@ type DisaggregatedSetReconciler struct {
 // +kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
 
-// Reconcile is the main reconciliation loop for DisaggregatedSet.
-// This is a stub implementation that will be replaced with the full controller logic.
-func (r *DisaggregatedSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	logger := log.FromContext(ctx)
-	logger.V(2).Info("Reconciling DisaggregatedSet", "name", req.Name, "namespace", req.Namespace)
+func (reconciler *DisaggregatedSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
Edwinhr716: nit: can this be r to match what LWS does?

pkg/controllers/disaggregatedset/workload_manager.go
```diff
@@ -0,0 +1,423 @@
...
+
+func (manager *LeaderWorkerSetManager) Create(ctx context.Context, params CreateParams) error {
+	workloadName := GenerateName(params.DisaggregatedSet.Name, params.Role, params.Revision)
+	replicas := int32(params.Replicas)
+	config := params.Config
+
+	workerLabels := mergeLabels(config.Spec.LeaderWorkerTemplate.WorkerTemplate.Labels, params.Labels)
+	workerAnnotations := copyAnnotations(config.Spec.LeaderWorkerTemplate.WorkerTemplate.Annotations)
+
+	lwsTemplate := leaderworkerset.LeaderWorkerTemplate{
```
Edwinhr716: Do we still need this operation if we have config.Spec.LeaderWorkerTemplate?

Edwinhr716: ditto for all operations that access `config.Spec.LeaderWorkerTemplate`

pkg/controllers/disaggregatedset/utils.go
```diff
@@ -0,0 +1,174 @@
...
+const NumRequiredRoles = 2
+
+const (
+	LabelDisaggRole = "disaggregatedset.x-k8s.io/role"
+	LabelDisaggName = "disaggregatedset.x-k8s.io/name"
+	LabelRevision   = "disaggregatedset.x-k8s.io/revision"
+)
+
+const (
+	AnnotationInitialReplicas = "disaggregatedset.x-k8s.io/initial-replicas"
```
Edwinhr716: Can we move this to the API file, similar to how LWS does it?

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
hasB4K: Fixed!

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
hasB4K: Fixed!

pkg/controllers/disaggregatedset/workload_manager.go
hasB4K: Fixed!

pkg/controllers/disaggregatedset/utils.go
hasB4K: Fixed!

cmd/main.go
```diff
@@ -38,7 +38,7 @@ import (
 	volcanov1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
 
 	configapi "sigs.k8s.io/lws/api/config/v1alpha1"
-	disaggregatedsetv1 "sigs.k8s.io/lws/api/disaggregatedset/v1"
+	disaggregatedset "sigs.k8s.io/lws/api/disaggregatedset/v1"
 	leaderworkersetv1 "sigs.k8s.io/lws/api/leaderworkerset/v1"
```
Edwinhr716, line 42: change to `leaderworkerset`

pkg/utils/disaggregatedset/utils.go
```diff
@@ -0,0 +1,164 @@
+/*
```
Edwinhr716, line 1: Can we move this file to the /util/ directory and have it in it's own package? There are some names that overlap (e.g SetInitialReplicas) and it becomes harder to track which one is being referenced. 

For example, WorkloadManager has a function called `SetInitialReplicas` and calls the util `SetInitialReplicas`.

pkg/controllers/disaggregatedset/workload_manager.go
```diff
@@ -0,0 +1,401 @@
...
+}
+
+func getLWSReplicas(leaderWorkerSet *leaderworkerset.LeaderWorkerSet) int32 {
+	if leaderWorkerSet.Spec.Replicas == nil {
+		return 1
+	}
+	return *leaderWorkerSet.Spec.Replicas
+}
+
+func groupWorkloadsByRevision(workloads []WorkloadInfo) GroupedWorkloads {
```
Edwinhr716: For functions that get used by different files (this one is used by disaggregatedset_controller and workload_manager), can we move them to a util file?

pkg/controllers/disaggregatedset/disaggregatedset_controller_test.go
```diff
@@ -0,0 +1,189 @@
...
+	scheme := runtime.NewScheme()
+	_ = disaggregatedset.AddToScheme(scheme)
+	_ = corev1.AddToScheme(scheme)
+	_ = leaderworkerset.AddToScheme(scheme)
+	return scheme
+}
+
+// createOldLeaderWorkerSet creates a LeaderWorkerSet representing an existing workload with the given revision.
+// Useful for simulating pre-existing workloads in rolling update tests.
+func createOldLeaderWorkerSet(disaggregatedSet *disaggregatedset.DisaggregatedSet, role, revision string, replicas int32) *leaderworkerset.LeaderWorkerSet {
```
Edwinhr716: I see a lot of functions across different test files which create LWS and disaggregatedSet objects. Can we create a wrapper class like we have for LWS objects https://github.com/kubernetes-sigs/lws/blob/main/test/wrappers/wrappers.go, or add functions that we need to existing wrappers?

pkg/controllers/disaggregatedset/workload_manager.go
```diff
@@ -0,0 +1,401 @@
...
+				Name:       params.DisaggregatedSet.Name,
+				UID:        params.DisaggregatedSet.UID,
+				Controller: ptr.To(true),
+			}},
+		},
+		Spec: leaderworkerset.LeaderWorkerSetSpec{
+			Replicas:             &replicas,
+			StartupPolicy:        config.Spec.StartupPolicy,
+			LeaderWorkerTemplate: lwsTemplate,
+			NetworkConfig:        lwsNetworkConfig,
```
Edwinhr716: Is there a reason why we pass `lwsTemplate`, `lwsNetworkConfig`, and `replicas` manually instead of just passing config.Spec here?

pkg/controllers/disaggregatedset/workload_manager.go
```diff
@@ -0,0 +1,401 @@
...
+	return maps.Clone(annotations)
+}
+
+func (manager *LeaderWorkerSetManager) Create(ctx context.Context, params CreateParams) error {
+	workloadName := GenerateName(params.DisaggregatedSet.Name, params.Role, params.Revision)
+	replicas := int32(params.Replicas)
+	config := params.Config
+
+	lwsTemplate := config.Spec.LeaderWorkerTemplate
+	lwsTemplate.WorkerTemplate.Labels = mergeLabels(config.Spec.LeaderWorkerTemplate.WorkerTemplate.Labels, params.Labels)
```
Edwinhr716: Before we needed to merge the labels because we were not passing `leaderWorkerSetTemplateSpec` directly. Do we still need this?  ditto with the LeaderTemplate merger.

cmd/main.go
Edwinhr716: actually, better to change it back to `disaggregatedsetv1`

pkg/controllers/disaggregatedset/workload_manager.go
hasB4K: The merge is still needed. `mergeLabels` injects system labels (`disaggregatedset.x-k8s.io/role`, `/name`, `/revision`) onto the pod templates, which aren't part of the user's spec. Services use these labels as selectors 😅 .

`copyAnnotations` is also needed because the struct copy is shallow (shares the underlying map), so without it we'd mutate the original config. I'll add a short comment in the code to clarify.

Edwinhr716: > Services use these labels as selectors 😅 .

Ah ok, makes sense.

pkg/utils/disaggregatedset/utils.go
hasB4K: done: https://github.com/kubernetes-sigs/lws/pull/836/changes/bc135f68eda825baa91a9e4ce225274396d77884#diff-d4ef8f2c640ea0cc9fca2d5bda8c8fe83c319b5ec39d4124aa83657d465b8e86

test/wrappers/disaggregatedset_wrappers.go
```diff
@@ -0,0 +1,180 @@
...
+				Labels: map[string]string{
+					disaggregatedsetv1.RoleLabelKey:     role,
+					disaggregatedsetv1.SetNameLabelKey:  name,
+					disaggregatedsetv1.RevisionLabelKey: revision,
+				},
+			},
+		},
+	}
+}
+
```
Edwinhr716: These lwsWrapper functions should be added to wrappers.go, not here

test/wrappers/disaggregatedset_wrappers.go
```diff
@@ -0,0 +1,180 @@
...
+			break
+		}
+	}
+	return w
+}
+
+// --- LWS builder for disaggregatedset tests ---
+
+func BuildDisaggregatedSetLWS(name, namespace, role, revision string) *LeaderWorkerSetWrapper {
+	return &LeaderWorkerSetWrapper{
```
Edwinhr716: Why is this function needed instead of calling `BuildBasicLeaderWorkerSet.Labels()`

pkg/controllers/disaggregatedset/executor_test.go
```diff
@@ -0,0 +1,1297 @@
...
+func newTestExecutor(fakeClient client.Client) *RollingUpdateExecutor {
+	return &RollingUpdateExecutor{
+		Client:          fakeClient,
+		WorkloadManager: NewLeaderWorkerSetManager(fakeClient),
+		Record:          events.NewFakeRecorder(100),
+	}
+}
+
+// createTestLWS creates a LeaderWorkerSet for testing.
+func createTestLWS(
```
Edwinhr716: I think it is easier to read the code and better for future proof if we remove these createTest<> functions and call the wrappers functions directly instead.

test/testutils/disaggregatedset/utils.go
```diff
@@ -0,0 +1,154 @@
+/*
```
Edwinhr716, line 1: I'm guessing the functions here won't be needed once we merge the LWS and DisaggregatedSet binaries? If so, can we create an issue to track it?

pkg/utils/disaggregatedset/utils.go
```diff
@@ -0,0 +1,206 @@
...
+			}
+		}
+
+		state[role] += replicas
+	}
+
+	return state
+}
+
+type WorkloadInfo struct {
```
Edwinhr716: For this, why not store the LWS directly? It feels like unnecessary abstraction to me, and makes it harder to read since this is yet another struct that I have to track down to see what it represents

Edwinhr716:
Something that makes the code harder to read is the usage of `Workload` instead of LWS. I think it makes sense to use LWS directly in the name instead, since whoever reads the code is expected to know that DisaggregatedSet uses LWS under the hood. 

There's also some instances where Workload can refer to a disaggregatedSet unit (e.g a variable called NewWorkload is of type of GroupedWorkload, which is a struct that represents the LeaderWorkerSets that compose a DisaggregatedSet)

We use the Sts name directly in the LWS codebase, so it would be consistent with that as well.

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -0,0 +1,430 @@
...
+
+func buildPlannerState(
+	ds *disaggregatedsetv1.DisaggregatedSet,
+	allRoleNames []string,
+	specRoleSet map[string]bool,
+	oldWorkloads disaggregatedsetutils.GroupedWorkloads,
+	newWorkload disaggregatedsetutils.GroupedWorkload,
+) (source, currentOld, currentNew, targetNew RoleReplicaState) {
+	n := len(allRoleNames)
+	source, currentOld, currentNew, targetNew = make(RoleReplicaState, n), make(RoleReplicaState, n), make(RoleReplicaState, n), make(RoleReplicaState, n)
```
Edwinhr716: instead of source, what about initialOld to make it consistent with the other variables?

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -0,0 +1,430 @@
...
+	specRoleSet, oldRoleSet := buildRoleSets(specRoleNames, oldWorkloads)
+
+	allRoleNames := append(slices.Clone(specRoleNames), removedRoles(oldRoleSet, specRoleSet)...)
+
+	if !isWorkloadStable(newWorkload, specRoleNames) {
+		log.V(1).Info("Waiting for new workload to stabilize")
+		return ctrl.Result{RequeueAfter: time.Second}, nil
+	}
+
+	source, currentOld, currentNew, targetNew := buildPlannerState(disaggregatedSet, allRoleNames, specRoleSet, oldWorkloads, newWorkload)
```
Edwinhr716: ditto, source -> initialOld

pkg/controllers/disaggregatedset/planner.go
```diff
@@ -0,0 +1,384 @@
...
+	}
+
+	if isNewAtTarget(currentNew, targetNew) {
+		return &UpdateStep{Past: make([]int, len(source)), New: currentNew}
+	}
+
+	nextNew := computeNextNewReplicas(targetNew, currentNew, totalSteps)
+	minOld := computeMinOld(source, currentNew, targetNew, config)
+
+	if step := tryScaleUp(currentOld, currentNew, nextNew, targetNew, config); step != nil {
```
Edwinhr716, line 341: So here, we scale up to nextNew first, then it takes another reconcilation loop to scale down to minOld? Why not do it in one go?

pkg/controllers/disaggregatedset/planner.go
```diff
@@ -0,0 +1,384 @@
...
+		return step
+	}
+	if step := tryForceDrain(currentOld, nextNew, source, targetNew, config); step != nil {
+		return step
+	}
+
+	return nil
+}
+
+func ComputeAllSteps(source, target RoleReplicaState, config []RollingUpdateConfig) []UpdateStep {
```
Edwinhr716: Is this called anywhere besides tests?

hasB4K: It's called in the hack script planner. I can remove this, since this hack script is not present in this PR.

pkg/controllers/disaggregatedset/planner.go
hasB4K: The scale-up and scale-down are intentionally split into separate reconciliation loops. The idea is: scale up new replicas first, wait for them to be Ready, then drain old ones. This way we never drop below the availability guarantee (targetNew - maxUnavailable).

If we did both in one step, we'd be removing old replicas before the new ones are actually serving traffic, which means potential downtime during the transition.

That said, when maxSurge is tight and we can't scale up without draining first, tryForceDrain already does both in a single step. So the planner does combine them when it has to, it just prefers not to.

hasB4K: That being said, I'm not happy with the current planner, I want to have a "percent-based" planner, the current one is not perfect sometimes. We can discuss this offline, I want this to be a follow-up PR.

api/disaggregatedset/v1/disaggregatedset_types.go
```diff
@@ -21,6 +21,20 @@ import (
 	leaderworkerset "sigs.k8s.io/lws/api/leaderworkerset/v1"
 )
 
+const (
+	// SetNameLabelKey records the DisaggregatedSet name that resources belong to.
+	SetNameLabelKey string = "disaggregatedset.x-k8s.io/name"
```
ahg-g, line 27: nit: indicate that the namespace is the same as the one where the resources are created .

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -44,20 +52,241 @@ type DisaggregatedSetReconciler struct {
...
+					return fmt.Errorf("failed to delete old LWS %s: %w", lws.Name, err)
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+func (r *DisaggregatedSetReconciler) cleanupDrainedLWS(ctx context.Context, disaggregatedSet *disaggregatedsetv1.DisaggregatedSet, revision string) error {
```
ahg-g, line 195: add a comment to summarize what this function does at the high level pls

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -44,20 +52,241 @@ type DisaggregatedSetReconciler struct {
...
+	log := logf.FromContext(ctx)
+
+	lwsList, err := r.LWSManager.List(ctx, disaggregatedSet.Namespace, disaggregatedSet.Name, "")
+	if err != nil {
+		return fmt.Errorf("failed to list LWS for cleanup: %w", err)
+	}
+
+	revisionReplicas := make(map[string]map[string]int)
```
ahg-g, line 205: pls document the assumed key and values of this map of maps

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -44,20 +52,241 @@ type DisaggregatedSetReconciler struct {
...
+		}
+		if revisionReplicas[lwsRevision] == nil {
+			revisionReplicas[lwsRevision] = make(map[string]int)
+		}
+		lwsRole := lws.Labels[disaggregatedsetv1.RoleLabelKey]
+		lwsReplicas := 0
+		if lws.Spec.Replicas != nil {
+			lwsReplicas = int(*lws.Spec.Replicas)
+		}
+		revisionReplicas[lwsRevision][lwsRole] = lwsReplicas
```
ahg-g, line 223: I assume the expectation here is that only one lws per role per revision should exist, since this can't be guaranteed, log a warning when this happens.

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -44,20 +52,241 @@ type DisaggregatedSetReconciler struct {
...
+
+	disaggregatedSet := &disaggregatedsetv1.DisaggregatedSet{}
+	if err := r.Get(ctx, req.NamespacedName, disaggregatedSet); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	log.Info("Reconciling DisaggregatedSet", "name", disaggregatedSet.Name, "namespace", disaggregatedSet.Namespace)
```
ahg-g, line 66: pls add comments to explain the steps for a single reconcile call. This makes it easier to review the code and for future maintainability.

pkg/controllers/disaggregatedset/lws_manager.go
```diff
@@ -0,0 +1,324 @@
...
+}
+
+func getLWSReplicas(leaderWorkerSet *leaderworkerset.LeaderWorkerSet) int32 {
+	if leaderWorkerSet.Spec.Replicas == nil {
+		return 1
+	}
+	return *leaderWorkerSet.Spec.Replicas
+}
+
+func (manager *LeaderWorkerSetManager) GetRevisionRolesList(
```
ahg-g, line 209: pls document the semantics of this function (primarily what it tries to calculate and returns)

pkg/controllers/disaggregatedset/lws_manager.go
```diff
@@ -0,0 +1,324 @@
...
+			oldLWS = append(oldLWS, lws)
+		}
+	}
+
+	oldRevisions := disaggregatedsetutils.GroupByRevision(oldLWS)
+	newGrouped := disaggregatedsetutils.GroupByRevision(newLWS)
+
+	var newRevision *disaggregatedsetutils.RevisionRoles
+	if len(newGrouped) > 0 {
+		newRevision = &newGrouped[0]
```
ahg-g, line 236: what assumptions are we making by taking the item of index 0? pls document this assumption

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -44,20 +52,241 @@ type DisaggregatedSetReconciler struct {
...
+		return ctrl.Result{}, err
+	}
+
+	var result ctrl.Result
+	roleNames := disaggregatedsetutils.GetRoleNames(disaggregatedSet)
+	totalOldReplicas := 0
+	for _, roleName := range roleNames {
+		totalOldReplicas += oldRevisions.GetTotalReplicasPerRole(roleName)
+	}
+	if len(oldRevisions) > 0 && totalOldReplicas > 0 {
```
ahg-g, line 102: document the decsion tree here and the difference between ReconcileRollingUpdateNew and reconcileSimple pls

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -44,20 +52,241 @@ type DisaggregatedSetReconciler struct {
...
+func (r *DisaggregatedSetReconciler) reconcileSimple(ctx context.Context, disaggregatedSet *disaggregatedsetv1.DisaggregatedSet, revision string) (ctrl.Result, error) {
+	roleConfigs := disaggregatedsetutils.GetRoleConfigs(disaggregatedSet)
+
+	for role, config := range roleConfigs {
+		if err := r.reconcileRoleSimple(ctx, disaggregatedSet, role, config, revision); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to reconcile %s role: %w", role, err)
+		}
+	}
+
+	if err := r.cleanupOldLWS(ctx, disaggregatedSet, revision); err != nil {
```
ahg-g: We do something similar earlier in `Reconcile` by calling `cleanupDrainedLWS`, why invoking another cleanup again here? what is the difference?

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -0,0 +1,445 @@
...
+	EventReasonLWSDeleted             = "LWSDeleted"
+)
+
+type RollingUpdateExecutor struct {
+	Client     client.Client
+	Record     events.EventRecorder
+	LWSManager *LeaderWorkerSetManager
+}
+
+func (executor *RollingUpdateExecutor) ReconcileRollingUpdateNew(
```
ahg-g, line 56: pls document at the high level the steps that this function takes to reconcile a rolling update

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -0,0 +1,445 @@
...
+	roleNames []string,
+	roleConfigs map[string]*disaggregatedsetv1.DisaggregatedRoleSpec,
+	oldRevisions disaggregatedsetutils.RevisionRolesList,
+) (ctrl.Result, error) {
+	log := logf.FromContext(ctx)
+	log.Info("Initiating new rolling update", "revision", revision)
+	executor.Record.Eventf(disaggregatedSet, nil, corev1.EventTypeNormal, EventReasonRollingUpdateStarted,
+		"Update", "Started rolling update to revision %s", revision)
+
+	for _, oldGrouped := range oldRevisions {
```
ahg-g, line 101: pls document what this step does.

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -0,0 +1,445 @@
...
+			if roleLWS.Spec.Replicas != nil {
+				replicas = int(*roleLWS.Spec.Replicas)
+			}
+			if _, err := executor.LWSManager.SetInitialReplicas(ctx, disaggregatedSet.Namespace, lwsName, replicas); err != nil {
+				log.Error(err, "Failed to set initial-replicas annotation", "lws", lwsName)
+			}
+		}
+	}
+
+	for _, roleName := range roleNames {
```
ahg-g, line 116: ditto, pls document what the next block aims to do

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -0,0 +1,445 @@
...
+			log.Info("Created new LWS", "role", roleName, "revision", revision)
+			executor.Record.Eventf(disaggregatedSet, nil, corev1.EventTypeNormal, EventReasonLWSCreated,
+				"Create", "Created %s LWS %s", roleName, lwsName)
+		}
+	}
+
+	return ctrl.Result{RequeueAfter: time.Second}, nil
+}
+
+func (executor *RollingUpdateExecutor) ReconcileRollingUpdate(
```
ahg-g, line 130: pls document at the high level the steps this function executes

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -0,0 +1,445 @@
...
+	newRevision disaggregatedsetutils.RevisionRoles,
+) (ctrl.Result, error) {
+	log := logf.FromContext(ctx)
+	specRoleNames := disaggregatedsetutils.GetRoleNames(disaggregatedSet)
+	specRoleSet, oldRoleSet := buildRoleSets(specRoleNames, oldRevisions)
+
+	allRoleNames := append(slices.Clone(specRoleNames), removedRoles(oldRoleSet, specRoleSet)...)
+
+	if !isRevisionStable(newRevision, specRoleNames) {
+		log.V(1).Info("Waiting for new revision to stabilize")
```
ahg-g, line 146: pls document what does it mean for a "new revision to stabilize"

ahg-g (commented):
Thanks, my comments are mostly on documenting the logic.

On testing, we seem to rely heavily on e2e tests, integration tests should be simpler and easier to setup (no need to all the kubectl logic), or am I missing something?

hasB4K:
> On testing, we seem to rely heavily on e2e tests, integration tests should be simpler and easier to setup (no need to all the kubectl logic), or am I missing something?

I'm doing both. I could remove the e2e tests before merging this PR to iterate on it. But I'm testing with and without it. The main reason was to also tests weird cases, and the integrations logic of the executor/planner etc. - but there is a lot of overlap between integrations tests and the other tests.

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -0,0 +1,445 @@
...
+
+	for _, roleName := range roleNames {
+		created, err := executor.ensureNewLWSExists(ctx, disaggregatedSet, revision, roleName, roleConfigs[roleName], 0)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		if created {
+			lwsName := disaggregatedsetutils.GenerateName(disaggregatedSet.Name, roleName, revision)
+			log.Info("Created new LWS", "role", roleName, "revision", revision)
+			executor.Record.Eventf(disaggregatedSet, nil, corev1.EventTypeNormal, EventReasonLWSCreated,
```
Edwinhr716: Is there a reason why we log / record this event here instead of inside lwsManager.Create function?

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
hasB4K: cleanupOldLWS was indeed redundant. Nice catch!

pkg/controllers/disaggregatedset/executor.go
hasB4K: No, this is now solved

hasB4K:
Regarding the e2e tests, I think we should refactor them in a follow PR if that's okay?

PR #840, opened 2026-05-01
Updated the main readme to discuss DisaggregatedSet

ahg-g:
Updated the main docs page to introduce DisaggregatedSet which was contributed by the llm-d project


Edwinhr716:
Should we wait for the concept of `slice` to be implemented before adding this?

ahg-g:
> Should we wait for the concept of `slice` to be implemented before adding this?

It is fine, we can add it now and follow up with the implementation.

PR #844, opened 2026-05-03
KEP-666: Gang Scheduling in LWS

### LINKED ISSUE #666: [RFC] Support kubernetes gang scheduling as a pod group provider

gang scheduling in LWS is an interesting feature.

There is an upstream effort in k8s around gang scheduling: https://github.com/kubernetes/enhancements/pull/5558

It would be great if we could provide a support via alpha APIs once the PR merges.

This way we could support gang scheduling intree.

**Why is this needed**:
See above
**Completion requirements**:

- [x] Design doc
- [ ] API change
- [ ] Docs update

yankay:
#### What type of PR is this?

/kind feature
/kind documentation

Initial draft of **KEP-666: Gang Scheduling in LWS**, integrating the upstream Workload and PodGroup APIs (alpha, [kubernetes/enhancements#5558](https://github.com/kubernetes/enhancements/pull/5558), [kubernetes/enhancements#5832](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5832-decouple-podgroup-api)) as a gang-scheduling provider for `LeaderWorkerSet`, alongside the existing third-party path in [KEP-407](https://github.com/kubernetes-sigs/lws/tree/main/keps/407-gang-scheduling).

##### Key design points

- **One PodGroup per replica.** An LWS with `replicas: N, size: M` produces N PodGroups, each with `MinCount: M`. A single shared PodGroup with `Replicas=N, MinCount=M` cannot express "every replica is complete" under the alpha API, so it would not actually prevent partial-replica scheduling.
- **Lifecycle invariants** (full details in the KEP): PodGroup `<lws-name>-<i>` is created **before** any pod with `group-index=i` (covering first creation, scale-up, and `maxSurge` bursts); LWS owns the Workload and steady-state PodGroups; the Workload `podGroupTemplate` is created once and never mutated; PodGroups are keyed by `group-index` and reused across revisions, with `minCount` updated in place when `Size` changes.
- **LWS-managed lifecycle only for alpha.** A user-managed mode (`gang.podGroupNamePrefix`) was considered and dropped — see Alternatives. It can be revisited once a concrete external consumer needs it.

See also "Limitations of the alpha PodGroup API" in the KEP for why a single `MinCount` scalar cannot express the per-role availability needed by [KEP-766 DisaggregatedSet](https://github.com/kubernetes-sigs/lws/tree/main/keps/766-DisaggregatedSet).

##### Related work

Parallel efforts on the same upstream APIs ([KEP-4671](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/4671-gang-scheduling) + [KEP-5832](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5832-decouple-podgroup-api)) in sibling projects, listed for reviewer context:

- [kubernetes-sigs/jobset#1068](https://github.com/kubernetes-sigs/jobset/pull/1068) — JobSet's `GangConfig` KEP (currently on the older alpha1 API; under design discussion).
- [kubernetes/enhancements#5548](https://github.com/kubernetes/enhancements/pull/5548) — Job-side Workload integration KEP (deferred until upstream Workload API ships in v1.36+).
- [kubernetes-sigs/kueue](https://github.com/kubernetes-sigs/kueue) — no direct KEP-4671 integration PR yet; the Escape Hatch (pre-set `pod.spec.schedulingGroup`) leaves room for Kueue or any future external owner to drive `Workload` lifecycle with zero new LWS-side API surface.

LWS targets the newer `scheduling.k8s.io/v1alpha2` decoupled API; reviewer questions raised on JobSet #1068 (owner refs, Workload lifecycle, defaulting/validation, feature-gate posture) are addressed in the corresponding KEP-666 sections.

Fixes #666

- Design-only PR (KEP markdown); implementation will follow once the design is agreed.
- Co-authored with @Edwinhr716.

```release-note
NONE
```


Edwinhr716:
I was working on a prototype https://github.com/Edwinhr716/lws/tree/was-poc, happy to collaborate here

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,250 @@
...
+```
+
+Each replica gets one PodGroup; `pg-0` is shown below, `pg-1` … `pg-3` follow the same shape with their `group-index` in the name:
+
+```yaml
+apiVersion: scheduling.k8s.io/v1alpha2
+kind: PodGroup
+metadata:
+  name: leaderworkerset-sample-pg-0
+  ownerReferences:
```
Edwinhr716, line 230: Thoughts on having the leader pod own the PodGroup instead? That way it gets cleaned up when LWS is autoscaled. It also simplifies the logic of managing the life cycle during maxSurge and rolling updates

yankay: Thanks @Edwinhr716. My one concern is forward compatibility: [KEP-5832 §Risks and Mitigations](https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/5832-decouple-podgroup-api/README.md#risks-and-mitigations) plans a validating admission controller for `Workload → PodGroup → Pod` creation order, with `UnschedulableAndUnresolvable` kept only as a "last line of defense". If that lands and is on by default, pod-owned would break at admission instead of degrading.

That said, the cleanup simplification is real and you're closer to the implementation — happy to go with pod-owned, just want to note this as a known follow-up.

yankay:
Pushed an update aligning the KEP with the original PoC Google Doc:

- **Naming.** `Workload <lws-name>`, `PodGroup <lws-name>-<group-index>` (= leader pod name), inner template `<lws-name>-pg-template` — matches the PoC and KEP-407, and follows LWS's existing per-replica naming.
- **Escape hatch.** Pre-set `pod.spec.schedulingGroup` opts out of LWS-managed lifecycle (Job controller pattern); replaces the rejected explicit `podGroupNamePrefix` knob.
- **SubGroup / `minCount < size`.** Documented as future work via hierarchical PodGroups (KEP-6012), per the PoC.
- **Admission.** Trimmed to the four functionally required rules (annotation immutability, Size immutability in LWS-managed mode, `gang + LeaderReady`, `gang + exclusive-topology`) plus a UX rejection when the v1alpha2 API resources are not registered.
- **Editorial.** Removed the redundant `Interaction with StartupPolicy` section, merged the prerequisites risks, and tightened a few wordy passages.

keps/666-gang-scheduling-in-lws/README.md
Edwinhr716: Thanks for pointing that out, I wasn't aware that there was going to be a validating admission order.

Mmm that makes it trickier for the LWS controller to manage the lifecycle of PodGroups, and is something we need to think about in the design.

Edwinhr716:
Something we need to discuss further here is whether or not it makes sense to integrate with the PodGroup and Workload API now, or does it make more sense to wait until https://github.com/kubernetes/enhancements/pull/6017 to address the limitations that I flagged here https://docs.google.com/document/d/1VqfNB1u8cmrRhMe0DKycX-bfaLgDm5cgHGWdWu-94cM/edit?tab=t.0#bookmark=kix.e0bqf1kap91e. 

If the former, we also need to think about the migration from simple PodGroups to using CompositePodGroup APIs

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,320 @@
...
+- Gang scheduling for [DisaggregatedSet][kep766] (needs hierarchical gangs — see [Limitations](#limitations-of-the-alpha-workload-and-podgroup-apis)).
+- `gang-scheduling="true"` combined with `leaderworkerset.sigs.k8s.io/exclusive-topology` (rejected at admission).
+- Mutating `LeaderWorkerTemplate.Size` while gang is enabled in LWS-managed mode (`gang.minCount` is immutable in alpha2; rejected at admission).
+- `gang-scheduling="true" && StartupPolicy: LeaderReady` (alpha cannot express leader-first gangs; rejected at admission — see [Limitations](#limitations-of-the-alpha-workload-and-podgroup-apis)).
+
+[kep766]: https://github.com/kubernetes-sigs/lws/tree/main/keps/766-DisaggregatedSet
+
+## Proposal
+
+When the LWS object carries `leaderworkerset.sigs.k8s.io/gang-scheduling: "true"`, LWS creates and owns one `scheduling.k8s.io/v1alpha2` Workload (holding a gang PodGroup template) plus one standalone `PodGroup` per replica; each PodGroup's `MinCount` defaults to `LeaderWorkerTemplate.Size`, so all pods of a replica co-schedule by default. The pod webhook sets each pod's `spec.schedulingGroup.podGroupName` from its `leaderworkerset.sigs.k8s.io/group-index` label.
```
kannon92: Will there be any API discovery on k8s clusters for this feature?

ie check if this API is available in addition to the annotation.

kannon92: Should we go with an API for alpha with a feature gate instead?

Annotations are a bit hacky and difficult to deprecate. Plus I'm not sure how you plan to support ResourceClaims or TopologyAwareScheduling without going with an API.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,320 @@
...
+
+[kep407]: https://github.com/kubernetes-sigs/lws/tree/main/keps/407-gang-scheduling
+[kep4671]: https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/4671-gang-scheduling
+
+### Goals
+
+- Validate [KEP-4671][kep4671] (Workload / PodGroup APIs) for multi-host inference use cases.
+- Support autoscaling at the replica level.
+
+### Non-Goals
```
kannon92, line 61: You should maybe call out all the other features for WAS as Non-Goals .

TAS, Workload disruption, PodGroupResourceClaims.

keps/666-gang-scheduling-in-lws/README.md
Edwinhr716: I'm fine with creating an API field. I did an annotation because it was an easy way to enable the prototype, but if we want to have an actual integration I agree we should add an API field

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,320 @@
...
+- **`SubGroupPolicy`**: the natural design is one PodGroup per SubGroup (each with its own `subgroup-exclusive-topology` constraint) plus a parent gang over all SubGroup PodGroups in the replica.
+- **`minCount < size`** (leader-first gang): a leader PodGroup with `MinCount = 1` plus a worker PodGroup with user-chosen `MinCount`, joined under one parent gang.
+- **`StartupPolicy: LeaderReady`**: same root cause — no `MinResources`, so a single-`MinCount` PodGroup can't admit the leader before workers exist.
+
+All four resolve via hierarchical PodGroups upstream (tracked by [KEP-6012][kep6012]; also [KEP-5832][podgroup-kep] §Future Plans). Until that lands, alpha creates a single replica-level PodGroup with `MinCount = LeaderWorkerTemplate.Size`; SubGroup keeps driving placement (topology + TPU env), but per-subgroup gang minimums are not enforced. KEP-766 will track the LWS-side design for DisaggregatedSet once hierarchy lands.
+
+[kep6012]: https://github.com/kubernetes/enhancements/issues/6012
+
+### Future Work: Hierarchical Gang via CompositePodGroup
```
Edwinhr716, line 266: Something else to call out here. A composite gang that places the leader and the workers in separate PodGroups

```text
CompositePodGroup serving-root
├─ PodGroup       leader         minCoint = 1 (parentRef=serving-root)
└─ PodGroup       workers        minCount = lws.size - 1 (parentRef=serving-root)
```

Edwinhr716: This covers the use case where the leader requests different resources from the worker, and the use case where we want to give priority to the leader when it comes to preemption

keps/666-gang-scheduling-in-lws/README.md
yankay: Yes — at admission the webhook resolves v1alpha2 Workload + PodGroup via a cached RESTMapper, missing → reject with an error naming the missing GVK. See the new API Discovery and Prerequisites section. Caveat: the upstream GenericWorkload gate itself can't be discovered, so it stays an install-time prereq.

yankay: > Should we go with an API for alpha with a feature gate instead?
> 
> Annotations are a bit hacky and difficult to deprecate. Plus I'm not sure how you plan to support ResourceClaims or TopologyAwareScheduling without going with an API.

HI @kannon92 Quick check — did "feature gate" mean an LWS-side gate, or upstream GenericWorkload as the de-facto guard? Latest push takes the latter: typed alpha `spec.gangScheduling`, no LWS gate (matches SubGroupPolicy / RolloutStrategy.MaxSurge). The LWS-side feature-gate scaffold question is tracked separately in [#850](https://github.com/kubernetes-sigs/lws/issues/850) — happy to flip back if you meant the LWS-gate reading.

yankay: > I'm fine with creating an API field. I did an annotation because it was an easy way to enable the prototype, but if we want to have an actual integration I agree we should add an API field

Done — typed alpha `spec.gangScheduling` field replaces the annotation. Empty struct in alpha (presence = opt-in); future TAS / DRA / RC knobs added additively. No LWS-side feature gate — webhook API discovery against upstream `GenericWorkload` is the guard. Scaffold question tracked separately in #850.

keps/666-gang-scheduling-in-lws/README.md
yankay: Added under Non-Goals: TAS (KEP-5732), workload-aware preemption / disruption (KEP-5710), PodGroup-shared ResourceClaims (KEP-5729). Escape Hatch is the alpha workaround.

keps/666-gang-scheduling-in-lws/README.md
yankay: Done in 15ed9ce — Limitations §Per-role gang policy + Future Work §Single-LWS per-role split (per-replica CompositePodGroup tree).

Open question: any concrete LWS workload where splitting leader and workers into separate PodGroups actually helps? In vLLM head and workers all sit inside the same TP group — same GPU shape, all-or-nothing — so two leaf PodGroups give the same scheduling outcome as a single `MinCount = size` PodGroup. Heterogeneous resources / leader preemption priority make sense in the abstract but I can't map them to a real LWS workload yet.

Edwinhr716: > any concrete LWS workload where splitting leader and workers into separate PodGroups actually helps?

Yes, take a look at axlearn for example https://github.com/apple/axlearn/blob/main/axlearn/cloud/gcp/pathways_utils.py#L1611. Their leader only requests CPU resources, while the workers request TPUs.

 To guarantee that the workers all fall into the same TPU slice, while also being able to run the leader in a separate CPU nodepool, they use `LeaderOnly` subgroup policy + subgroup-exclusive-topology. That use case would be covered by having separate PodGroups for leader and workers

yankay: > > any concrete LWS workload where splitting leader and workers into separate PodGroups actually helps?
> 
> Yes, take a look at axlearn for example https://github.com/apple/axlearn/blob/main/axlearn/cloud/gcp/pathways_utils.py#L1611. Their leader only requests CPU resources, while the workers request TPUs.
> 
> To guarantee that the workers all fall into the same TPU slice, while also being able to run the leader in a separate CPU nodepool, they use `LeaderOnly` subgroup policy + subgroup-exclusive-topology. That use case would be covered by having separate PodGroups for leader and workers

Thanks — pulled into the KEP as the heterogeneous-role motivating shape (CPU leader + single-accelerator-slice workers). Concrete reference: [axlearn pathways_utils.py#L1611](https://github.com/apple/axlearn/blob/main/axlearn/cloud/gcp/pathways_utils.py#L1611).

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,370 @@
...
+    // +optional
+    GangScheduling *GangSchedulingPolicy `json:"gangScheduling,omitempty"`
+}
+
+// Empty in alpha — presence is the only knob. TAS / DRA / ResourceClaims
+// and hierarchical (KEP-6012) knobs are added additively as they stabilize.
+type GangSchedulingPolicy struct{}
+```
+
+No LWS-side feature gate: the upstream `GenericWorkload` gate already controls whether `kube-apiserver` preserves `pod.spec.schedulingGroup`, and the webhook's [API discovery](#api-discovery-and-prerequisites) propagates that into LWS admission. Matches how LWS handles other typed alpha fields (`SubGroupPolicy`, `RolloutStrategy.MaxSurge`); a project-wide `pkg/features` scaffold, if ever needed, is tracked in [#850][lws-feature-gate-issue].
```
ardaguclu: So once GenericWorkload FG is promoted to stable in Kubernetes, it will be GA'ed here too?.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,370 @@
...
+```go
+// api/leaderworkerset/v1/leaderworkerset_types.go
+type LeaderWorkerSetSpec struct {
+    // ... existing fields ...
+
+    // GangScheduling opts the LWS into LWS-managed gang scheduling via the
+    // upstream Workload / PodGroup APIs (one PodGroup per replica,
+    // MinCount = LeaderWorkerTemplate.Size). Alpha; subject to change.
+    // +optional
+    GangScheduling *GangSchedulingPolicy `json:"gangScheduling,omitempty"`
```
ardaguclu, line 114: This field can not be set if feature gate is not enabled on the cluster?.

ardaguclu: What would happen if the LWS (including this feature) will be deployed onto old Kubernetes cluster?. As far as I understand, this is backwards compatible, right?.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,370 @@
...
+
+**One shared PodGroup per LWS with `Replicas=N, MinCount=M`** (Edwin's original draft).
+Rejected: `MinCount` only requires M co-scheduled pods, with no notion of which replica they belong to — the scheduler may legally pick M pods from different replicas, none complete, and the model still cannot start. Per-replica PodGroups make each replica an independent all-or-nothing unit.
+
+**Rely on [KEP-407][kep407] only**.
+KEP-407 targets third-party schedulers (Volcano / coscheduling / YuniKorn) via their own PodGroup CRDs; this KEP targets the upstream-native `scheduling.k8s.io/v1alpha2` Workload and PodGroup APIs. The two evolve independently — different prerequisites, different API surfaces, no shared data path — and a single LWS object opts into at most one.
```
ardaguclu: As described above, we have added Volcano based gang-scheduling already. Users are expected to use one of them (Kubernetes gang-scheduling, Volcano)?

yankay: Good point — pushed an update so they're not mutually exclusive. New [Unified Provider Model](https://github.com/kubernetes-sigs/lws/pull/844/changes#diff-4cef677e4df989d04655908d61d914aa2bc6fadf0bb8809f49920fe8ebe834a6R133) makes `spec.gangScheduling` the single opt-in across all gang backends (this KEP + KEP-407's Volcano; KAI / coscheduling possible later), following Grove's `Backend` framework.

keps/666-gang-scheduling-in-lws/README.md
yankay: > What would happen if the LWS (including this feature) will be deployed onto old Kubernetes cluster?. As far as I understand, this is backwards compatible, right?.

Yes — added an explicit [Backwards Compatibility](https://github.com/kubernetes-sigs/lws/pull/844/changes#diff-4cef677e4df989d04655908d61d914aa2bc6fadf0bb8809f49920fe8ebe834a6R94) covering the three cases:

- Field unset → zero behavior change.
- Field set, v1alpha2 APIs missing → admission rejects with the missing GVK named; no half-created objects.
- APIs registered but GenericWorkload gate off → install-time prerequisite, not a runtime failure.

yankay: > This field can not be set if feature gate is not enabled on the cluster?.

Yes — spec.gangScheduling is an LWS field, but the LWS validating webhook rejects it at admission if the upstream v1alpha2 Workload/PodGroup GVKs aren't registered (see [API Discovery](https://github.com/kubernetes-sigs/lws/pull/844/changes#diff-4cef677e4df989d04655908d61d914aa2bc6fadf0bb8809f49920fe8ebe834a6R145)).

keps/666-gang-scheduling-in-lws/README.md
yankay: Yes — LWS GangScheduling tracks the upstream GenericWorkload lifecycle: alpha → beta (default-on, aligned with upstream beta) → removed at GA. See updated [§Graduation Criteria](https://github.com/kubernetes-sigs/lws/pull/844/changes#diff-4cef677e4df989d04655908d61d914aa2bc6fadf0bb8809f49920fe8ebe834a6R350).

yankay:
> Something we need to discuss further here is whether or not it makes sense to integrate with the PodGroup and Workload API now, or does it make more sense to wait until [kubernetes/enhancements#6017](https://github.com/kubernetes/enhancements/pull/6017) to address the limitations that I flagged here https://docs.google.com/document/d/1VqfNB1u8cmrRhMe0DKycX-bfaLgDm5cgHGWdWu-94cM/edit?tab=t.0#bookmark=kix.e0bqf1kap91e.
> 
> If the former, we also need to think about the migration from simple PodGroups to using CompositePodGroup APIs

Thanks @Edwinhr716 — I think the latest revision already covers both:

- **Integrate now vs wait for [kubernetes/enhancements#6017](https://github.com/kubernetes/enhancements/pull/6017):** Graduation Criteria tracks upstream `GenericWorkload`'s lifecycle — no blocking on #6017, but aligned when it lands. PoC-doc limitations listed under Future Work.
- **Migration to CompositePodGroup:** Unified Provider Model + Backwards Compatibility + API Discovery — `spec.gangScheduling` is a single opt-in across backends, so CompositePodGroup slots in later without breaking existing users.

Does this match what you had in mind?

yankay:
Hi @Edwinhr716 and @ahg-g — I believe the open review threads are addressed in the latest revision. Would really appreciate another pass whenever it fits your schedule. Thanks so much for the feedback so far!

keps/666-gang-scheduling-in-lws/README.md
yankay: > Should we go with an API for alpha with a feature gate instead?
> 
> Annotations are a bit hacky and difficult to deprecate. Plus I'm not sure how you plan to support ResourceClaims or TopologyAwareScheduling without going with an API.

Following up on this — also adopted your [feedback on #852](https://github.com/kubernetes-sigs/lws/pull/852#issuecomment-4467561258): the `GangScheduling` gate is delivered via the LWS Configuration API rather than a CLI flag, since CLI-only gates make rolling upgrades painful. KEP updated in the latest push.

keps/666-gang-scheduling-in-lws/README.md
Edwinhr716: > happy to go with pod-owned, just want to note this as a known follow-up.

Is this called out anywhere? Don't see it in graduation criteria

Edwinhr716:
Mostly looks good to me, just want to make sure we call out the leaderPod ownership issue for beta

yankay:
> docs: call out PodGroup ownership

Thanks @Edwinhr716 

Thanks, good point. I called out leader-pod ownership as an alpha shortcut and added a beta criterion to validate it against deletion protection or switch to an LWS-owned/revision-keyed lifecycle.

mm4tt:
Please take a look at kep.k8s.io/6089, where we propose API building blocks for controllers integrating with WAS. This allows expressing scheduling intent (e.g., scheduling policy, topology, disruption mode) directly on the controller side. The Job controller is currently integrating with this (https://github.com/kubernetes/enhancements/pull/6153), and we hope JobSet will follow. 

LWS adopting this approach would provide a strong signal for kep.k8s.io/6089 beta graduation in 1.38 and ensure ecosystem consistency by standardizing how scheduling constraints are expressed. Since the underlying mechanisms are already in-place, I'd recommend to align the API design correctly from the start.

yankay:
> Please take a look at kep.k8s.io/6089, where we propose API building blocks for controllers integrating with WAS. This allows expressing scheduling intent (e.g., scheduling policy, topology, disruption mode) directly on the controller side. The Job controller is currently integrating with this ([kubernetes/enhancements#6153](https://github.com/kubernetes/enhancements/pull/6153)), and we hope JobSet will follow.
> 
> LWS adopting this approach would provide a strong signal for kep.k8s.io/6089 beta graduation in 1.38 and ensure ecosystem consistency by standardizing how scheduling constraints are expressed. Since the underlying mechanisms are already in-place, I'd recommend to align the API design correctly from the start.

That makes sense, thanks. I’ll study this and use it as a reference.

yankay:
To avoid duplication and fragmented discussion, let’s continue in [[PR 979](https://github.com/kubernetes-sigs/lws/pull/979)](https://github.com/kubernetes-sigs/lws/pull/979). Thanks, [Paco](https://github.com/pacoxu). Given the work in [[kubernetes/enhancements PR 6153](https://github.com/kubernetes/enhancements/pull/6153)](https://github.com/kubernetes/enhancements/pull/6153), the original KEP needs substantial updates, and a new PR will better consolidate the discussion and conclusions.

PR #852, opened 2026-05-14
feat: add pkg/features scaffold for LWS feature gates

### LINKED ISSUE #850: Add a project-wide pkg/features scaffold for LWS feature gates
**What would you like to be added:**

Add a `pkg/features/` scaffold so LWS can gate alpha API fields with `+featureGate=` markers, matching [Kueue](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/features/kube_features.go) and [JobSet](https://github.com/kubernetes-sigs/jobset/blob/main/pkg/features/features.go).

This PR is plumbing only — no feature constants. First consumer will be [KEP-666 Gang Scheduling](https://github.com/kubernetes-sigs/lws/pull/844) (`GangScheduling`).

Scope:
- `pkg/features/features.go` registering `defaultFeatureGates` on `utilfeature.DefaultMutableFeatureGate` from `init()`, with a unit test.
- `cmd/main.go` adds `--feature-gates=Foo=true,Bar=false` via `AddFlag(pflag.CommandLine)`.
- Helm chart / kustomize manifests expose a `featureGates` value.

**Why is this needed:**

Today alpha behaviors are gated by typed optional fields or annotations. Neither matches the kube-sigs convention for alpha fields on stable types. Landing the scaffold standalone avoids each feature KEP re-introducing it.

/kind feature
/area api

aspr0d:
## What type of PR is this?
/kind feature

## What this PR does / why we need it:

Adds a project-wide `pkg/features` scaffold mirroring [Kueue](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/features/kube_features.go) and [JobSet](https://github.com/kubernetes-sigs/jobset/blob/main/pkg/features/features.go), so alpha/experimental behaviors can be gated with `+featureGate=` markers backed by a real feature gate registry instead of typed optional fields or annotations.

This change is **plumbing only**: the `defaultFeatureGates` map is intentionally empty so the first consumer (KEP-666 Gang Scheduling) can introduce its constant in a follow-up PR with no infrastructure work, per the issue scope.

### Included

- `pkg/features/features.go` registers `defaultFeatureGates` on `utilfeature.DefaultMutableFeatureGate` from `init()` and exposes `Enabled`, `SetEnable`, and `SetFeatureGateDuringTest` helpers.
- `pkg/features/features_test.go` covers the `SetEnable` and `SetFeatureGateDuringTest` paths against a temporary registered feature.
- `cmd/main.go` adds `--feature-gates=Foo=true,Bar=false` (parsed via `utilfeature.DefaultMutableFeatureGate.Set`) and lists the known features in the flag usage string.
- `charts/lws` exposes a `featureGates` Helm value and forwards it as `--feature-gates` when set.

### Verified locally

- `go build ./...` and `go vet ./...` clean.
- `go test ./pkg/features/...` passes (`TestSetEnable`, `TestSetFeatureGateDuringTest`).
- `./manager --help` shows `--feature-gates` and renders the known feature list.
- `./manager --feature-gates=WatchList=false` exits past the feature-gate phase (subsequent in-cluster config error is unrelated).
- `helm template lws charts/lws -f values.yaml` with `featureGates: "Foo=true,Bar=false"` correctly renders `- --feature-gates=Foo=true,Bar=false`; with the default empty value the flag is omitted.

## Which issue(s) this PR fixes:

Fixes #850

## Special notes for your reviewer:

- I intentionally did **not** touch `config/manager/manager.yaml` (kustomize): the kustomize manager wires its configuration through a ConfigMap rather than additional CLI args, and adding the flag there would conflict with the `--config` flow. Happy to expose it there too if reviewers prefer; let me know which mechanism you want.
- No feature constants are added in this PR by design, so the example line in `defaultFeatureGates` is a comment only.

```release-note
Introduce a project-wide feature-gate scaffold under `pkg/features`. A new `--feature-gates=Foo=true,Bar=false` flag and a corresponding `featureGates` Helm value are now available; no feature constants are registered yet.
```


Mostafahassen1:
Hi @aspr0d  👋

If you'd like to become a new contributor, please complete the contributor agreement first before making any contributions.

Regular contributors should [join the org](https://git.k8s.io/community/community-membership.md#member)

Thank you for contributing to the project 🚀

pkg/features/features.go
```diff
@@ -0,0 +1,58 @@
...
+// when adding or removing one entry.
+var defaultFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
+	// Example: {Default: false, PreRelease: featuregate.Alpha},
+}
+
+// SetFeatureGateDuringTest is a helper for setting a feature gate value
+// for the duration of a test.
+func SetFeatureGateDuringTest(tb testing.TB, f featuregate.Feature, value bool) {
+	featuregatetesting.SetFeatureGateDuringTest(tb, utilfeature.DefaultFeatureGate, f, value)
+}
```
Mostafahassen1: Please remove the "testing" and `featuregatetesting` imports from `features.go`, as test packages must not be included in production code.

pkg/features/features.go
```diff
@@ -0,0 +1,58 @@
...
+// Enabled is a helper for utilfeature.DefaultFeatureGate.Enabled().
+func Enabled(f featuregate.Feature) bool {
+	return utilfeature.DefaultFeatureGate.Enabled(f)
+}
+
+// SetEnable is a helper to flip a feature gate from code paths such as
+// integration tests. Prefer SetFeatureGateDuringTest for unit tests.
+func SetEnable(f featuregate.Feature, v bool) error {
+	return utilfeature.DefaultMutableFeatureGate.Set(fmt.Sprintf("%s=%v", f, v))
+}
```
Mostafahassen1, line 50: > Please remove the "testing" and featuregatetesting imports from features.go, as test packages must not be included in production code.

I think this also

pkg/features/features.go
```diff
@@ -0,0 +1,58 @@
...
+	"testing"
+
+	"k8s.io/apimachinery/pkg/util/runtime"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
+	"k8s.io/component-base/featuregate"
+	featuregatetesting "k8s.io/component-base/featuregate/testing"
+)
+
+func init() {
+	runtime.Must(utilfeature.DefaultMutableFeatureGate.Add(defaultFeatureGates))
```
Mostafahassen1, line 28: Really, I  don't know the `Add` here must be ' Add ' or `AddVersioned.`

I noticed we are using `Add()` for the feature gates. In newer Kubernetes core versions, the standard is shifting toward [`AddVersioned()`](https://pkg.go.dev/k8s.io/component-base/featuregate#MutableVersionedFeatureGate.AddVersioned) to better track feature lifecycles (Alpha/Beta/GA transitions). 

Since LWS has versioned releases, we might want to consider using `AddVersioned()` from day one

What's your opinion:  @yankay @Edwinhr716

aspr0d:
Fixed the commit message: replaced `Fixes #850` with `See issue 850 for context.` to avoid the auto-close keyword. Force-pushed

pkg/features/features.go
```diff
@@ -0,0 +1,58 @@
+/*
```
Mostafahassen1, line 1: I noticed we are missing a helper to log the active feature gates at startup. 

Could we add a `LogFeatureGates(log logr.Logger)` function in the `features` package and call it in `main.go` during the controller setup? This is highly valuable for production debugging so we can easily see which features are enabled/disabled when checking the controller logs.

https://github.com/kubernetes-sigs/kueue/blob/main/pkg/features/kube_features.go#L612

We must add it in this step or not @yankay @Edwinhr716

cmd/main.go
```diff
@@ -44,6 +46,7 @@ import (
 	"sigs.k8s.io/lws/pkg/config"
 	"sigs.k8s.io/lws/pkg/controllers"
 	disaggregatedsetcontroller "sigs.k8s.io/lws/pkg/controllers/disaggregatedset"
+	_ "sigs.k8s.io/lws/pkg/features"
```
Mostafahassen1, line 48: Could you clarify why we are using a blank import `_` here?

pkg/features/features_test.go
```diff
@@ -0,0 +1,74 @@
...
+		t.Fatalf("expected %s to be disabled by default", testFeature)
+	}
+
+	if err := SetEnable(testFeature, true); err != nil {
+		t.Fatalf("SetEnable returned error: %v", err)
+	}
+	t.Cleanup(func() {
+		_ = SetEnable(testFeature, false)
+	})
+
```
Mostafahassen1: I noticed we are using `SetEnable` combined with `t.Cleanup` here. Modifying global state in unit tests this way can be slightly fragile and lead to flaky tests if the cleanup order changes or if a panic occurs before registration.

To align with standard Kubernetes testing patterns, we should use [`featuregatetesting.SetFeatureGateDuringTest`](https://pkg.go.dev/k8s.io/component-base/featuregate/testing#SetFeatureGateDuringTest) for unit tests instead. It safely handles state mutation and guarantees proper reversion. 

`SetEnable` is generally better suited for integration tests. Could we rewrite these unit tests to use `SetFeatureGateDuringTest`?

Mostafahassen1:
Hi @Edwinhr716, I just finished my shadow review and left a few comments

aspr0d:
Thanks for the shadow review @Mostafahassen1.

**Test imports in production code** - moved `SetFeatureGateDuringTest` (along with the `testing` and `featuregatetesting` imports) into a new `pkg/features/testing.go` file. `features.go` now has no test-package imports.

**Blank import in cmd/main.go** - the `_ "sigs.k8s.io/lws/pkg/features"` blank import is there to trigger the `init()` in `features.go`, which registers the feature gates with `utilfeature.DefaultMutableFeatureGate`. Without it the registration would not run before the flag is parsed. Updated the commit message to make this explicit.

**TestSetEnable cleanup** - replaced the manual `t.Cleanup` with a `SetFeatureGateDuringTest` call at the top of the test so state is always restored via the tested helper.

**Add vs AddVersioned / LogFeatureGates** - happy to follow whatever the maintainers prefer. These can be addressed in a follow-up or as part of this PR if requested.

Force-pushed.

Mostafahassen1:
>Test imports in production code - moved SetFeatureGateDuringTest (along with the testing and featuregatetesting imports) into a new pkg/features/testing.go file. features.go now has no test-package imports.

>Blank import in cmd/main.go - the _ "sigs.k8s.io/lws/pkg/features" blank import is there to trigger the init() in features.go, which registers the feature gates with utilfeature.DefaultMutableFeatureGate. Without it the registration would not run before the flag is parsed. Updated the commit message to make this explicit.

>TestSetEnable cleanup - replaced the manual t.Cleanup with a SetFeatureGateDuringTest call at the top of the test so state is always restored via the tested helper.

>Add vs AddVersioned / LogFeatureGates - happy to follow whatever the maintainers prefer. These can be addressed in a follow-up or as part of this PR if requested.

Thank you  @aspr0d for your time. To make the review process easier, could you please add your answers in the same place as the error? Also, if you fixed it, please click `Resolve comment`

image

cmd/main.go
aspr0d: The blank import `_ "sigs.k8s.io/lws/pkg/features"` is needed to trigger the `init()` function in `pkg/features/features.go`, which registers the LWS feature gates with `utilfeature.DefaultMutableFeatureGate`. Without it the registration would not run before the `--feature-gates` flag is parsed in `cmd/main.go`. The package itself is not directly referenced in `main.go`, so Go would otherwise drop the import as unused.

cmd/main.go
```diff
@@ -84,10 +87,15 @@ func main() {
 		leaderElectResourceLock  string
 		leaderElectionID         string
 		configFile               string
+		featureGates             string
 		printVersion             bool
 	)
 
 	flag.BoolVar(&printVersion, "version", false, "Print version information and exit")
+	flag.StringVar(&featureGates, "feature-gates", "",
```
kannon92: In Kueue/JobSet we want to deprecate the use of feature gates in the CLI for the config api.

I would recommend going with a configuration api option for feature gates.

kannon92:
/hold

https://github.com/kubernetes-sigs/lws/pull/852/changes#r3253175937

Happy to discuss more on this. In Kueue we originally went with CLI arguments but it was found that supporting upgrades via CLI changes is quite painful. I added featureGate support to the configuration api which is a bit easier to roll out via upgrades.

Plus this matches the Kubernetes process where we are encourage the use of component config options rather than growing the list of CLI commands.

In JobSet/Kueue we would prefer to not support CLI arguments but we introduced both at the time. You already have a configuration api for LWS so I recommend setting featureGates in that API rather than growing the CLI list.

cmd/main.go
aspr0d: Switched to a Configuration API field. Force-pushed.

- Removed the `--feature-gates` flag from `cmd/main.go`
- Added `FeatureGates map[string]bool` to `Configuration` in `api/config/v1alpha1`
- Apply `cfg.FeatureGates` via `utilfeature.DefaultMutableFeatureGate.SetFromMap` after `config.Load`
- Helm: `featureGates` is now a map and rendered into the manager ConfigMap (no longer a Deployment arg)
- `pkg/config` test covers parsing `featureGates` from YAML

Let me know if you would also like a `LogFeatureGates` helper logged at startup in this PR or a follow-up.

pkg/features/features.go
Mostafahassen1: If the `SetEnable` method is related to testing, I prefer it to be in the test file

Mostafahassen1:
@Edwinhr716 I finished the shadow review, and I don't have anything new to add.

pkg/features/features.go
yankay: I'd lean toward AddVersioned from day one.

pkg/features/features.go
yankay: Agree, this is useful for production debugging.

PR #851, opened 2026-05-09
fix: surface error message in FailedCreate events for leader and worker statefulsets

Mostafahassen1:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it
When creating a leader or worker StatefulSet fails, the warning event
emitted by the controller drops the actual error message, leaving the
operator with no actionable information beyond the StatefulSet name.

This PR includes the underlying error in the event message so users can
diagnose misconfigured specs (e.g. missing volumes, quota exceeded,
webhook rejections) directly from `kubectl describe leaderworkerset`
without having to dig into controller logs.

#### Which issue(s) this PR fixes
Fixes #845

#### Special notes for your reviewer
Minimal change — only the format strings in two Eventf calls are
updated to include `: %v` with the error. No logic, flow, or test
changes required.

Issue #845, opened 2026-05-05
[FEATURE REQUEST]: Better feedback when LWS is malformed


robertgshaw2-redhat:
**What would you like to be added**:
- LWS does not do a great job of surfacing why a deployment failed
- for instance, I recently had a `volumeMount` without a corresponding volume, which failed, but there is basically nothing in the status to suggest this was the root cause

```
robertshaw@robshaw-mac vllm % oc describe leaderworkersets wide-ep-lws-nvidia-gpu-vllm-decode
Name:         wide-ep-lws-nvidia-gpu-vllm-decode
Namespace:    default
Labels:       llm-d.ai/accelerator-variant=gpu
              llm-d.ai/accelerator-vendor=nvidia
              llm-d.ai/guide=wide-ep-lws
              llm-d.ai/model=DeepSeek-R1-0528
              llm-d.ai/role=decode
Annotations:  <none>
API Version:  leaderworkerset.x-k8s.io/v1
Kind:         LeaderWorkerSet
Metadata:
  Creation Timestamp:  2026-05-05T01:09:45Z
  Generation:          1
  Resource Version:    1777943385640783002
  UID:                 886d3e7f-3f9c-43b9-8964-fa719b6c45b2
Spec:
  Leader Worker Template:
    Restart Policy:  RecreateGroupOnPodRestart
    Size:            2
    Worker Template:
      Metadata:
        Annotations:
          networking.gke.io/default-interface:  eth0
          networking.gke.io/interfaces:         [
  {"interfaceName":"eth0","network":"default"},
  {"interfaceName":"eth2","network":"rdma-0"},
  {"interfaceName":"eth3","network":"rdma-1"},
  {"interfaceName":"eth4","network":"rdma-2"},
  {"interfaceName":"eth5","network":"rdma-3"},
  {"interfaceName":"eth6","network":"rdma-4"},
  {"interfaceName":"eth7","network":"rdma-5"},
  {"interfaceName":"eth8","network":"rdma-6"},
  {"interfaceName":"eth9","network":"rdma-7"}
]

        Labels:
          llm-d.ai/accelerator-variant:  gpu
          llm-d.ai/accelerator-vendor:   nvidia
          llm-d.ai/guide:                wide-ep-lws
          llm-d.ai/inference-serving:    true
          llm-d.ai/model:                DeepSeek-R1-0528
          llm-d.ai/role:                 decode
      Spec:
        Containers:
          Args:
[...]

PR #853, opened 2026-05-14
fix: include underlying error in worker statefulset FailedCreate event

aspr0d:
## What type of PR is this?
/kind bug

## What this PR does / why we need it:

The worker statefulset `FailedCreate` event currently logs only the leader pod name and drops the underlying error from `r.Create(...)`. Operators inspecting `kubectl describe leaderworkerset` therefore see that a worker statefulset failed to create without any actionable information about *why* (missing volumes, quota exceeded, webhook rejections, etc.).

This change appends \`: %v\` with the error to the existing format string so the warning event is self-contained, mirroring:
- The leader statefulset path that PR #851 is currently updating.
- The headless service path that already formats the error (\`leaderworkerset_controller.go\`).

### Before

\`\`\`
Warning  FailedCreate  10s  leaderworkerset  Failed to create worker statefulset for leader pod lws-sample-0
\`\`\`

### After

\`\`\`
Warning  FailedCreate  10s  leaderworkerset  Failed to create worker statefulset for leader pod lws-sample-0: admission webhook \"validation.lws.x-k8s.io\" denied the request: spec.template.spec.containers[0].volumeMounts[1].name: Not found: \"hf-cache\"
\`\`\`

Verified locally with \`go build ./...\`, \`go vet ./...\`, and the existing unit tests in \`pkg/controllers/\`.

## Which issue(s) this PR fixes:

Partially addresses #845. Together with #851 (leader statefulset path), this closes the two non-trivial \`FailedCreate\` call sites flagged by @Edwinhr716 in [this comment](https://github.com/kubernetes-sigs/lws/issues/845#issuecomment-3499050000).

## Special notes for your reviewer:

- The change is intentionally one-line and scoped to the worker statefulset path so it can land independently of #851 with no merge conflict on \`leaderworkerset_controller.go\`.
- No new tests added because there is no existing event-emission test harness in \`pod_controller_test.go\` and this is a format-string change; happy to add one with a fake recorder if reviewers want it.
- A follow-up that surfaces non-ready pod \`Reason\`/\`Message\` through \`updateConditions()\` (the third item in @Mostafahassen1's [comment](https://github.com/kubernetes-sigs/lws/pull/851#issuecomment-3508900000)) can land separately once this and #851 are in.

\`\`\`release-note
Worker statefulset \`FailedCreate\` events now include the underlying error so operators can diagnose creation failures directly from \`kubectl describe leaderworkerset\` without consulting controller logs.
\`\`\`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pkg/controllers/pod_controller.go
+++ b/pkg/controllers/pod_controller.go
@@ -191,7 +191,7 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 		}
 		if err = r.Create(ctx, workerStatefulSet); err != nil {
 			if client.IgnoreAlreadyExists(err) != nil {
-				r.Record.Eventf(&leaderWorkerSet, &pod, corev1.EventTypeWarning, FailedCreate, Create, fmt.Sprintf("Failed to create worker statefulset for leader pod %s", pod.Name))
+				r.Record.Eventf(&leaderWorkerSet, &pod, corev1.EventTypeWarning, FailedCreate, Create, fmt.Sprintf("Failed to create worker statefulset for leader pod %s: %v", pod.Name, err))
 			}
 			return ctrl.Result{}, client.IgnoreAlreadyExists(err)
 		}


Edwinhr716:
Already covered by #851

aspr0d:
Thanks for the pointer - #851 covers exactly this change. Closing in favour of that PR.

PR #870, opened 2026-06-01
Fix: (disaggregatedset) tighten ForSingleActiveRevision waiter

### LINKED ISSUE #826: Flaky e2e: disaggregatedset 'Rolling Update with Coordinated Drain' due to race in ForSingleActiveRevision waiter
**What happened**:

The disaggregatedset e2e test `Rolling Update with Coordinated Drain / should complete rolling update with both roles scaling together` (`disaggregatedset/test/e2e/e2e_test.go:213`) is flaky:

```
[FAILED] Revision <newRev> has prefill replicas but no decode (orphaned)
```

Example: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_lws/752/pull-lws-test-e2e-main-1-34/2049099489146310656

**Root cause**:

`kubectl.ForSingleActiveRevision` (`disaggregatedset/test/utils/kubectl/waiters.go:68`) only checks that exactly one revision has total `spec.replicas > 0`. It can return during an intermediate state where the new revision has only one role scaled up:

- old revision: prefill=0, decode=0 (drained)
- new revision: prefill=1, decode=0 (decode scale-up not yet applied)

The next assertion then sees prefill>0, decode=0 and reports the new revision as orphaned.

**Suggested fix**:

In `ForSingleActiveRevision`, also require that every role of the active revision has `replicas > 0` before returning.

yankay:

`ForSingleActiveRevision` summed `spec.replicas` across all roles of a revision and treated any revision with sum > 0 as active. Two escape windows let the waiter return before the rollout was actually settled:

1. **Mid-rollout** — during coordinated drain the new revision briefly has one role scaled up (e.g. `prefill=1, decode=0`); the waiter returned, and the next assertion reported the revision as orphaned.
2. **Apply-but-not-reconciled** — immediately after `kubectl apply`, only the old revision exists with all roles `> 0`. The revision-agnostic check accepted that as a single active revision before the controller had even created the new revision, so the rollout was effectively unverified.

This PR:

- Tracks per-revision how many role LWS still have `replicas=0` and only treats a revision as active when its total is `> 0` AND no role is at 0.
- Threads the old revision through the waiter and requires it to be drained before declaring the rollout settled. All e2e callers updated to pass `oldRevision`.
- Extracts the predicate as `evalSingleActiveRevision` and adds table-driven unit tests for the snapshot classes.

The waiter copy under `disaggregatedset/test/utils/kubectl/waiters.go` is left untouched — that folder is dead code, removed as part of #789. Supersedes #828.

Fixes #826

```release-note
NONE
```

### DIFF
--- a/test/e2e/disaggregatedset/e2e_test.go
+++ b/test/e2e/disaggregatedset/e2e_test.go
@@ -190,6 +190,9 @@ var _ = Describe("DisaggregatedSet E2E Tests", Ordered, func() {
 			By("waiting for initial deployment to stabilize")
 			kubectl.ForRunningPodCount(deploymentName, 4)
 
+			oldRevision := kubectl.GetRevision(deploymentName)
+			Expect(oldRevision).NotTo(BeEmpty())
+
 			By("triggering rolling update by changing image")
 			yamlV2 := fixtures.PrefillDecode(deploymentName,
 				fixtures.Role{Replicas: 2, Image: "registry.k8s.io/pause:3.10"},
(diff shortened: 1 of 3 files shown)


hasB4K (approved):
Hey there! Thanks for the fix!

I'm not sure if we need to add `TestEvalSingleActiveRevision` as a new test. But I'm also not against it.

LGTM

PR #871, opened 2026-06-02
[DisaggregatedSet][5/N] add the helm chart

yankay:
#### What type of PR is this?

/kind feature

Part of #789. The DisaggregatedSet controller is already bundled in the manager binary; this PR brings the chart artifacts (CRD, RBAC, webhook) into the unified `charts/lws` so users can deploy LeaderWorkerSet and DisaggregatedSet with a single `helm install`.

- Add the DisaggregatedSet CRD under `charts/lws/crds/`, alongside the existing LWS CRD. Same layout as the sibling jobset chart.
- Add `disaggregatedset_{admin,editor,viewer}_role.yaml` and a `vdisaggregatedset.kb.io` entry on the existing `lws-validating-webhook-configuration`. These are gated by a new `enableDisaggregatedSet` value (default `false`).
- Extend the manager `ClusterRole` with DisaggregatedSet verbs unconditionally — the bundled controller always starts and needs them.
- Fix the validating webhook path / rule `apiVersions` to `v1`. The standalone chart used `v1alpha1`, which does not match the `+kubebuilder:webhook` marker in `pkg/webhooks/disaggregatedset/disaggregatedset_webhook.go` or the API package under `api/disaggregatedset/v1/` — all admission requests would have 404'd.
- Split the `crds` Makefile target into one file per CRD, with `test -s` guards so a missing CRD fails the build instead of silently writing an empty file. The old target wrote both CRDs into the LWS filename.
- Mark `disaggregatedset/charts/disaggregatedset` as deprecated. Removal is the final step of #789.

**CRD layout note.** The CRD ships under `crds/` (jobset-style), not `templates/crd/` (kueue-style). Helm only installs `crds/` content during `helm install`, so the chart README documents a one-time `kubectl apply` before `helm upgrade` from a release that didn't ship the CRD. Migrating to `templates/crd/` would change the upgrade contract for the LWS CRD too and is intentionally out of scope.

Part of #789

The diff looks large (~17k lines) but ~17k of that is the auto-generated DisaggregatedSet CRD yaml moved into `charts/lws/crds/`. Hand-written changes are roughly 150 lines across `templates/`, `values.yaml`, README, and the Makefile.

Verified on a fresh kind v1.35 cluster: install with the flag off → only LWS resources, LWS sample reconciles; `helm upgrade` with the flag on → webhook + 3 ClusterRoles appear; a DisaggregatedSet is admitted and reconciled into two child LeaderWorkerSets.

```release-note
Helm: the `charts/lws` chart now ships the DisaggregatedSet CRD, RBAC, and validating webhook. Set `enableDisaggregatedSet=true` to install the editor/viewer/admin ClusterRoles and the validating webhook entry. The standalone `disaggregatedset/charts/disaggregatedset` chart is deprecated.
```


ahg-g:
/approve

thanks!

Issue #848, opened 2026-05-06
[DisaggregatedSet] Placement Policy


Edwinhr716:
**What would you like to be added**: 

Add placement policies to guarantee the following:

 - All pods in a disaggregatedSet unit can be scheduled on the same topology. Similar to exclusive topology in LWS.
- given two disaggregatedSet instances, they should be able to be scheduled on the same domain
- For a disaggregatedSet with two slices (see #846), these should be placed on different domains

This can be a new API field

```go
type DisaggregatedSetSpec struct {
	// +optional
PlacementPolicy *PlacementPolicyType `json:"placementPolicy,omitempty"`	
}

type PlacementPolicy struct {
	// +kubebuilder:validation:Enum={Default, ExclusivePlacement, ReplicaExclusive}
	// +kubebuilder:default=Default
	// +optional
	Type *PlacementPolicyType `json:"placementPolicyType,omitempty"`
	// +optional
	Topology string `json:"topology,omitempty"
}

type PlacementPolicyType string

const (
	// No guarantees on placement
	PlacementPolicyTypeDefault = "Default"
	
	// 1:1 placement between topology and disaggregatedSet replica
	PlacementPolicyTypeExclusive = "ExclusivePlacement"

	// DisaggregatedSet replicas from the same disaggregatedSet will be placed
	// on seperate domains. Conversely, disaggregatedSet replicas from different
	// disaggregatedSet objects can be placed together. 
	PlacementPolicyTypeReplica = "ExculsiveReplica"
)
```

This can be achieved by automatically injecting pod affinity and pod anti-affinity like we do in LWS for exclusive topology. 

For the `ExclusiveReplica` policy, we can make it so that the pod anti-affinity only applies on pods with the same disaggregatedSetName label. 

``` golang
corev1.PodAntiAffinityTerm{
  LabelSelector: &metav1.LabelSelector{MatchExpressions: []metav1.LabelSelectorRequirement{
	{
	 	Key:      podAffinityKey,
		Operator: metav1.LabelSelectorOpExists,
	,
	{
		Key:      podAffinityKey,
		Operator: metav1.LabelSelectorOpNotIn,
		Values:   []string{groupUniqueKey},
      },
	{
		Key:      podAffinityKey,
		Operator: metav1.LabelSelectorOpIn,
		Values:   []string{disaggregatedSetName},
      },
   }
  },
  TopologyKey: topologyKey,
}

```

[...]

PR #873, opened 2026-06-08
Add slices field to DisaggregatedSet

### LINKED ISSUE #846: [DisaggregatedSet] Add field to scale disaggregatedSet unit
**What would you like to be added**: Add a new field to make it possible to scale DisaggregatedSet unit

```go
	// +optional
	// +kubebuilder:default=1
	Slices *int32 `json:"replicas,omitempty"`
}
```

**Why is this needed**: Analogous to LWS replica, we need a way to scale the unit of DisaggregatedSet to deploy the same P/D combination multiple times, instead of needing to deploy a new yaml for the same combination.

**Completion requirements**:

- [X] Design doc
- [X] API change
- [ ] Docs update

BenjaminBraunDev:
## Summary

Adds `spec.slices` to DisaggregatedSet, an integer (default 1, minimum 1) that replicates the entire role topology into N independent copies. Each slice is a complete set of all roles (for example prefill and decode) that rolls out independently of the other slices.

Today a DisaggregatedSet describes exactly one copy of the topology, so running K identical copies means duplicating YAML. `slices` makes that a single field and is the foundation for per-domain placement policy (#848).

Fixes #846

## Design

- A slice is the outer scaling identity for revision-role LWS instances within a DisaggregatedSet. It is carried in a label (`disaggregatedset.x-k8s.io/slice`) and in resource names as `<ds>-<slice>-<revision>-<role>`, with slice before revision because a slice survives across rollouts while a revision is ephemeral.
- We chose the terminology 'slices' instead of 'replicas' to avoid overloading the term with LWS replicas.
- `slices` is a scale knob, not a template change. It is lives in `DisaggregatedSetSpec` at a higher level than `LeaderWorkerTemplate`, so changing it never triggers a rollout (a rolling update of any existing slices). Adding a slice brings up a fresh copy at the current revision, and removing a slice directly deletes the highest-index slice's resources (pods terminate via the normal grace period).
- Each slice is an independent rolling-update domain. The reconcile body now runs once per slice, so the existing rollout planner is reused and unchanged per slice.

## What changed

- **API** (`api/disaggregatedset/v1`): new `Slices *int32` field and `SliceLabelKey` constant.
- **utils** (`pkg/utils/disaggregatedset`): `GenerateName` and `GenerateLabels` take a slice and emit the slice label, plus a `GetSlices` accessor. `ComputeRevision` is unchanged (slices stay out of the hash).
- **controller**: `Reconcile` loops over `[0, slices)` and reconciles each slice independently, then a new `cleanupRemovedSlices` deletes LWS and services for slice indices at or above the desired count.
- **lws_manager / executor**: thread the slice through listing, creation, and the rolling-update calls. The planner (`planner.go`) is unchanged.
- **service_manager**: services are created per (slice, revision, role), with the slice in the name, labels, and selector, so role-to-role pairing stays within a slice. Adds `CleanupRemovedSlices` for scale-down.
- **generated + samples**: regenerated CRDs (both `config/crd/bases` and the helm chart copy) and deepcopy, and added `slices` to the sample.

## Testing

- Unit tests for the controller, executor, service manager, and utils, including new cases for multi-slice fan-out and slice scale-down.
- End-to-end manual testing on a GKE cluster with the controller and webhooks deployed:
  - `slices: 2` creates one LWS set, pods, and per-slice services per slice with correct names and labels.
  - Scaling slices down deletes only the removed slice, scaling up rebuilds it.
  - A template change rolls each slice independently while keeping a complete same-version set serving per slice.

## Not in scope

Placement policy (#848), rebalancing on failure, and any aggregate cross-slice service are left for follow-ups.

```release-note
Add scaling parameter "slices" to DisaggregatedSet CRD spec
```


BenjaminBraunDev:
NOTE: this is a naming change versus the pre-slices DS, which produced `<ds>-<revision>-<role>` with no slice segment. So upgrading an existing DS to this version renames its LWS (`<ds>-<rev>-<role>` to `<ds>-0-<rev>-<role>`), which means the controller would have to recreate them.

BenjaminBraunDev:
@hasB4K now that the KEP is in, could you take a look at this?

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -99,22 +101,23 @@ func (executor *RollingUpdateExecutor) initRollingUpdate(
...
 			replicas := 1
 			if roleLWS.Spec.Replicas != nil {
 				replicas = int(*roleLWS.Spec.Replicas)
 			}
-			if _, err := executor.LWSManager.SetInitialReplicas(ctx, disaggregatedSet.Namespace, lwsName, replicas); err != nil {
-				log.Error(err, "Failed to set initial-replicas annotation", "lws", lwsName)
+			// Address by the LWS's actual name so a legacy slice-0 object (whose name
+			// has no slice segment) is updated rather than missed.
+			if _, err := executor.LWSManager.SetInitialReplicas(ctx, disaggregatedSet.Namespace, roleLWS.Name, replicas); err != nil {
+				log.Error(err, "Failed to set initial-replicas annotation", "lws", roleLWS.Name)
```
hasB4K, line 112: Not sure to understand this?

api/disaggregatedset/v1/disaggregatedset_types.go
```diff
@@ -70,6 +73,14 @@ type DisaggregatedSetSpec struct {
...
 	// +required
 	Roles []DisaggregatedRoleSpec `json:"roles"`
+
+	// Slices is the number of independent copies of the whole role topology.
+	// Each slice is a complete set of all roles that rolls out independently.
+	// Changing Slices scales copies up or down and does not trigger a rollout.
+	// +optional
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=1
+	Slices *int32 `json:"slices,omitempty"`
```
hasB4K, line 84: I think it would be easier to support (with a deprecation warning) slice == 0 for the legacy case, with default being at 0 for the time being.

That way folks would simple add "slice: 1" in their deployment, and in the next version it would be gone.

pkg/controllers/disaggregatedset/service_manager.go
```diff
@@ -214,6 +213,57 @@ func (manager *ServiceManager) cleanupDrainedServices(
...
+			}
+			return fmt.Errorf("failed to delete service %s: %w", service.Name, err)
+		}
+	}
+
+	return nil
+}
+
+func GenerateServiceName(baseName string, slice int, revision, role string) string {
+	return fmt.Sprintf("%s-%d-%s-%s-prv", baseName, slice, revision, role)
```
hasB4K: If slice == 0, it would be easier to handle here no?

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -72,57 +74,130 @@ func (r *DisaggregatedSetReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 	// Step 3: Reconcile LWS objects.
 	executor := r.createRollingUpdateExecutor()
+	roleNames := disaggregatedsetutils.GetRoleNames(disaggregatedSet)
+
+	// Backward compatibility: when slices > 1, a pre-slices (label-less) slice-0
+	// deployment that still sits at the target revision must be migrated to the
+	// slice-aware form before any sibling slice is created, otherwise the legacy
+	// slice-agnostic service would also select the siblings' same-revision pods.
+	// While that migration runs we block the slice loop.
+	if sliceCount > 1 {
```
hasB4K, line 99: What happens if we do this:
* update from legacy to slice 2
* update from slice 2 to slice 1
  * if the rollout went fine
  * or mid rollout

Also, I think we need to avoid asking the user to multiply by two (slice 2) the size of their deployment just to no longer be in the legacy state. Using slice == 0 could be only be set in a few places, and avoid most of the current complexity.

It might requires of KEP update (in this PR) regarding the migration

hasB4K (commented):
Overall looks good, but the migration part looks a bit brittle I think.
Happy to chat on this if you think I'm wrong.

api/disaggregatedset/v1/disaggregatedset_types.go
hasB4K, line 84: We should probably add a maxItems by safety?

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -45,8 +47,8 @@ type DisaggregatedSetReconciler struct {
 // +kubebuilder:rbac:groups=disaggregatedset.x-k8s.io,resources=disaggregatedsets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=disaggregatedset.x-k8s.io,resources=disaggregatedsets/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=disaggregatedset.x-k8s.io,resources=disaggregatedsets/finalizers,verbs=update
-// +kubebuilder:rbac:groups=leaderworkersetv1.x-k8s.io,resources=leaderworkersets,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=leaderworkersetv1.x-k8s.io,resources=leaderworkersets/status,verbs=get
+// +kubebuilder:rbac:groups=leaderworkerset.x-k8s.io,resources=leaderworkersets,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=leaderworkerset.x-k8s.io,resources=leaderworkersets/status,verbs=get
```
hasB4K, line 51: why this change?

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -72,57 +74,130 @@ func (r *DisaggregatedSetReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+		if migrating {
+			return migResult, nil
+		}
+	}
+
+	var result ctrl.Result
+	for slice := range sliceCount {
+		sliceResult, err := r.reconcileSlice(ctx, executor, disaggregatedSet, slice, revision, roleNames)
+		if err != nil {
+			return sliceResult, err
```
hasB4K: Not sure this loop is quite right. If any slice errors, we skip the rest for this cycle. So a stuck slice at index 3 means slices 4+ never get reconciled, even though nothing is wrong with them.

Feels a bit off from the "slices roll independently" story in the KEP. Would it be better to collect the errors and keep going? Something like appending to an errs slice and returning errors.Join(errs...) at the end maybe?

BenjaminBraunDev:
> Overall looks good, but the migration part looks a bit brittle I think. Happy to chat on this if you think I'm wrong.

Yeah, agreed. Dropped the same-revision migration state machine and the blocking gate in favor of a simple recreate by deleting the legacy slice-0 + its service and letting the normal loop rebuild it in slice-aware form

pkg/controllers/disaggregatedset/executor.go
BenjaminBraunDev: This is the compatibility path. A pre-slices LWS is named `<ds>-<rev>-<role>` with no slice segment, so we address it by its original object name rather than a slice-aware name, otherwise the drain would miss it. Added a clarifying comment.

api/disaggregatedset/v1/disaggregatedset_types.go
BenjaminBraunDev: The issue with default 0 is every new DS is born with legacy naming and recreates the first time slices is set. Default 1 means new DS are slice-aware from creation. Legacy is detected from existing label-less objects and adopted in place. I simplified the transition to be non-blocking and just delete old legacy lws upon slice update.

pkg/controllers/disaggregatedset/service_manager.go
BenjaminBraunDev: Yes, though I just removed this. It was the old naming path and buildService now derives the name from the LWS (lws.Name + "-prv"), so there's no slice-0 case to special-case here.

api/disaggregatedset/v1/disaggregatedset_types.go
BenjaminBraunDev: That's probably a good safety measure haha, added this with max 100

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
BenjaminBraunDev: Oh good call, fixed loop to collects per-slice errors and return errors.Join() so a stuck slice doesn't block

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
BenjaminBraunDev: The old marker used leaderworkersetv1.x-k8s.io ([controller.go#L48-L49](https://github.com/kubernetes-sigs/lws/blob/aca57665186a6626000f59c7a0ce4db75fcaae60/pkg/controllers/disaggregatedset/disaggregatedset_controller.go#L48-L49)), which isn't a real API group. The actual LWS group is leaderworkerset.x-k8s.io ([groupversion_info.go#L29](https://github.com/kubernetes-sigs/lws/blob/aca57665186a6626000f59c7a0ce4db75fcaae60/api/leaderworkerset/v1/groupversion_info.go#L29))

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
BenjaminBraunDev: Reworked this: dropped the same-revision surge migration and the blocking gate (the old migrateLegacySlice0 / finishLegacyMigration). The new recreateLegacySlice0 is non-blocking and just deletes the legacy slice-0 objects and lets the normal reconcile recreate it slice-aware.

- **No "multiply by 2":** when `slices` goes above 1 over a legacy slice-0, we just delete it (Service first, then the LWS) and the normal reconcile recreates it slice-aware next to the siblings. One slice-0 restart, no surge or doubling.
- **legacy → 2 → 1:** there's no migration state machine anymore, so nothing gets caught mid-flight. The `2 → 1` is just the normal slice teardown.
- Kept the default at `1` (not `0`) so new DisaggregatedSets are slice-aware from creation. Default `0` would make every new DS legacy and recreate on the first `slices` change.

Updated the KEP migration section to match

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
hasB4K: I guess it was a typo, thanks!

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
hasB4K: Why do we need sliceCount > 1? I think I don't understand this part

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -72,57 +74,132 @@ func (r *DisaggregatedSetReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 	return result, nil
 }
 
+// soonerRequeue keeps the soonest non-zero RequeueAfter across slices.
+func soonerRequeue(a, b ctrl.Result) ctrl.Result {
+	if b.RequeueAfter > 0 && (a.RequeueAfter == 0 || b.RequeueAfter < a.RequeueAfter) {
+		return b
+	}
+	return a
+}
```
hasB4K: nit: could this be in an util maybe? with a naming slightly more explicit

pkg/controllers/disaggregatedset/disaggregatedset_controller_test.go
```diff
@@ -129,11 +157,271 @@ func TestScalingWithoutRollingUpdate(t *testing.T) {
...
+	// Slice 1 (>= desired) is deleted.
+	s1p, _ := lwsManager.Get(ctx, disaggregatedSet.Namespace, disaggregatedsetutils.GenerateName(disaggregatedSet.Name, 1, revision, testControllerRolePrefill))
+	assert.Nil(t, s1p, "slice 1 prefill should be deleted")
+	s1d, _ := lwsManager.Get(ctx, disaggregatedSet.Namespace, disaggregatedsetutils.GenerateName(disaggregatedSet.Name, 1, revision, testControllerRoleDecode))
+	assert.Nil(t, s1d, "slice 1 decode should be deleted")
+}
+
+// TestLegacyAdoptedInPlace: a single-slice DisaggregatedSet from a pre-slices release
+// (label-less slice-0 LWS at the target revision) is adopted in place, not duplicated.
+func TestLegacyAdoptedInPlace(t *testing.T) {
```
hasB4K, line 275: So this test check that a pre slice DS is not updated right?
Could you add a test that check that if we update a legacy DS (by modifying the podTemplate by adding an env var, or bumping an image) it will become a DS slice aware?

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -233,15 +310,54 @@ func (r *DisaggregatedSetReconciler) cleanupDrainedLWS(ctx context.Context, disa
...
+// recreateLegacySlice0 converts a pre-slices (label-less) slice-0 deployment to the
+// slice-aware form when slices is increased above 1. The legacy slice-0 uses legacy names
+// and a slice-agnostic service (selector {name, role, revision}) that would also select
+// the new siblings' same-revision pods. For each role it deletes the legacy service first
+// (so a partial failure leaves the LWS present and this retries, and the service never
+// lingers alongside siblings), then the legacy LWS. The caller runs this before the slice
+// loop, which then recreates slice 0 in slice-aware form. This restarts slice 0 once,
+// which we accept in place of an in-place migration. A legacy slice-0 at an older revision
+// is left alone: the normal rolling update migrates it without a restart.
+func (r *DisaggregatedSetReconciler) recreateLegacySlice0(
```
hasB4K, line 343: Could you add a TODO in the spot that should be removed in the 0.11.0 LWS version?
In my understanding, 0.10.0 will have the slice feature, and in 0.11.0 (or maybe later) we will remove the legacy version.

hasB4K (approved):
The only blocking point on my side:
> Could you add a test that check that if we update a legacy DS (by modifying the podTemplate by adding an env var, or bumping an image) it will become a DS slice aware?

Approving otherwise

hasB4K:
/lgtm
I'm LGTM-ing to unblock you, but please add the missing test for my peace of mind 😇 - it should be fairly easy

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
BenjaminBraunDev: no problem!

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
BenjaminBraunDev: From a discussion I had with @Edwinhr716, he mentioned it would be unexpected behavior if updating DS to support slices caused a deletion / recreate, thus we only trigger the migration when the user sets the slice count above the default, since the legacy objects work fine right up until the slice count is > 1

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
BenjaminBraunDev: Yeah that sounds good

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
BenjaminBraunDev: Renamed it to earliestRequeue for clarity. I left it in the controller package rather than a util since it's specific to the reconcile loop (just merges ctrl.Result to keep the soonest requeue across slices) and isn't used anywhere else. Though still happy to move it if you'd prefer.

pkg/controllers/disaggregatedset/disaggregatedset_controller_test.go
BenjaminBraunDev: Yes, done

Edwinhr716:
/lgtm

Will let @hasB4K do a final review since you added a test before approving

hasB4K (approved):
/lgtm Thanks for the update

PR #813, opened 2026-04-16
docs(kep): draft fail-fast restart budget and init-phase DNS for LWS

panpan0000:
## Summary
- Add KEP draft for LWS fail-fast restart budget with terminal  condition.
- Add optional  support for init-phase peer DNS.
- Prioritize fail-fast as primary motivation; DNS support is secondary.

## Key Points
- Add  (opt-in, nil default).
- Add new LWS condition type .
- Keep current behavior unchanged when unset.
- Reuse existing LWS env vars; no new env introduced.

## Notes
- This commit is docs-only and scoped to .

relevant to https://github.com/kubeflow/trainer/pull/3417

Co-Author: GPT-5.4

PR #877, opened 2026-06-09
Add bounded group recovery for LeaderWorkerSet

panpan0000:
#### What type of PR is this?

/kind feature
/kind api-change

This PR implements the bounded group recovery part of KEP-820.

- Adds the optional `spec.leaderWorkerTemplate.maxGroupRestarts` field for `RecreateGroupOnPodRestart` and `RecreateGroupAfterStart`.
- Counts controller-initiated group recreation per Pod-template revision and replica ordinal. The exhaustion attempt is not counted because it does not start another recreation.
- When the budget is exhausted, LWS marks the replica degraded, adds cleanup finalizers to the leader and current worker Pods that can still receive them, and deletes the leader with foreground propagation. This terminates the group and lets Kubernetes release its scheduled resources after the Pods become terminal, while retaining those Pod API objects for inspection.
- Retained Pod objects do not guarantee native container logs. Workloads that need durable diagnostics should export logs externally.
- Recovery is explicit and per group: set `leaderworkerset.sigs.k8s.io/recover=true` on the retained leader Pod. LWS clears only that revision/replica counter and removes the worker finalizers before the leader finalizer.
- Editing `maxGroupRestarts` does not recover an exhausted group or create a new Pod-template revision. A recovery annotation that predates exhaustion is cleared.
- LWS deletion, actual scale-down, and rollout are teardown operations. They release cleanup finalizers without treating deletion as recovery. Active MaxSurge ordinals retain their counters and remain subject to the budget until the leader StatefulSet removes them.
- The validating webhook rejects `maxGroupRestarts` with `groupIdentity: Hash`, because a recreated Hash group has no stable ordinal identity for carrying the counter.
- LWS always publishes the `Degraded` condition; it is `False` with reason `AsExpected` when no replica has exhausted its budget.
- If the exhausted replica is the only unfinished work, LWS reports `Progressing=False`. Other replicas or a separate rollout can still make progress, so `Degraded=True` and `Progressing=True` may coexist.

Init-phase DNS and headless Service lifecycle changes are not part of this PR.

Implements KEP-820 from https://github.com/kubernetes-sigs/lws/pull/813

- This is a circuit breaker for LWS-initiated group recreation. It does not change kubelet container restart behavior or make the whole LWS terminal.
- Finalizers retain terminating Pod API objects, not running containers, resource reservations, or durable logs. A Pod that was already deleting when exhaustion was detected cannot gain a new finalizer and may disappear.
- Restart counts use one LWS-level JSON map keyed by `<revision>/<groupIndex>`. The exhausted marker is stored on the leader Pod; cleanup finalizers are stored on the leader and retained worker Pods.
- The exhaustion behavior is fixed. This PR does not add a `Retain` policy or the ownership redesign discussed in #956.
- Generated API, CRD, Helm CRD, apply-configuration, and reference documentation are included.

#### Test plan

- `CGO_ENABLED=0 go test ./api/... ./pkg/... ./cmd/...`
- `CGO_ENABLED=0 go vet ./api/... ./pkg/... ./cmd/...`
- Controller integration coverage for zero and one restart budgets, both restart policies, exhausted groups remaining stopped after spec edits, explicit annotation recovery, and LWS deletion
- Webhook integration coverage for the restart-policy and group-identity matrices
- E2E coverage for one allowed recreation, budget exhaustion, terminal retained Pod objects, exact counter preservation, and annotation-based recovery
- `CGO_ENABLED=0 gmake verify`

```release-note
LeaderWorkerSet adds `spec.leaderWorkerTemplate.maxGroupRestarts`. When a replica exhausts this budget, LWS terminates the group, retains terminal Pod API objects that can receive cleanup finalizers, stops automatic group recreation, and reports `Degraded=True` with reason `ReplicaRestartBudgetExceeded`. Set `leaderworkerset.sigs.k8s.io/recover=true` on the retained leader Pod to reset that replica's budget and resume recovery.
```


panpan0000:
Hi, all , it's ready for review now.

panpan0000:
ping to keep PR alive

panpan0000:
Updated after a long hiatus — this PR is rebased onto latest main and all conflicts are resolved (CI should go green again). Summary of the changes since the last push:

**Rebase / conflict resolution**
- Rebased onto main (2026-08). The only non-trivial resolution is in `pkg/controllers/leaderworkerset_controller_test.go`, keeping both `TestEnqueueLWSRequests` (from #957) and the new Failed-condition tests.

**Backward compatibility fix: publishNotReadyAddresses now defaults to true**
- main hardcodes `PublishNotReadyAddresses: true` on LWS-owned headless Services (since #197). Defaulting the new field to `false` would have flipped existing Services on upgrade and broken init-phase DNS for current users.
- The field is now `*bool`, unset/absent = true (historical behavior), `false` = explicit opt-out.
- KEP doc in this PR updated to match.

**Init-phase DNS timing**
- Under `UniquePerReplica`, the per-replica headless Service was previously created lazily by the pod reconciler, so `LWS_LEADER_ADDRESS` could NXDOMAIN early in init. The LWS reconciler now creates it eagerly for every leader pod.

**Cleanups**
- Dropped a stale `config/rbac/role.yaml` artifact (it referenced a non-existent `leaderworkersetv1.x-k8s.io` API group) — the feature needs no RBAC changes.
- Fixed the `MaxGroupRestartsExceeded` event to carry the actual message instead of the literal `Delete` string.

@yankay @kerthcet @edwinhr716 @ahg-g — would you mind taking a look? This is the implementation for the merged KEP-820 (#813). Thanks!

panpan0000:
One more round of self-review fixes pushed just now:

1. **Cap the restart counter at the exhausted transition**: previously every failure event from a crash-looping pod re-incremented the `group-restart-count` annotation (unbounded growth + a pair of patches per event). Now the counter only bumps once, on the `count == maxGroupRestarts` transition that surfaces the Failed condition; further failures are no-ops. Added a unit test case for the already-exceeded path.
2. **Locked in the backward-compat default with a unit test**: `TestCreateHeadlessServicePublishNotReadyAddresses` covers nil NetworkConfig / nil field -> true (the historical hardcoded value), explicit true/false on create, and the patch-back-to-true path when the field is cleared on an existing Service.
3. **KEP doc**: clarified that counter persistence uses both the pod-level `group-restart-count` and the LWS-level `group-restart-counts` annotations, and that init-phase DNS publication stays on by default as it always has been.

Local verification: `go build ./...`, `go test ./pkg/... ./api/...`, `go vet ./...` all green.

panpan0000:
Addressed the review feedback (thanks!):

1. **Counter fallback ambiguity** — the budget read path now takes `max(persisted, pod-annotation)` instead of special-casing "persisted == 0 means absent", and `syncLeaderRestartCountAnnotation` now only raises the pod annotation, never downgrades it. Either side of the (deliberately non-atomic) patch pair lagging behind can no longer undercount the budget. Unit tests added for both drift directions.
2. **Unfiltered leader-pod watch** — added `builder.WithPredicates`: create/delete events still enqueue (eager per-leader Service, group recreation), and update events enqueue **only when the group-restart-count annotation changed**. This keeps the Failed condition surfacing promptly (the pod annotation patch is the reliable trigger — the concurrent LWS annotation patch may be reconciled against a stale pod cache) while dropping kubelet status churn from the hot path. A create-only predicate would have delayed Failed until some unrelated event, so updates on annotation change are intentional.
3. **Non-atomicity documented in code** — added a comment on `incrementGroupRestartCount` referencing the KEP's best-effort caveat; skipped `retry.RetryOnConflict` since this codebase doesn't use that pattern anywhere.
4. Removed the stale `ownerRefWrapper` comment in `pod_controller_test.go`.

Local: `go build ./...`, `go test ./pkg/... ./api/...`, `go vet ./...` all green.

panpan0000:
Hi @ahg-g @Edwinhr716, 
this PR implements the merged KEP-820 (fail-fast restart budget + init-phase DNS), is rebased , CI test green (e2e/integration running), and has incorporated all feedback so far 

 would you mind taking a look when you get a chance?

panpan0000:
The `pull-lws-test-integration-main` failure was a test-expectation mismatch caused by the new CRD schema default: the apiserver now materializes `publishNotReadyAddresses: true` on created objects, while the test wrapper built expected objects with a nil field. Fixed by mirroring the default in `wrappers.BuildLeaderWorkerSet`/`SubdomainPolicy()`, and the full integration suite (both suites, 96 webhook specs + controllers) now passes locally. Pushed as `1c51ea2`.

panpan0000:
Also ran the 4 new e2e tests on my local OrbStack cluster with this branch's controller deployed: 
`publishNotReadyAddresses` propagation for both Shared and UniquePerReplica, plus the two `maxGroupRestarts` -> Failed scenarios. All green.

panpan0000:
@yankay , for your question "What's the benifit of my PR besides preflight use case, and can STS has same problem ? "

| Problem | Plain StatefulSet | LWS (default `RecreateGroupOnPodRestart`) |
|---|---|---|
| Unbounded restarts | Yes: CrashLoopBackOff throttles the *rate* (capped at 5min) but never caps the *count*; restarts forever | Yes: group recreated forever |
| Loses previous logs | No: pod object persists, `logs --previous` returns the last crash's logs | Yes: pod is deleted/recreated, `--previous` → `previous terminated container not found` |
| Restart counter | Yes: `containerStatus.restartCount` accumulates (observed 0→1→2→3) | No: resets to 0 on every recreation; nothing records group restarts |
| Backoff | Yes: kubelet exponential backoff | No: pod deletion resets kubelet backoff; whole group recreated every ~54s in my test |
| Terminal failure state | None | None (Available/Progressing flap forever) |

yankay:
Thanks for this improvement! I noticed two P1 correctness issues:

- Legacy revisions may omit `publishNotReadyAddresses`, while upgraded objects default it to `true`. Revision comparison can therefore trigger an unintended rollout. Could we normalize missing to `true`, preserve explicit `false`, and add a legacy-revision regression test?
- With `UniquePerReplica`, a Pod may start before its Service exists, causing init-time NXDOMAIN. Could we create per-replica Services before applying the StatefulSet, including scale-up and MaxSurge ordinals, while retaining bounded DNS retries?

I think these should be addressed before merge. Thanks! :)

panpan0000:
@yankay Thanks for catching these. Both issues are addressed in ad9d95b: 
legacy revisions normalize a missing `publishNotReadyAddresses` to true while preserving explicit false, 
and `UniquePerReplica` Services are prepared before StatefulSet apply for scale-up/MaxSurge, 
with lifecycle regression tests added. Unit, integration, verify, and main E2E checks are green

panpan0000:
@yankay Follow-up: the first upgrade rerun exposed an overly strict test validator, not a workload regression. Legacy and normalized revisions are semantically equal, so the old StatefulSet hash is intentionally retained to avoid a rollout. That validator is fixed in 7b532de; the upgrade E2E rerun is now pending.

panpan0000:
@yankay I found the upgrade validator had the same legacy-hash assumption for worker StatefulSets as for the leader StatefulSet. That compatibility check is now fixed in 359b59e as well, and the upgrade job has been re-triggered.

panpan0000:
@yankay Thanks for catching both P1 issues. Both controller fixes and regression tests are in place: legacy missing `publishNotReadyAddresses` normalizes to true (explicit false preserved), and UniquePerReplica Services are created before StatefulSet apply for scale-up/MaxSurge. The upgrade reruns then exposed legacy-hash assumptions in the test validators; those are now fixed through `303eb3b`, and the upgrade job has been triggered again.

panpan0000:
Follow-up review cleanup in `b37a62b`: nil `networkConfig.subdomainPolicy` now safely defaults to Shared in controller paths; persisted negative restart counts are rejected; and the misleading publishNotReady/default and nil maxGroupRestarts test names are corrected. Focused controller/webhook/revision tests pass.

panpan0000:
I just realized  that my code about `publishNotReadyAddresses` should be **removed** from both the KEP and this PR.
  
LWS already creates its headless Services with `PublishNotReadyAddresses: true`, which is exactly the behavior required for init-phase peer DNS. 
The actual correctness gap is the UniquePerReplica ordering race:  the Pod can start before its per-replica Service exists.
So the changing of supporting an explicit false does not advance the preflight goal, can disable that goal when used, and adds API/CRD surface plus revision-upgrade compatibility without a concrete user requirement.

So I will keep the historical hardcoded true, (retaining Service-before-StatefulSet creation for initial replicas, scale-up, and MaxSurge, and removing the new field, propagation/update logic, revision normalization, generated API artifacts, docs, and toggle-specific tests. If a real opt-out use case emerges, it can be proposed separately with its own semantics and compatibility contract.)




I've just update the PR

panpan0000:
@yankay After revisiting KEP-820 from first principles, I removed `spec.networkConfig.publishNotReadyAddresses` from the KEP and PR.

LWS already creates its headless Services with `PublishNotReadyAddresses: true`; that historical behavior is exactly what init-phase peer DNS needs. The correctness gap is the `UniquePerReplica` ordering race, so the fix keeps Service creation before the StatefulSet for initial replicas, scale-up, and MaxSurge ordinals.

Supporting an explicit `false` did not advance the preflight goal and added API/revision compatibility surface without a concrete use case. The API/CRD/applyconfig/docs, false/toggle tests, and legacy-revision normalization have been removed. The change is pushed as `93b8681`.

panpan0000:
@yankay  PTAL

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -103,6 +103,15 @@ const (
...
 	RecreateGroupAfterStartAnnotationKey string = "leaderworkerset.sigs.k8s.io/experimental-recreate-group-after-start"
+
+	// GroupRestartCountAnnotationKey records the number of times the leader pod of a
+	// group has been recreated via the RecreateGroupOnPodRestart path. It is used
+	// to enforce an opt-in restart budget (spec.leaderWorkerTemplate.maxGroupRestarts).
+	GroupRestartCountAnnotationKey string = "leaderworkerset.sigs.k8s.io/group-restart-count"
+
+	// GroupRestartCountsAnnotationKey stores per-group restart counts on the LWS object
+	// so the budget survives leader pod recreation.
+	GroupRestartCountsAnnotationKey string = "leaderworkerset.sigs.k8s.io/group-restart-counts"
```
ahg-g, line 110: I am concerned about the subtle difference in the name (s at the end), may cause bugs in the future.

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -103,6 +103,15 @@ const (
...
 	// all pods in the group are not pending
 	RecreateGroupAfterStartAnnotationKey string = "leaderworkerset.sigs.k8s.io/experimental-recreate-group-after-start"
+
+	// GroupRestartCountAnnotationKey records the number of times the leader pod of a
+	// group has been recreated via the RecreateGroupOnPodRestart path. It is used
+	// to enforce an opt-in restart budget (spec.leaderWorkerTemplate.maxGroupRestarts).
+	GroupRestartCountAnnotationKey string = "leaderworkerset.sigs.k8s.io/group-restart-count"
+
+	// GroupRestartCountsAnnotationKey stores per-group restart counts on the LWS object
+	// so the budget survives leader pod recreation.
```
ahg-g: pls document the format used here for storing the per-group restart counts

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -417,6 +437,13 @@ const (
 	// is true when the lws is in upgrade process after the (leader/worker) template is updated. If only replicas is modified, it will
 	// not be considered as UpdateInProgress.
 	LeaderWorkerSetUpdateInProgress LeaderWorkerSetConditionType = "UpdateInProgress"
+
+	// LeaderWorkerSetFailed means lws has reached a terminal failure state. This
+	// is currently set when a group has exhausted the restart budget configured
+	// via spec.leaderWorkerTemplate.maxGroupRestarts. Existing built-in
+	// conditions (Available, Progressing, UpdateInProgress) remain orthogonal;
+	// callers must tolerate this new type safely.
```
ahg-g: pls document the state of LWS when this state is reached. Does this mean the whole LWS is marked as failed or the specific replica that is continuously failing?

Also, mention how an operator can recover from this failed state.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -162,6 +165,14 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, err
 	}
 
+	// Services must exist before the StatefulSet can create Pods. In
+	// UniquePerReplica mode this includes scale-up and MaxSurge ordinals.
```
ahg-g: > Services must exist before the StatefulSet can create Pods.

> peer DNS FQDN publication during the init-container phase depends on a per-replica Service that is created lazily (after the leader pod appears), so early init containers can hit NXDOMAIN races. It applies to any workload with init-phase peer communication and to any repeated group recreation caused by failures in init-containers or main containers.

Currently creating the headless service for the SubdomainUniquePerReplica case is a fast follow to creating the pod object and it is actually done before creating the workers sts, it is not tied to kubelet starting the leader pod. I am not sure I understand how creating the services before creating the leader pods sustainably address the issue of delayed DNS record. If we want to ensure order, why not block creating the workers sts on having the associated headless service present in pod_controller.go; just like we defer creating the workers sts until the leader is ready at https://github.com/kubernetes-sigs/lws/blob/main/pkg/controllers/pod_controller.go#L143 for the LeaderReadyStartupPolicy case.

ahg-g (commented):
Thanks @panpan0000 , this PR implements two independent features, I suggest we split the PR into two to address each one separately. I have some suggestions for each, but I feel we are mixing things up when implemented in the same PR.

api/leaderworkerset/v1/leaderworkerset_types.go
ahg-g: It seems this means the whole LWS is considered failed, I am not sure this is the semantics we want for a deployment-like API where the replicas are independent entities. Remember this is not a Job API that is expected to run to completion (either successfully or not). However, I think that terminating a consistently failing replica (continuously landing on a faulty accelerator) could be reasonable as long as there is a way for the operator to recover from the failure via manual intervention perhaps?

panpan0000: good point @ahg-g  , I mixed up the group and whole LWS ......

(1) "stop group meaningless looping restart" is one thing, it's controller logic.
(2) "reflect group failure clearly in somewhere" is another thing, which is state logic.

for second one , maybe LWS will end up with :
```
Degraded=True
Reason=ReplicaRestartBudgetExceeded
Message="Replica 0 exhausted maxGroupRestarts"
```

pkg/controllers/leaderworkerset_controller.go
panpan0000: hmmm........ 

At the very beginning, I was concerning `UniquePerReplica` not avail during init-container phase...
But with default `publishNotReadyAddresses=True`, this concern has gone.

Second concern is the timing for DNS resolve them .. 

but as you said @ahg-g , timing window now is unavoidable .. maybe adding  waiting and retrying in preflight code scripts make more sense.... ? ( I will split PR or remove that part... Let me think more a bit)

panpan0000:
> Thanks @panpan0000 , this PR implements two independent features, I suggest we split the PR into two to address each one separately. I have some suggestions for each, but I feel we are mixing things up when implemented in the same PR.

Thanks, I removed the init-phase DNS and Service lifecycle changes, so this PR now only handles bounded group recovery.

api/leaderworkerset/v1/leaderworkerset_types.go
panpan0000: renamed to `GroupRestartCountsAnnotationKey`

api/leaderworkerset/v1/leaderworkerset_types.go
panpan0000: done

pkg/controllers/leaderworkerset_controller.go
panpan0000: I removed the DNS and eager Service creation changes from this PR, and that work can be discussed separately.

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -104,6 +104,19 @@ const (
 	// all pods in the group are not pending
 	RecreateGroupAfterStartAnnotationKey string = "leaderworkerset.sigs.k8s.io/experimental-recreate-group-after-start"
 
+	// GroupRestartCountsAnnotationKey stores a JSON map of restart budget consumed by
+	// controller-initiated group recreation. Keys use "<revision>/<groupIndex>" and
```
ahg-g: We have a new deployment mode where the groups don't have an ordinal index: https://github.com/kubernetes-sigs/lws/tree/main/keps/898-group-identity

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -219,6 +232,19 @@ type LeaderWorkerTemplate struct {
 	// +optional
 	RestartPolicy RestartPolicyType `json:"restartPolicy,omitempty"`
 
+	// maxGroupRestarts bounds how many times the controller can recreate a group
+	// via the RecreateGroupOnPodRestart path. Once exhausted, the controller keeps
+	// the failed group for debugging and stops recreating it. Deleting the retained
+	// leader pod resets that revision/group's budget and allows recovery. It is
+	// opt-in: when unset (nil), the existing unbounded recreation behavior is
```
ahg-g: pls document without reference to past behavior, something like: "when unset (nil), max restarts is unlimited".

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -219,6 +232,19 @@ type LeaderWorkerTemplate struct {
...
+	// preserved. This field is only valid when
+	// spec.leaderWorkerTemplate.restartPolicy is RecreateGroupOnPodRestart; the
```
ahg-g: Why can't we support RecreateGroupAfterStart as well?

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -104,6 +104,19 @@ const (
 	// all pods in the group are not pending
 	RecreateGroupAfterStartAnnotationKey string = "leaderworkerset.sigs.k8s.io/experimental-recreate-group-after-start"
 
+	// GroupRestartCountsAnnotationKey stores a JSON map of restart budget consumed by
```
ahg-g: for each of the below annotations or finalizer, pls explicitly mention on which object(s) they are set: disaggset, lws, leaders sts/deployment, workers sts, leader pod, worker pod

pkg/controllers/pod_controller.go
```diff
@@ -118,9 +121,19 @@ func (r *PodReconciler) reconcilePod(ctx context.Context, req podReconcileReques
 	// get the leaderWorkerSet object
 	var leaderWorkerSet leaderworkerset.LeaderWorkerSet
 	if err := r.Get(ctx, types.NamespacedName{Name: lwsName, Namespace: pod.Namespace}, &leaderWorkerSet); err != nil {
+		if apierrors.IsNotFound(err) && hasFinalizer(&pod, leaderworkerset.GroupRestartBudgetCleanupFinalizer) {
```
ahg-g: should this condition also consider the case when the lws DeletionTimestamp is set? Remember that if lws is being deleted with foreground policy, it will always exist until all child objects are also deleted.

keps/820-distributed-preflight-check/README.md
```diff
@@ -1,237 +1,328 @@
...
-### Story
+### Goals
 
-When users run distributed pre-checks (for example NCCL tests) in LWS init-containers,
-a failure can push the group into an infinite recreate loop in the `RecreateGroupOnPodRestart` path.
-This KEP adds a fail-fast boundary after N group recreation attempts, and also ensures
-leader FQDN can be resolved during init phase when explicitly enabled.
+1. Stop LWS-initiated group recreation after a user-selected number of
+   attempts.
+2. Retain the current failed group for diagnosis after the budget is exhausted.
```
ahg-g, line 64: What is being retained exactly? what happens to the resources allocated to the reset of the group? these resources must be released.

keps/820-distributed-preflight-check/README.md
```diff
@@ -1,237 +1,328 @@
...
+3. Report partial failure without making the whole LWS terminal.
```
ahg-g, line 65: Can an operator resume a specific failed group after the budget is exhausted? I think this is needed.

pkg/controllers/pod_controller.go
```diff
@@ -118,9 +121,19 @@ func (r *PodReconciler) reconcilePod(ctx context.Context, req podReconcileReques
...
+			return ctrl.Result{}, r.removeGroupRestartBudgetFinalizer(ctx, &pod)
+		}
 		// If lws not found, it's mostly because deleted, ignore the error as Pods will be GCed finally.
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	if podutils.LeaderPod(pod) && pod.DeletionTimestamp != nil &&
```
ahg-g, line 140: What are we trying to do here exactly? this code seems to be allowing any failed group to restart: it lets the leader to be processed with the deletion unconditionally, which will trigger the parent sts/deployment to recreate it again. Is this intended? I thought we wanted to "freeze" a failed group with exhausted budget.

pkg/controllers/pod_controller.go
```diff
@@ -362,6 +375,36 @@ func (r *PodReconciler) handleRestartPolicy(ctx context.Context, pod corev1.Pod,
...
+			if err != nil {
+				return false, fmt.Errorf("marking exhausted restart budget for %s: %w", leader.Name, err)
+			}
+			if marked {
+				r.Record.Eventf(&leaderWorkerSet, &leader, corev1.EventTypeWarning, "ReplicaRestartBudgetExceeded",
+					"Stopped recreating group %s after %d controller-initiated restarts", groupIndex, count)
+			}
+			return false, nil
+		}
+		if leader.Annotations[leaderworkerset.GroupRestartBudgetExhaustedAnnotationKey] == "true" {
```
ahg-g: I am not following why we are doing this (recovering the group). What case does this code handle?

ahg-g (commented):
This is a great feature! I left a number of comments that hopefully make sense.

I also think we need proper integration test coverage for this, basically for the table documenting the behavior you have in the KEP.

api/leaderworkerset/v1/leaderworkerset_types.go
panpan0000: Good catch. I now reject `maxGroupRestarts` with `groupIdentity: Hash` because a recreated Hash group has no stable identity for carrying the counter, and added webhook tests for LWS and DisaggregatedSet.

api/leaderworkerset/v1/leaderworkerset_types.go
panpan0000: Updated the API comment: when `maxGroupRestarts` is nil, group recreation is unlimited.

api/leaderworkerset/v1/leaderworkerset_types.go
panpan0000: Supported now. `maxGroupRestarts` works with `RecreateGroupAfterStart` too, with unit and controller integration coverage.

api/leaderworkerset/v1/leaderworkerset_types.go
panpan0000: Updated the API comments and KEP: the count map is on the LWS, while the exhausted marker and cleanup finalizer are on the retained leader Pod.

pkg/controllers/pod_controller.go
panpan0000: Fixed. If the LWS is being deleted, the Pod controller only removes its cleanup finalizer; it does not clear the count or start group recovery, and this has unit and integration coverage.

keps/820-distributed-preflight-check/README.md
panpan0000: I clarified that the whole group is retained, so its resources remain reserved. Safely releasing the other Pods while preserving the actual failed Pod needs broader ownership or suspension semantics, especially for Ordinal groups, so I linked #956 instead of adding partial deletion here.

keps/820-distributed-preflight-check/README.md
panpan0000: Yes. Deleting that group's retained leader is the explicit per-group recovery action; it clears only that revision/group counter, and the integration test now covers this flow.

pkg/controllers/pod_controller.go
panpan0000: Fixed. Spec or Pod updates keep the group frozen; only external deletion of the retained leader clears its counter and allows replacement, while LWS deletion is handled separately.

panpan0000: Fixed. Spec or Pod updates keep the group frozen; only external deletion of the retained leader clears its counter and allows replacement, while LWS deletion is handled separately.

pkg/controllers/pod_controller.go
panpan0000: I removed `resumeGroupRestartRecovery`. Changing `maxGroupRestarts` no longer resumes a retained group; the operator must delete the retained leader.

panpan0000:
Pushed the updates in 9e750d5. DNS is out of this PR now. I added integration tests for the restart limits, both restart policies, keeping an exhausted group frozen after spec changes, manual recovery, and LWS deletion. I also added webhook tests for the policy and group identity combinations.

The only thing we still need to agree on is resource retention. Keeping the failed group also keeps its resources reserved. https://github.com/kubernetes-sigs/lws/pull/877#discussion_r3923522281


I documented this trade-off and linked #956.

keps/820-distributed-preflight-check/README.md
panpan0000: It's hard to tradeoff, keep the group so that we can `kubectl logs` the pod.. but yes, `crashbackoff` pods still occupied GPU. But in this this initial circuit breaker (preflight test) case,  retaining the group would be acceptable (avoid failing silently  )

ahg-g: I think releasing the resources is important, the way to do this while maintaining access to the failed pods objects and logs is to add finalizers to worker pods too.

yankay:
Thanks for clarifying the recovery semantics and removing `resumeGroupRestartRecovery`. I tested `9e750d5` in Kind and found another path that still replaces Pods after a budget-only edit.

`getPatch()` includes the entire `leaderWorkerTemplate`, including `maxGroupRestarts`. Changing the budget therefore changes the revision hash, which is written into the leader StatefulSet's Pod template labels. This triggers a rolling update independently of the Pod controller's exhausted-group check.

I reproduced two cases:
- For an exhausted, retained group with budget `0`, changing it to `1` or unsetting it replaced the retained leader without explicitly deleting that leader.
- For a healthy group, changing only the budget from `1` to `2` also replaced the leader, with no injected failures.

This means a recovery-policy adjustment can still discard the retained debugging context or unexpectedly roll healthy workloads. Could we exclude `maxGroupRestarts` from rollout identity, keep comparisons compatible with existing revisions, and add lifecycle regression tests? Budget-only edits should preserve Pod identity and restart counts, while actual Pod-template changes should still trigger the expected rollout.

Another finding is: Active surge counters are pruned. pruneScaledDownGroupRestartCounts treats ordinals ≥ spec.replicas as removed. With replicas, maxSurge, and budget all set to 1, the surge ordinal was recreated three times. Could cleanup distinguish active surge from actual scale-down? Using replicas + maxSurge alone could retain stale counters. Tests should cover scale-down during rollout, surge removal, and subsequent scale-up.

keps/820-distributed-preflight-check/README.md
panpan0000: > I think releasing the resources is important, the way to do this while maintaining access to the failed pods objects and logs is to add finalizers to worker pods too.

Aha, @ahg-g   it's a good idea ~ why I didn't realize that .. hahaha 
when Kubernetes >= 1.27, Finalizer will retain the pod(just API resource) in "Failed "state,  meanwhile let kubelet to **release the GPU** , and kubectl logs can still obtain the  "dead pod"'s log with best effort  (unless kubelet rotates /var/logs ).
I will try to add this 


But **one more thing**, 
previously, we can rebuild the group by deleting the leader pod : `kubectl delete pod serving-0`
but it won't work now since group run out of max-restart-budget exhaustion,  the leader has already been deleted and is being held in Terminating by the finalizer, 
so deleting it again cannot act as a new recovery signal. 

We may need a separate signal, maybe we can ONLY rebuild the group at that time by 
```
kubectl annotate pod serving-0   leaderworkerset.sigs.k8s.io/recover=true
```

What do you think  ? @ahg-g

panpan0000: wait........ Finalizer will NOT release GPU resource ...  aka, the Pod pending on Terminating still occupies GPU in scheduler perspective ............

panpan0000: wait, @ahg-g 

Finalizer will stop pod and release GPU, But the log of pod will **lost immediately** 

```
# kubectl -n peter delete pod experiment-sts-0 --wait=false
pod "experiment-sts-0" deleted from peter namespace

# kubectl -n peter get pod experiment-sts-0 -o wide
NAME               READY   STATUS   RESTARTS   AGE   IP               NODE            NOMINATED NODE   READINESS GATES
experiment-sts-0   0/1     Error    0          53s   10.233.100.224   test-hydra-23   <none>           <none>


# kubectl -n peter logs experiment-sts-0
unable to retrieve container logs for containerd://b6b64c71f623ebb6c2b4d6949f1e5e9686a23ae4c8f9a750cfda2911afac122dcloudshell-worker-njkzfmngz8:~# 

```

ahg-g: The logs should be exported to an external logging service like Grafana Loki.

ahg-g: @panpan0000 @yankay what do you think? isn't this a fair assumption that logs should generally be persisted on an external medium? I think generally holding on to the resources is a bad outcome.

panpan0000: @ahg-g , That sounds reasonable as the default. I agree that holding GPUs indefinitely is not a good default, and external logging is common in production. **However**, it is not always available, most small-medium inference cluster don't have full ELK stack, and logging may also be missing for preflight and debugging workflows.
Would you be open to an explicit policy such as `restartBudgetExhaustionPolicy: Terminate | **Retain**`? Terminate would be the default and release the group's resources, while Retain would be opt-in and keep the failed group for debugging, with the resource cost clearly documented.

CC @yankay

ahg-g: yes, my only worry is added complexity, but hopefully it is not too bad.

yankay: HI @panpan0000 

I see Pod retention and troubleshooting as separate concerns. For basic diagnostics, `terminationMessagePolicy: FallbackToLogsOnError` can provide a short log excerpt of up to 2 KiB or 80 lines, whichever is smaller. That may be enough for an initial diagnosis without retaining the entire running group.  Ref https://kubernetes.io/docs/tasks/debug/debug-application/determine-reason-pod-failure/#customizing-the-termination-message

Given how scarce GPU resources are, I'd favor `Terminate` as the default, with `Retain` as an explicit opt-in policy for cases that need more extensive debugging.

BTW, would it be worth supporting a configurable interval or exponential backoff between group restarts to avoid rapid restart loops? This is separate from the restart budget and could be discussed as a follow-up, outside the scope of this PR.

panpan0000: @yankay  thank you for you rely:
1. Yes, `Terminate` is the **default** policy. (`restartBudgetExhaustionPolicy: Terminate | Retain`)
2. IMO,  `terminationMessage` with only 2048 bytes will be shortage for main container crash, you know , python core dump are looooong ..  So I'm afraid introducing `restartBudgetExhaustionPolicy` is better ? what do you think @ahg-g ?
3

ahg-g: I would try and keep it simple for now, as I mentioned ideally logs should be streamed to an external system

panpan0000: Ok, ok , I'll keep it simple:
 after the budget is exhausted, 
 LWS **terminates** the whole group to **release** resources, 
 retains only the terminal Pod **API objects** with **finalizers**,
 and relies on external logging for durable logs.
 
 and adding more user-facing Docs in PR

panpan0000:
Current behaviour:

```
Group failure 
              → recreate up to maxGroupRestarts
              → budget exhausted
                → terminate group and release resources
                → retain terminal Pod API objects
                → Degraded=True
  User    → fix the issue and annotate the retained leader with recover=true
              → recreate the group with a fresh budget
 ```
 
 
 @ahg-g @yankay

**Comment by yankay** [comment]
HI @panpan0000 

Thanks for the careful work on this :)

Could you rebase onto `main` to pick up #1052 and #1053, then rerun Trivy?

I found one issue: active surge groups can have an ordinal `>= spec.replicas`, so the checks at `leaderworkerset_controller.go:252` and `pod_controller.go:606` can clear their restart counts and release their cleanup finalizers, bypassing the budget. Could these checks account for the rollout's effective replica target? I reproduced the finalizer release in envtest while the StatefulSet still targeted the surge group; subsequent recreation is inferred rather than directly tested.

Separately, `recover=true` looks workable to me, and I don't see a concrete issue with it. I'm just wondering whether there might be a better interface for manual recovery. I don't have a specific alternative in mind, so this is a non-blocking design question.

File: test/e2e/e2e_test.go [hunk 14]
```diff
@@ -327,25 +328,103 @@ var _ = ginkgo.Describe("leaderWorkerSet e2e tests", func() {
...
+				return 0, err
+			}
+			var total int32
+			for _, count := range counts {
+				total += count
+			}
+			return total, nil
+		}, timeout, interval).Should(gomega.Equal(int32(1)))
 
-		testing.ExpectValidLeaderStatefulSet(ctx, k8sClient, lws, 7)
```
yankay, line 340: Could we keep the existing test, or ensure its replacement preserves the MaxSurge Service scale-up and cleanup coverage? Thanks :-)

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -489,6 +573,9 @@ func (r *LeaderWorkerSetReconciler) updateConditions(ctx context.Context, lws *l
 		if index < int(*lws.Spec.Replicas) {
 			if ready {
 				readyNonBurstWorkerCount++
```
yankay: `readyNonBurstWorkerCount` seems unused now. Could we remove it?

panpan0000:
@yankay Thanks. Fixed: merged latest main with #1052/#1053, kept restart counts and finalizers for active MaxSurge ordinals, and added regression tests. I kept `recover=true` unchanged.

test/e2e/e2e_test.go
panpan0000: Fixed. I restored the MaxSurge Service scale-up and cleanup test, and kept the restart-budget E2E separately.

pkg/controllers/leaderworkerset_controller.go
panpan0000: Fixed. Removed it.

yankay:
HI @panpan0000 
Thanks for the updates! Recovery and cleanup passed local testing. 
/lgtm

pkg/controllers/pod_controller.go
```diff
@@ -372,6 +435,235 @@ func (r *PodReconciler) handleRestartPolicy(ctx context.Context, pod corev1.Pod,
...
+		patch := client.MergeFrom(pod.DeepCopy())
+		controllerutil.AddFinalizer(pod, leaderworkerset.GroupRestartBudgetCleanupFinalizer)
+		if err := r.Patch(ctx, pod, patch); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *PodReconciler) removeGroupRestartBudgetFinalizers(ctx context.Context, leader *corev1.Pod) error {
```
ahg-g: removeGroupRestartBudgetFinalizer and removeGroupRestartBudgetFinalizers have the exact same signature and the name has a subtle difference (the s at the end), this is a recipe for a future bug

pkg/controllers/pod_controller.go
```diff
@@ -372,6 +435,235 @@ func (r *PodReconciler) handleRestartPolicy(ctx context.Context, pod corev1.Pod,
...
+	for i := range pods {
+		pod := &pods[i]
+		if podutils.LeaderPod(*pod) || !controllerutil.ContainsFinalizer(pod, leaderworkerset.GroupRestartBudgetCleanupFinalizer) {
+			continue
+		}
+		if err := r.removeGroupRestartBudgetFinalizer(ctx, pod); err != nil {
+			return err
+		}
+	}
+	if controllerutil.ContainsFinalizer(leader, leaderworkerset.GroupRestartBudgetCleanupFinalizer) {
```
ahg-g: move this check to inside removeGroupRestartBudgetFinalizer, this will remove a bunch of repeated checks in this PR.

pkg/controllers/pod_controller.go
```diff
@@ -372,6 +435,235 @@ func (r *PodReconciler) handleRestartPolicy(ctx context.Context, pod corev1.Pod,
...
+	return mutateGroupRestartCounts(ctx, r.Client, client.ObjectKeyFromObject(lws), func(_ *leaderworkerset.LeaderWorkerSet, counts map[string]int32) (bool, error) {
+		if _, found := counts[countKey]; !found {
+			return false, nil
+		}
+		delete(counts, countKey)
+		return true, nil
+	})
+}
+
+func (r *PodReconciler) groupLifecycleTeardownRequested(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, leader *corev1.Pod) (bool, error) {
```
ahg-g, line 655: pls document the semantics of this function and when it is expected to be invoked.

pkg/controllers/pod_controller.go
```diff
@@ -118,9 +122,47 @@ func (r *PodReconciler) reconcilePod(ctx context.Context, req podReconcileReques
...
+		if controllerutil.ContainsFinalizer(&pod, leaderworkerset.GroupRestartBudgetCleanupFinalizer) {
+			if podutils.LeaderPod(pod) {
+				return ctrl.Result{}, r.removeGroupRestartBudgetFinalizers(ctx, &pod)
+			}
+			return ctrl.Result{}, r.removeGroupRestartBudgetFinalizer(ctx, &pod)
+		}
+		return ctrl.Result{}, nil
+	}
+	if podutils.LeaderPod(pod) && pod.Annotations[leaderworkerset.GroupRestartBudgetExhaustedAnnotationKey] == "true" {
+		teardown, err := r.groupLifecycleTeardownRequested(ctx, &leaderWorkerSet, &pod)
```
ahg-g, line 153: I am not following what this means, what case are we handling here?

pkg/controllers/pod_controller.go
```diff
@@ -118,9 +122,47 @@ func (r *PodReconciler) reconcilePod(ctx context.Context, req podReconcileReques
...
+	if leaderWorkerSet.DeletionTimestamp != nil {
+		if controllerutil.ContainsFinalizer(&pod, leaderworkerset.GroupRestartBudgetCleanupFinalizer) {
+			if podutils.LeaderPod(pod) {
+				return ctrl.Result{}, r.removeGroupRestartBudgetFinalizers(ctx, &pod)
+			}
+			return ctrl.Result{}, r.removeGroupRestartBudgetFinalizer(ctx, &pod)
+		}
+		return ctrl.Result{}, nil
+	}
+	if podutils.LeaderPod(pod) && pod.Annotations[leaderworkerset.GroupRestartBudgetExhaustedAnnotationKey] == "true" {
```
ahg-g, line 152: pls document the semantics being coded in this if block.

ahg-g: also, this seems like something that handleRestartPolicy handles

ahg-g:
Thanks @panpan0000 , it is unfortunate that we can't handle the GroupIdentity=hash case, but I understand the challenge and we can leave it for a followup.

ahg-g:
@panpan0000 do you think we can get the mostly documentation related comments resolved today/tomorrow to catch the release (planned this week)?

pkg/controllers/pod_controller.go
panpan0000: This handles an LWS that has already been deleted while its Pods still carry our restart-budget cleanup finalizer. Without this path, those Pods could remain stuck terminating. A leader event releases the group; a worker event releases that Pod. I added a comment and kept this separate from restart recovery, since there is no LWS left to update.

pkg/controllers/pod_controller.go
panpan0000: Documented this branch as LWS teardown. It releases budget finalizers without resetting the count or starting recovery; handleRestartPolicy only handles Pod restart or deletion triggers, so this needs a separate path.

pkg/controllers/pod_controller.go
panpan0000: Moved the finalizer check into the single-Pod helper, so callers no longer need to repeat it.

pkg/controllers/pod_controller.go
panpan0000: Documented :  like 
- when the group helper runs, 
- why it releases workers before the leader,
-  how recovery handles the restart count.

pkg/controllers/pod_controller.go
panpan0000: Renamed the single-Pod and group helpers to make their different scopes clear at each call site.

api/leaderworkerset/v1/leaderworkerset_types.go
panpan0000: The final behavior is per replica: an exhausted group stops automatic recovery, while the LWS reports Degraded=True and other replicas can keep running. The operator can set recover=true on the retained leader Pod to reset that group budget and resume recovery.

keps/820-distributed-preflight-check/README.md
panpan0000: The current implementation terminates the whole group and adds cleanup finalizers to leader and worker Pods that can still receive them. Resources are released once those Pods become terminal; the retained objects support inspection, while durable logs require external logging.

api/leaderworkerset/v1/leaderworkerset_types.go
panpan0000: The annotation comment now specifies a JSON map with <revision>/<groupIndex> keys and non-negative restart counts.

panpan0000:
Thanks @ahg-g, I have addressed and resolved the review threads. sorry for late

panpan0000:
god... it's flushed by CLA alters........ T_T

pacoxu:
CI failure is unrelated to this PR, and should be fixed by https://github.com/kubernetes-sigs/lws/pull/1092. And you may need to rebase this after https://github.com/kubernetes-sigs/lws/pull/1092 is merged.

PR #879, opened 2026-06-10
[docs] Add DisaggregatedSet slices KEP

BenjaminBraunDev:
Adds KEP documentation for DisaggregatedSet slices feature

Issue: #846
PR: #873

### DIFF
--- a/keps/846-disaggregatedset-slices/README.md
+++ b/keps/846-disaggregatedset-slices/README.md
@@ -1,0 +1,238 @@
+# KEP-846: DisaggregatedSet Slices
+
+<!--
+This KEP proposes adding a `slices` parameter to DisaggregatedSet that replicates the whole role topology into N independent copies, each rolling out independently.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Scale identical deployments](#story-1-scale-identical-deployments)
+    - [Story 2: One slice per accelerator domain](#story-2-one-slice-per-accelerator-domain)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API](#api)
+  - [Naming and Labels](#naming-and-labels)
+  - [Slices Are Excluded From the Revision Hash](#slices-are-excluded-from-the-revision-hash)
+  - [Per-Slice Reconciliation](#per-slice-reconciliation)
+  - [Services](#services)
+  - [Scale-Down](#scale-down)
+  - [Object Cardinality](#object-cardinality)
+  - [Test Plan](#test-plan)
+    - [Unit tests](#unit-tests)
+    - [Integration tests](#integration-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Alternative 1: Use per-role replicas](#alternative-1-use-per-role-replicas)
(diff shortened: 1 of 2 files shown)


keps/846-disaggregatedset-slices/README.md
```diff
@@ -0,0 +1,238 @@
...
+
+- Slices are identical copies of the spec. The only thing that differs between slice 0 and slice 1 is the slice index in their labels and names.
+- A slice is the durable, outer identity; a revision is ephemeral. Within a slice, normally one revision is live, and two coexist transiently while that slice rolls.
+- Because slices roll independently, different slices can momentarily be on different revisions during a rollout. This is intentional and is what lets a single slice (domain) be updated or recover on its own.
+
+### Risks and Mitigations
+
+**Risk**: Adding the slice segment changes the generated LWS names (`<ds>-<revision>-<role>` becomes `<ds>-<slice>-<revision>-<role>`), so upgrading an existing DisaggregatedSet recreates its LWS objects.
+
+**Mitigation**: DisaggregatedSet has not shipped in a tagged release yet, so there is no deployed name format to preserve. The single-slice case (`slices: 1`) is slice index 0, producing exactly one copy with deterministic naming.
```
Edwinhr716: Good call out with the risk, but the mitigation won't work anymore as we are creating a new release for DisaggregatedSet that won't include this feature, see #884.

Something we can do here which is a bit hacky is when we fetch for existing LWS objects here https://github.com/kubernetes-sigs/lws/blob/main/pkg/controllers/disaggregatedset/disaggregatedset_controller.go#L155, we check if the LWS object has the slice label. If it isn't, it means that it was created before the slice feature was implemented, so we preserve the ds-revision-role naming scheme and do not inject the slice label.

Once an update happens, we change the naming scheme and add the new label.

We did something similar with controller revision https://github.com/kubernetes-sigs/lws/blob/main/keps/238-controller-revision/README.md?plain=1#L172

@ahg-g would changing the name of the LWS objects violate expectations? Or is it fine to change?

Edwinhr716:
Overall lgtm aside from the mitigation

keps/846-disaggregatedset-slices/README.md
hasB4K: The KEP looks good :fire: - the only thing I'm worried about is this. We have already have deployments of DisaggregatedSet in prod.

hasB4K:
Same comment than @Edwinhr716:

LGTM, except the mitigation part

keps/846-disaggregatedset-slices/README.md
BenjaminBraunDev: Okay got it, I'll implement a check for the slice label. So on recreate / rollout it will get updated to the new format, but we need to support DS managing lws roles that have no slice parameter at all.

BenjaminBraunDev: @Edwinhr716 @hasB4K Actually thinking more on this, I think we should have the DS immediately update any legacy lws, the reason being the legacy services' pod selectors only look at the labels for {ds, revision, role} to identify their pods, while the services going forwards also filter on slice label (and label their pods as such). I don't want to have to change the way LWS identifies its pods, so I think the cleanest solution is to immediately force an update upon discovery of an lws without a slice label.

BenjaminBraunDev: Done. Edwin and I discussed this and concluded we can perform a rolling update to move the original legacy slice to the new format with a rolling update upon slice count becoming >1. See latest commit for context.

BenjaminBraunDev:
@hasB4K Addressed the mitigation change, can we checkoff this KEP?

Edwinhr716:
/lgtm

will hold off approval until @hasB4K approves

Edwinhr716:
/lgtm

will hold off approval until @hasB4K approves

hasB4K:
/lgtm

Thanks!

hasB4K (approved):

Edwinhr716:
test infra is stuck. Will bypass since only 1.35 e2e is failing, and it is only updating the docs.

Issue #880, opened 2026-06-12
No safe upgrade path via helm: v0.7.0 → v0.8.0 deletes CRDs


NoobDream2568:
**What happened**:

When upgrading LWS from v0.7.0 (or any earlier version where CRDs lived under `charts/lws/templates/crds/`) to v0.8.0 using `helm upgrade`, the CRD `leaderworkersets.leaderworkerset.x-k8s.io` is **deleted** by Helm, which cascades to the deletion of all existing `LeaderWorkerSet` custom resources in the cluster.

**Root cause**: In commit `2b87638` ("Try to fix Helm so it does not allow deleting CRDs"), the CRD files were moved from `charts/lws/templates/crds/` to `charts/lws/crds/`. This was intended to prevent accidental CRD deletion, since Helm's special `crds/` directory is only installed on `helm install` and never updated or deleted on `helm upgrade`.

However, this relocation introduces a breaking upgrade path for existing users. Helm's upgrade logic computes a manifest diff between the **stored old release manifest** (which includes CRDs as regular template resources) and the **new chart template** (which no longer contains CRDs under `templates/`). Because the CRDs appear in the old manifest but not in the new template, Helm treats them as deleted resources and removes them — along with all custom resources that depend on them.

**What you expected to happen**:

`helm upgrade` from v0.7.0 to v0.8.0 should preserve existing CRDs and upgrade them to the new schema (Yet it's impossible, once the crd is moved).

**How to reproduce it (as minimally and precisely as possible)**:

```bash
# 1. Install v0.7.0
helm pull oci://registry.k8s.io/lws/charts/lws --version 0.7.0
helm install lws lws-0.7.0.tgz --namespace lws-system --create-namespace

# 2. Upgrade to v0.8.0
helm upgrade lws oci://registry.k8s.io/lws/charts/lws --version 0.8.0 --namespace lws-system

# 3. Observe: CRD is gone
kubectl get crd leaderworkersets.leaderworkerset.x-k8s.io
# No resources found (CRD no longer exists)
```

**Anything else we need to know?**:

For users who have already installed v0.7.0 (or earlier) with Helm, the only upgrade path is to `helm uninstall` the release and then `helm install` the new version. In most non-extreme scenarios this is acceptable — the controller pod will be recreated and the cluster returns to a working state, though any existing `LeaderWorkerSet` custom resources will have been lost during the CRD cascade deletion.

[...]

PR #883, opened 2026-06-14
docs: document safe Helm upgrade path for CRDs

anneheartrecord:

When upgrading the LWS Helm release, Helm never updates or deletes the CRDs that ship under the chart's special `crds/` directory — it only installs them on the very first `helm install` (per the [Helm CRD best practices](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/)). As a result CRD schema changes and newly added CRDs (e.g. `DisaggregatedSet`) never reach existing installations through `helm upgrade` alone, and there was no documented, safe in-place upgrade path. Users who worked around this with `helm uninstall` + `helm install` lost all their `LeaderWorkerSet` custom resources to the CRD cascade deletion.

This PR documents the supported upgrade procedure in both the chart README and the site installation docs:

- Apply the CRDs explicitly with `kubectl apply --server-side --force-conflicts` before running `helm upgrade`, covering both the `leaderworkerset` and `disaggregatedset` CRDs (the README previously documented only the `disaggregatedset` CRD).
- Add a new "Upgrade by Helm" section to `site/content/en/docs/installation/_index.md` (the site previously had no Helm upgrade guidance at all).
- Warn explicitly against the `helm uninstall` + `helm install` workaround, which deletes the CRDs and cascades to the deletion of every custom resource in the cluster.

This is documentation-only; the chart structure is unchanged. CRDs intentionally remain under `crds/` (moving them back into `templates/` would re-trigger the v0.7.0 to v0.8.0 deletion cascade for users already on the `crds/`-based release). The literal "bundle CRDs into a pre-upgrade hook Job" suggestion from the issue is impractical because the two CRD files are ~1.3 MB and ~1.5 MB and cannot fit in a ConfigMap (etcd object size limit), so the lowest-risk fix is to document the supported `kubectl apply` reconcile step that maintainers already adopted for the DisaggregatedSet CRD.

Refs #880

```release-note
docs: Document the safe in-place Helm upgrade path for LWS CRDs. Because Helm installs files under the chart's `crds/` directory only on the initial install, run `kubectl apply --server-side --force-conflicts -f <crds>` to reconcile the CRDs before `helm upgrade`; do not use `helm uninstall` + `helm install`, which deletes the CRDs and all LeaderWorkerSet custom resources.
```

### DIFF
--- a/charts/lws/README.md
+++ b/charts/lws/README.md
@@ -94,10 +94,11 @@ kubectl apply --server-side --force-conflicts \
 helm upgrade lws charts/lws --namespace lws-system
 ```
 
-> **Note**: do not run `helm uninstall` followed by `helm install` to "refresh"
-> CRDs. Uninstalling deletes the CRDs, which cascades to the deletion of every
-> `LeaderWorkerSet` and `DisaggregatedSet` custom resource in the cluster. Always
-> upgrade in place and reconcile CRDs with the `kubectl apply` step above.
+> **Note**: neither `helm upgrade` nor `helm uninstall` + `helm install` will
+> update CRD schemas — Helm never modifies CRDs placed in the `crds/` directory
+> after the initial install, and does not delete them on `helm uninstall` either.
+> Always reconcile CRD schemas explicitly with the `kubectl apply` step above
+> before upgrading.
 
 ### Configuration
 
--- a/charts/lws/README.md
+++ b/charts/lws/README.md
@@ -8,6 +8,7 @@
     - [Installing the chart](#installing-the-chart)
         - [Install chart using Helm v3.0+](#install-chart-using-helm-v30)
         - [Verify that controller pods are running properly.](#verify-that-controller-pods-are-running-properly)
+    - [Upgrading the chart](#upgrading-the-chart)
     - [Configuration](#configuration)
 <!-- /toc -->
 
@@ -68,17 +69,35 @@ controller are always installed with the chart. To also install the
 editor/viewer/admin ClusterRoles and validating webhook, set
 `enableDisaggregatedSet` to `true`.
 
-Helm installs CRDs during `helm install`, but does not install newly added CRDs
-during `helm upgrade`. Before upgrading an existing LWS Helm release to a chart
-version that introduces DisaggregatedSet, apply the new CRD from the repository
-root:
+### Upgrading the chart
+
+The chart ships its CRDs under the special `crds/` directory. Helm installs the
+files in that directory only during the initial `helm install`; it never updates
+or deletes them on `helm upgrade` (see the
+[Helm documentation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/)).
+
+This has two consequences when upgrading an existing release:
+
+- CRD schema changes (new fields, printer columns, conversion settings, etc.) do
+  **not** reach the cluster through `helm upgrade`.
+- A newly added CRD (for example `DisaggregatedSet`) is **not** installed by
+  `helm upgrade`.
+
+Apply the CRDs explicitly before upgrading so the cluster always has the schema
+the new chart version expects. `--server-side` lets the apply coexist with the
+ownership Helm records for the resource:
 
 ```bash
-kubectl apply --server-side \
+kubectl apply --server-side --force-conflicts \
+  -f charts/lws/crds/leaderworkerset.x-k8s.io_leaderworkersets.yaml \
   -f charts/lws/crds/disaggregatedset.x-k8s.io_disaggregatedsets.yaml
-helm upgrade lws charts/lws --namespace lws-system \
-  --set enableDisaggregatedSet=true
+helm upgrade lws charts/lws --namespace lws-system
 ```
+
+> **Note**: do not run `helm uninstall` followed by `helm install` to "refresh"
+> CRDs. Uninstalling deletes the CRDs, which cascades to the deletion of every
+> `LeaderWorkerSet` and `DisaggregatedSet` custom resource in the cluster. Always
+> upgrade in place and reconcile CRDs with the `kubectl apply` step above.
 
 ### Configuration
 
--- a/charts/lws/README.md
+++ b/charts/lws/README.md
@@ -100,6 +100,27 @@ helm upgrade lws charts/lws --namespace lws-system
 > Always reconcile CRD schemas explicitly with the `kubectl apply` step above
 > before upgrading.
 
+#### Upgrading from v0.7.0 or earlier
+
+Chart versions up to v0.7.0 rendered the `LeaderWorkerSet` CRD from
+`templates/crds/`, so the CRD is part of the Helm release manifest. Starting
+with v0.8.0 the CRD ships from the `crds/` directory and is no longer part of
+the release. Without preparation, the first `helm upgrade` across that boundary
+treats the CRD as removed from the release and deletes it — cascading to the
+deletion of every `LeaderWorkerSet` in the cluster (see
+[#880](https://github.com/kubernetes-sigs/lws/issues/880)).
+
+Before the first upgrade from v0.7.0 or earlier, run this one-time step so Helm
+keeps the CRD when it leaves the release:
+
+```bash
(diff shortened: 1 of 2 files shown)


NoobDream2568:
Unfortunately, `kubectl apply --server-side --force-conflicts` does not help — Helm still removes the CRD during `helm upgrade`.

Helm's deletion decision is based on the manifest diff stored in the release Secret, not on the live cluster state. So even if the CRD is updated in the cluster before upgrade, Helm still sees it in the old manifest and deletes it because it's absent from the new chart's templates.

As for upgrading CRDs going forward, a proven practice is to build an `lws-crd-upgrade` image (containing the CRD files and `kubectl`) and run it as a Job during the `pre-install` and `pre-upgrade` Helm hook phases. This way the CRDs are applied via `kubectl create`/`kubectl replace` every time, and there is no need to place CRDs explicitly under the special `crds/` directory — IMO removing them from `crds/` would not cause CRD uninstallation.
#880

anneheartrecord:
Thanks @NoobDream2568 — your description is accurate for CRDs placed in `templates/`, where Helm tracks them in the release Secret and can delete them on upgrade/uninstall. However, lws places both its CRDs exclusively in `charts/lws/crds/`:

```
charts/lws/crds/
├── disaggregatedset.x-k8s.io_disaggregatedsets.yaml
└── leaderworkerset.x-k8s.io_leaderworkersets.yaml
```

The [Helm docs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/) explicitly distinguish the two approaches:

> CRDs in the `crds/` directory are **not** part of the release. Helm installs them on `helm install` only and **never updates or deletes** them on `helm upgrade` or `helm uninstall`.

Because `crds/`-directory CRDs are outside the release manifest, Helm's manifest-diff logic never touches them. The `kubectl apply --server-side --force-conflicts` step is not there to "prevent Helm from deleting the CRD" — it is there because `helm upgrade` will _never_ update CRD schemas in `crds/` at all, so an explicit apply before upgrading is required to ensure the cluster has the updated schema the new chart version expects.

You are right that our Note was inaccurate: it claimed `helm uninstall` deletes the CRDs, which is wrong for `crds/`-directory CRDs. I pushed a correction in 2c8a106 — the Note now correctly states that Helm does **not** delete `crds/`-directory CRDs on uninstall.

yankay:
Thanks for the update.

Following up on https://github.com/kubernetes-sigs/lws/issues/880#issuecomment-4715134850, I think this documents the safe path for releases that already use `crds/`, but may not fully cover the v0.7.0-or-earlier migration path from `templates/crds/`.

@Edwinhr716 @ahg-g Would you mind sharing your thoughts on #880

anneheartrecord:
Good point — this is exactly the original #880 scenario, and the current doc doesn't cover it.

For releases ≤ v0.7.0 the CRDs lived under `charts/lws/templates/crds/`, so Helm tracked them in the release. When v0.8.0 moves them to the unmanaged `crds/` directory, `helm upgrade` sees the CRD removed from the rendered manifest and deletes it, which cascades to every `LeaderWorkerSet`.

The safe one-time step before that first upgrade is to detach the CRD from Helm's bookkeeping so the move doesn't delete it:

```bash
kubectl annotate crd leaderworkersets.leaderworkerset.x-k8s.io \
  helm.sh/resource-policy=keep --overwrite
```

After that, the `kubectl apply --server-side` + `helm upgrade` path in this PR is non-destructive. I'm happy to add an "Upgrading from v0.7.0 or earlier" subsection covering this.

If you'd rather fix it at the chart level instead — e.g. ship the CRDs carrying `helm.sh/resource-policy: keep` so the template→`crds/` move never deletes them — I can go that way too. Let me know which you'd prefer.

anneheartrecord:
Bumping this one. To keep it moving, two ways forward: I can fold in a short "Upgrading from v0.7.0" subsection so it also covers the pre-`crds/` migration @yankay raised on #880, or keep this PR scoped to the already-on-`crds/` upgrade path and handle v0.7.0 separately. Either is fine by me — whichever you and @Edwinhr716 prefer. CI is green.

anneheartrecord:
To keep this from stalling on the #880 discussion: I've kept this PR scoped to the already-on-`crds/` upgrade path, which is the ready and non-controversial part. The v0.7.0-or-earlier migration (detaching the CRD from Helm's bookkeeping before the `templates/crds/` → `crds/` move) is a separate concern; I'll send it as a follow-up once @Edwinhr716 / @ahg-g weigh in on #880. This one should be safe to merge as-is, and CI is green.

Edwinhr716:
I shared my thoughts in #880. @yankay will delegate to you to approve

yankay:
HI @anneheartrecord 

Thanks a lot for continuing to move this PR forward and for improving the Helm CRD upgrade documentation. I think this is very helpful for users.

My understanding is that #886 provides a possible full fix and is worth considering, though that approach is relatively heavy because it introduces a new image. As @Edwinhr716  mentioned in #880, I also think the historical v0.7.0-or-earlier to v0.8.0+ migration path should eventually be fixed properly.

For this PR, I suggest changing `Fixes #880` to `Refs #880` in PR description, so we don’t close an issue that is not fully resolved yet. At the same time, I agree that improving the documentation before a complete fix lands is still valuable. After the issue reference is adjusted, I’d be happy to help move this PR forward.

I have a similar docs-oriented draft in https://github.com/yankay/lws/pull/14, and some of its may be useful to use. That said, I would prefer to keep this PR moving first. Overall, I’m in favor of moving forward with this after the issue reference is adjusted.

anneheartrecord:
Thanks @yankay, @Edwinhr716. Done — the description now reads `Refs #880` instead of `Fixes #880`, so this won't close the issue while the full v0.7.0→v0.8.0+ migration is still pending (tracked in #880 / #886).

Scope here stays docs-only: the supported `kubectl apply --server-side --force-conflicts` reconcile step before `helm upgrade`, plus the warning against `helm uninstall` + `helm install`. Happy to fold in anything useful from your draft in yankay/lws#14 if you point out specific bits; otherwise this should be ready for review.

anneheartrecord:
Thanks @yankay! Good news — the PR description already uses `Refs #880` (not `Fixes`), so merging this won't auto-close the issue; GitHub's linked "closing issues" list is empty for this PR. So the reference is already in the state you asked for.

Agreed on keeping this scoped to documenting the safe upgrade path for the `crds/`-based releases, and leaving the full v0.7.0-or-earlier → v0.8.0+ migration to a proper fix (whether #886's approach or another). Happy to fold in anything useful from your draft in yankay/lws#14 if you'd like. Ready to move forward whenever you are — thanks for helping push it along!

yankay:
Hi @anneheartrecord,

I think this PR should also document the v0.7.0-or-earlier migration path. Otherwise users moving from templates/crds/ to crds/ may follow this guide but still hit the CRD deletion problem from [kubernetes-sigs/lws#880](https://github.com/kubernetes-sigs/lws/issues/880).

Could you fold in a short “Upgrading from v0.7.0 or earlier” section, including the one-time helm.sh/resource-policy=keep step before the first upgrade? The install-guide warning and troubleshooting section from [yankay/lws#14](https://github.com/yankay/lws/pull/14) may be useful references.

I think this PR is valuable, and I’d be happy to move it forward once that case is covered too.

anneheartrecord:
Done @yankay — added an "Upgrading from v0.7.0 or earlier" section to both the chart README and the site installation guide (df9b7eb). It explains the `templates/crds/` → `crds/` move and the one-time step before the first upgrade across that boundary:

```bash
kubectl annotate crd leaderworkersets.leaderworkerset.x-k8s.io \
  helm.sh/resource-policy=keep --overwrite
```

so Helm keeps the CRD when it leaves the release instead of deleting it (and every LeaderWorkerSet with it). After that, the regular flow applies: `kubectl apply --server-side` the CRDs, then `helm upgrade`. PTAL.

yankay:
HI @anneheartrecord 

Thanks for adding the v0.7.0-or-earlier migration path. This covers the templates/crds -> crds case I was concerned about.

/approve

PR #886, opened 2026-06-17
fix(helm): add pre-install/pre-upgrade hook for CRD management and prevent helm upgrade deleting v0.7.0 CRDs

### LINKED ISSUE #880: No safe upgrade path via helm: v0.7.0 → v0.8.0 deletes CRDs

NoobDream2568:
#### What type of PR is this?

/kind bug

When upgrading from v0.7.0 (CRDs in `templates/crds/`) to v0.8.0 (CRDs in `crds/`), Helm's manifest-diff logic deletes the LeaderWorkerSet CRD because it was tracked in the old release manifest but no longer appears in the new rendered templates. This causes cascade deletion of all LeaderWorkerSet custom resources in the cluster.

This PR adds a pre-install/pre-upgrade hook Job that manages CRDs independently of the Helm release manifest. The hook runs three phases:

1. **Apply CRDs** — `kubectl create` (new) or `kubectl replace` (existing) with the updated schema
2. **Patch conversion webhooks** — restore the `spec.conversion` webhook configuration
3. **Annotate with `helm.sh/resource-policy: keep`** — prevent Helm from deleting CRDs that were previously tracked under `templates/`

Phase 3 must run after Phase 1 because `kubectl replace` overwrites all metadata including annotations, which would strip the `resource-policy` annotation if applied earlier. This approach follows the precedent set by [secrets-store-csi-driver#656](https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/656).

Fixed #880

Verified end-to-end on a kind cluster (v0.7.0 → v0.8.0):
- LeaderWorkerSet CRD: preserved, schema updated (`volumeClaimTemplates` field present)
- DisaggregatedSet CRD: newly created
- `helm.sh/resource-policy: keep` annotation: correctly set on both CRDs
- All existing custom resources: intact

The `crds/` directory can optionally be removed since the hook fully handles the CRD lifecycle for both install and upgrade. Keeping it is harmless (Helm installs on first `helm install` only, never updates or deletes).

```release-note
Add a pre-install/pre-upgrade hook Job to manage CRDs, preventing CRD cascade deletion when upgrading from v0.7.0 (CRDs tracked in `templates/`) to v0.8.0+ (CRDs in `crds/`). No manual `kubectl apply` step is required.
```

### DIFF
--- a/charts/lws/templates/upgrade/crd-upgrade.yaml
+++ b/charts/lws/templates/upgrade/crd-upgrade.yaml
@@ -1,0 +1,81 @@
+{{ if .Values.crdUpgrade.enabled -}}
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: lws-crds-upgrade-{{ replace "." "" .Chart.AppVersion }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-4"
+    "helm.sh/hook-delete-policy": before-hook-creation
+spec:
+  {{- if .Values.crdUpgrade.ttlSecondsAfterFinished }}
+  ttlSecondsAfterFinished: {{ .Values.crdUpgrade.ttlSecondsAfterFinished }}
+  {{- end }}
+  template:
+    spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.crdUpgrade.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.crdUpgrade.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: lws-crds-upgrade
+      containers:
+        - name: lws-crds-upgrade
+          image: "{{ .Values.crdUpgrade.image.repository }}:{{ .Values.crdUpgrade.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.crdUpgrade.image.pullPolicy | default "IfNotPresent" }}
+          env:
+            - name: WEBHOOK_NAMESPACE
+              value: {{ .Release.Namespace | quote }}
+            - name: WEBHOOK_SERVICE
+              value: lws-webhook-service
+      restartPolicy: OnFailure
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: lws-crds-upgrade
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded,hook-failed,before-hook-creation
+rules:
+  - apiGroups: ["apiextensions.k8s.io"]
+    resources: ["customresourcedefinitions"]
+    verbs: ["create", "get", "update", "patch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: lws-crds-upgrade
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded,hook-failed,before-hook-creation
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: lws-crds-upgrade
+subjects:
+  - kind: ServiceAccount
+    name: lws-crds-upgrade
+    namespace: {{ .Release.Namespace }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: lws-crds-upgrade
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded,hook-failed,before-hook-creation
+{{- end }}
--- a/tools/crd-upgrade/Dockerfile
+++ b/tools/crd-upgrade/Dockerfile
@@ -1,0 +1,24 @@
+# CRD Upgrader image for lws
+# This image contains all CRD files and kubectl to upgrade CRDs during Helm upgrade
+FROM alpine:3.20
+
+# Copy CRD files and upgrade script
+COPY ./config/crd/bases /lws/crds
+COPY ./tools/crd-upgrade/ensure-crds-up-to-date.sh /lws/ensure-crds-up-to-date.sh
+
+# Install necessary tools
+RUN apk add --update --no-cache bash curl && \
+    chmod +x /lws/ensure-crds-up-to-date.sh
+
+# Install kubectl
+ARG KUBECTL_VERSION=v1.31.0
+ARG TARGETARCH
+RUN ARCH="${TARGETARCH:-amd64}" && \
+    curl -LO "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/${ARCH}/kubectl" && \
+    curl -LO "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/${ARCH}/kubectl.sha256" && \
+    echo "$(cat kubectl.sha256)  kubectl" | sha256sum -c && \
+    mv kubectl /usr/local/bin/kubectl && \
+    chmod +x /usr/local/bin/kubectl && \
+    rm kubectl.sha256
+
+ENTRYPOINT ["bash", "/lws/ensure-crds-up-to-date.sh"]
--- a/tools/crd-upgrade/ensure-crds-up-to-date.sh
+++ b/tools/crd-upgrade/ensure-crds-up-to-date.sh
@@ -1,0 +1,90 @@
+#!/bin/bash
+set -e
+
+# CRDs that require a conversion webhook clientConfig.
+# After install/replace, spec.conversion is patched with the correct service
+# name and namespace so Kubernetes routes conversion requests to the webhook.
+CONVERSION_CRDS=(
+  "leaderworkersets.leaderworkerset.x-k8s.io"
+)
+
+# Namespace and service name for the conversion webhook.
+# Override via environment variables when invoking this script.
+WEBHOOK_NAMESPACE="${WEBHOOK_NAMESPACE:-lws-system}"
+WEBHOOK_SERVICE="${WEBHOOK_SERVICE:-lws-webhook-service}"
+
+# Collect CRD names from the bundled YAML files.
+CRD_NAMES=()
+while IFS= read -r -d '' crdfile; do
+  crd_name=$(grep -m1 '^  name:' "$crdfile" | awk '{print $2}')
+  [[ -n "$crd_name" ]] && CRD_NAMES+=("$crd_name")
+done < <(find /lws/crds -maxdepth 1 -name '*.yaml' -print0)
+
+###############################################################################
+# Phase 1 — Apply CRDs (create or replace)
+###############################################################################
+echo "=== Phase 1: Applying CRDs ==="
+
+for crd_name in "${CRD_NAMES[@]}"; do
+  # Find the matching file.
+  crdfile=$(find /lws/crds -maxdepth 1 -name '*.yaml' | while read -r f; do
+    grep -q "name: $crd_name" "$f" && echo "$f"
+  done | head -1)
+  crdshort=${crdfile##*/}
+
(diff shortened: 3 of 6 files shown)


charts/lws/templates/upgrade/crd-upgrade.yaml
```diff
@@ -0,0 +1,81 @@
...
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-4"
+    "helm.sh/hook-delete-policy": before-hook-creation
+spec:
+  {{- if .Values.crdUpgrade.ttlSecondsAfterFinished }}
+  ttlSecondsAfterFinished: {{ .Values.crdUpgrade.ttlSecondsAfterFinished }}
+  {{- end }}
+  template:
+    spec:
```
yankay, line 17: HI @NoobDream2568 
Thanks for adding this hook job. Could we also add pod/container securityContext here? The manager already runs with non-root and restricted settings, but this hook Pod currently has no securityContext. In namespaces enforcing restricted Pod Security Admission, the Pod may be rejected and block the Helm upgrade.

tools/crd-upgrade/ensure-crds-up-to-date.sh
```diff
@@ -0,0 +1,90 @@
...
+for crd_name in "${CRD_NAMES[@]}"; do
+  # Find the matching file.
+  crdfile=$(find /lws/crds -maxdepth 1 -name '*.yaml' | while read -r f; do
+    grep -q "name: $crd_name" "$f" && echo "$f"
+  done | head -1)
+  crdshort=${crdfile##*/}
+
+  if kubectl get crd "$crd_name" >/dev/null 2>&1; then
+    echo "$crdshort found, replacing CRD..."
+    kubectl replace -f "$crdfile"
```
yankay, line 37: Could we avoid `kubectl replace` here? In my local test the upgrade succeeded, but the CRD labels from `charts/lws/crds` were removed after the hook ran. Using the final chart CRDs with `kubectl apply --server-side` would preserve the chart output better and reduce the chance of overwriting metadata or fields managed elsewhere.

charts/lws/templates/upgrade/crd-upgrade.yaml
```diff
@@ -0,0 +1,81 @@
...
+            - name: WEBHOOK_NAMESPACE
+              value: {{ .Release.Namespace | quote }}
+            - name: WEBHOOK_SERVICE
+              value: lws-webhook-service
+      restartPolicy: OnFailure
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: lws-crds-upgrade
```
yankay, line 45: Could we name these hook ServiceAccount/RBAC resources with the chart fullname helper? Most RBAC resources in this chart use `{{ include "lws.fullname" . }}`, while these hook resources are hard-coded as `lws-crds-upgrade`. Using the same naming pattern would reduce collision risk and keep the chart consistent.

tools/crd-upgrade/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# CRD Upgrader image for lws
```
yankay, line 1: Maybe we can keep the automated upgrade experience here but avoid a second LWS-specific image. kube-prometheus-stack uses a standard kubectl image for its CRD upgrade job when that job is enabled, and another option would be reusing the existing LWS manager image with a small upgrade subcommand. That would make the release path much simpler.

tools/crd-upgrade/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# This image contains all CRD files and kubectl to upgrade CRDs during Helm upgrade
+FROM alpine:3.20
+
+# Copy CRD files and upgrade script
+COPY ./config/crd/bases /lws/crds
```
yankay, line 6: Nice idea to bundle the CRDs with the hook image. One concern is that this copies `config/crd/bases`, while the chart ships the final CRDs from `charts/lws/crds`. Could we use the same final manifests that the chart publishes, so the hook applies exactly what users would expect from the chart?

yankay (commented):
HI @NoobDream2568 

Thanks again for working on this. I tested the hook flow locally, and once the upgrader image is available, the v0.7.0 -> v0.8.0 upgrade path does preserve the CRD and existing LeaderWorkerSets. So I think the core idea is valid.

My main concern is the new LWS-specific upgrader image. Adding a second required release image feels a bit heavy for this chart, because it needs to be built, promoted to registry.k8s.io, documented, and verified for every release. Could we keep the same automated upgrade experience but avoid a new LWS image, for example by using a standard kubectl image with the chart CRDs, or by reusing the existing manager image with a small upgrade script ?

yankay:
As a follow-up, it may be worth adding focused upgrade e2e coverage for this path. Kueue has a similar upgrade suite that installs a released version, creates pre-existing resources, upgrades to the current manifests, and verifies that old resources and webhook/CRD behavior still work. For LWS, a smaller version could install the released v0.7.0 chart, create a small LeaderWorkerSet, upgrade to the current chart, and verify that the CRD plus the existing LeaderWorkerSet are preserved.

NoobDream2568:
Thanks for the review. I understand the concern about adding a second image.

I looked at alternatives to avoid a new image:

1. **Reusing the manager image**: Not feasible — the LWS manager uses `gcr.io/distroless/static:nonroot` which has no shell and no kubectl. Adding them would break the distroless principle and bloat the image by ~50MB.

2. **ConfigMap + compressed CRDs** (similar to kube-prometheus-stack's `crds.bz2` approach): This would avoid a new image but adds significant complexity:
   - A Makefile target to compress CRDs into bz2
   - An init container (e.g. busybox) to decompress at runtime
   - A main container with kubectl to apply the decompressed files
   - More moving parts to debug and maintain

The dedicated `lws-upgrade-crd` image keeps the hook Job to a single container with a straightforward script: apply CRDs, patch webhooks, annotate resource-policy. The image is small (~80MB, Alpine + kubectl + CRD files), and the build/push is already wired into the Makefile (`make docker-build-crd-upgrader` / `make docker-push-crd-upgrader`).

So probably the dedicated image is the cleanest solution for the upgrade experience we want to provide, in my opinion.

BenAyedMedAla (commented):
Went through the script — the 3-phase ordering (replace → patch conversion webhook → annotate `resource-policy: keep`, in that order because `replace` strips annotations) is correct and well-reasoned. RBAC is minimal and matches what the script actually needs.

On the image question: the distroless-manager-image counter-argument holds up — no shell/kubectl in `gcr.io/distroless/static:nonroot` without compromising the whole point of using distroless. The ConfigMap+bz2 alternative is a fair amount of extra moving parts for not much gain. Dedicated image seems like the more defensible tradeoff here.

One open item from earlier in the thread: the focused upgrade-e2e coverage for this exact v0.7.0→v0.8.0 path is still missing (confirmed — #1003 explicitly scoped it out). Might be worth adding as a condition for merging, since it's the only way to keep this path from silently regressing again.

ahg-g:
@yankay is this ready to merge?

yankay:
Thanks for following up. I still have some concerns before considering this ready to merge.

The data-loss risk only affects the one-time transition from v0.7.0 or earlier to v0.8.0+. Subsequent upgrades do not have the same problem, and #883 now documents the required annotation and CRD apply steps.

This PR turns that migration into a permanent image and runtime dependency. The image is not yet integrated into `cloudbuild.yaml`, the release checklist, or the promotion workflow, and it would require ongoing ownership, security updates, and release verification.

A lighter alternative could be a migration-only `pre-upgrade` hook using a configurable `registry.k8s.io/kubectl` image. It would only add `helm.sh/resource-policy=keep` with `get`/`patch` RBAC, leaving CRD updates to the documented server-side apply step.

However, I also have reservations about this alternative: it would still run on every upgrade and require an additional image pull. Since users already need to follow the CRD upgrade documentation, keeping this one-time migration documented may be simpler unless fully automated upgrades are an explicit requirement.

BenAyedMedAla:
Dug into the currently-failing required check (`pull-lws-test-e2e-upgrade-helm-main`, [build 2095414119808110592](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_lws/886/pull-lws-test-e2e-upgrade-helm-main/2095414119808110592)) to see whether it's a flake before commenting.

It isn't — the pre-upgrade hook Job pod events show:

```
Failed to pull image "registry.k8s.io/lws/lws-upgrade-crd:v0.10.0": ... not found
Error: ImagePullBackOff
```

`Error: UPGRADE FAILED: pre-upgrade hooks failed: 1 error occurred: * timed out waiting for the condition` is Helm giving up waiting on that pod. So this confirms @yankay's concern above isn't just a future maintenance worry — `registry.k8s.io/lws/lws-upgrade-crd` doesn't exist yet because nothing publishes it there (no `cloudbuild.yaml` / k8s-staging image-promoter entry for it), so `crdUpgrade.enabled: true` by default means every real `helm install`/`helm upgrade` would hit this same `ImagePullBackOff`, not just this CI job. That's a hard blocker on its own, independent of the image-vs-lighter-hook tradeoff being discussed.

Separate, smaller finding while reading `tools/crd-upgrade/Dockerfile`: `ARG KUBECTL_VERSION=v1.31.0` is hardcoded. LWS's own required E2E matrix now covers Kubernetes 1.34-1.37 (kubernetes-sigs/lws#1012, kubernetes/test-infra#37850), so pinning this migration tool's kubectl to 1.31 puts it several minors outside the version-skew policy's supported client range relative to the clusters this hook would actually run against.

I'd lean toward @yankay's lighter alternative (a `pre-upgrade`-only hook using the existing `registry.k8s.io/kubectl` image, or documenting the manual step as #883 already does) over adding a new image + promotion pipeline to maintain — but wanted to get the concrete failure evidence in the thread either way.

yankay:
Setting aside the CRD deletion risk when upgrading from v0.7.0, I’m still unsure whether automating future CRD updates with a hook would justify the added maintenance complexity, or whether clearly documenting the upgrade steps would be sufficient.  :-)
I don’t have a clearer recommendation or a better alternative to offer yet.

ahg-g:
I would lean towards clear documentation than automation in this case

ahg-g:
I suggest we close this and add clear documentation on how to upgrade

PR #903, opened 2026-06-25
Helm: fix Prometheus ServiceMonitor for 'enablePrometheus: true'

EladDolev:
#### What type of PR is this?

/kind bug

this PR fixes Prometheus scraping which is not working at all because of a labels mismatch.

```release-note
NONE
```

### DIFF
--- a/charts/lws/templates/manager/service.yaml
+++ b/charts/lws/templates/manager/service.yaml
@@ -16,4 +16,5 @@ spec:
       targetPort: 8443
   selector:
     control-plane: controller-manager
+    {{- include "lws.selectorLabels" . | nindent 4 }}
 {{- end }}
--- a/charts/lws/templates/prometheus/monitor.yaml
+++ b/charts/lws/templates/prometheus/monitor.yaml
@@ -34,5 +34,5 @@ spec:
         {{- end }}
   selector:
     matchLabels:
-      control-plane: controller-manager
+      {{- include "lws.labels" . | nindent 6 }}
 {{- end }}


charts/lws/templates/manager/service.yaml
EladDolev, line 1: if this service only relevant when Prometheus is enabled and has the suffix `-metrics-service` then it probably belongs under `/prometheus` and not under `/manager`

happy to move it 😸

charts/lws/templates/manager/service.yaml
```diff
@@ -16,4 +16,5 @@ spec:
       targetPort: 8443
   selector:
     control-plane: controller-manager
+    {{- include "lws.selectorLabels" . | nindent 4 }}
```
EladDolev, line 19: that prevents a potential bug if there are other pods unrelated to this chart that happen to have `control-plane: controller-manager` label

charts/lws/templates/manager/service.yaml
EladDolev, line 6: that's actually wrong IMHO and will prevent having multiple releases but I left it untouched and only used `"lws.labels"` for the ServiceMonitor `selector.matchLabels` which won't also support multiple releases

EladDolev:
hello @Edwinhr716 👋 

is there anything I can do on my side to help getting this PR merged ?

Edwinhr716:
test 1.35 keeps getting stuck, I'll overpass that test since it is only a helm chart update

PR #907, opened 2026-07-01
DisaggregatedSet: pipeline rolling updates with fractional lockstep

hasB4K:
#### What type of PR is this?

/kind feature
/kind cleanup

The existing DisaggregatedSet rollout waits for all replicas in the current step to become Ready before it can continue. One slow-starting pod can therefore stop progress for every role. Larger `MaxSurge` and `MaxUnavailable` values also do not consistently translate into faster progress.

This PR pipelines rolling updates with **fractional lockstep**:

- The planner advances the old and new sides independently on role-proportional fractional steps.
- Roles may be on different steps, but their progress stays inside one moving coordination window.
- The executor uses both LWS Spec and Ready counts to enforce per-role surge, availability, and pending-readiness limits.
- Old revisions are drained newest first and are never intentionally left without one of their roles. If no safe partial drain, coordinated retirement, or replacement growth is possible, the rollout waits and emits an event.
- The rollout completes only after old Specs reach zero and every target role is Ready.

Interrupted rollouts are handled explicitly. The existing `initial-replicas` annotation tracks the intended size of the current revision and freezes when that revision becomes old. During an A → B → C rollout, the old-side baseline for each role is the largest intended value from A or B; their current Specs are still summed when checking physical capacity. Missing target roles are recreated at zero replicas before planning resumes.

Rollout state is reconstructed after a controller restart from the existing LWS objects, using the `disaggregatedset.x-k8s.io/revision` label and the `initial-replicas` annotation.

The planner and executor form one behavioral change and need to land together. The history is organized into four commits so they can still be reviewed in layers.

#### Review guide

1. `32ba240` — fractional-lockstep planner and planner unit tests.
2. `746fcec` — readiness bounds, coordinated drain, interrupted-rollout state, target-role recovery, and controller integration.
3. `e59a565` — invariant-based rollout and slow-readiness e2e coverage.
4. `3b5a0a0` — KEP and reference documentation.

The main implementation flow is in `pkg/controllers/disaggregatedset/executor.go`. The readiness, surge, availability, and coordination-window checks are isolated in `pkg/controllers/disaggregatedset/rollout_bounds.go`.

#### Testing

- `make test` — 618 tests passed.
- `go test -run '^$' ./test/e2e/disaggregatedset` — e2e suite compiles.

```release-note
DisaggregatedSet rolling updates now pipeline slow-starting replicas through a role-proportional fractional window while preserving per-role surge and availability limits.
```


yankay:
HI @hasB4K 


Thanks a lot for working on this. I really appreciate the detailed write-up  :)

I agree with the rollout-gate change: moving from `pending == 0` to tolerating `pending <= maxSurge + maxUnavailable` looks like the right fix for slow pods blocking the rollout.

**Update after our discussion:** I reread [KEP 766](https://github.com/kubernetes-sigs/lws/tree/main/keps/766-DisaggregatedSet), which describes coordinated rolling updates that keep roles in lockstep while respecting per-role surge constraints. The proportional P:D behavior therefore makes sense. I also agree that the projection remains within the `maxSurge`/`maxUnavailable` contract, since these fields define upper limits that do not need to be fully consumed.

I withdraw my earlier concerns about the semantics and splitting the changes. I have no objection to the overall direction and will continue reviewing the implementation details. A follow-up documentation update may still be helpful, but it does not need to block this PR.

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -337,79 +412,211 @@ func (executor *RollingUpdateExecutor) scaleDownOld(
...
-	for _, v := range s {
-		if v > 0 {
-			return false
+	// Deadlock fallback: aliveness may have skipped every revision (no drain
+	// applied) even though planner still has budget to spend. This happens
+	// when no revision can safely partial-drain AND no revision can be
+	// fully retired within budget. Force-retire the newest non-empty revision
+	// to unstick; the aliveness invariant is briefly broken, but progress
+	// resumes next reconcile.
+	if !drainedAny {
```
BenjaminBraunDev: Because this fallback can't tell why nothing drained, the termination wait guard above exits with drainedAny false, so while a retired revision's pods are still in preStop, this force-retires the next healthy revision to 0 with no budget check. Repro in a unit test: A at 4P/4D, budget 1 per role, B retired with pods terminating, result is A wiped to 0/0.

Suggest not firing the fallback when the no-drain reason was the guard or an aliveness skip, and capping it at the drain budget either way. One thing to think through: if the terminating pods stay stuck for many reconciles, the guard blocks all progress indefinitely, so that path probably needs a bounded wait perhapse?

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -141,20 +141,21 @@ func (executor *RollingUpdateExecutor) ReconcileRollingUpdate(
...
 		return ctrl.Result{RequeueAfter: time.Second}, nil
 	}
 
-	initialOld, currentOld, currentNew, targetNew := buildPlannerState(disaggregatedSet, allRoleNames, specRoleSet, oldRevisions, newRevision)
-	config := extractRollingUpdateConfig(disaggregatedSet, allRoleNames)
+	initialOld, currentOld, currentNew, targetNew := buildPlannerStateMaps(disaggregatedSet, allRoleNames, specRoleSet, oldRevisions, newRevision)
 
-	nextStep := ComputeNextStep(initialOld, currentOld, currentNew, targetNew, config)
+	nextStep := ComputeNextStep(allRoleNames, initialOld, currentOld, currentNew, targetNew, config)
 	if nextStep == nil {
```
BenjaminBraunDev: nil here means both "complete" and "nothing changed this tick", and the second is reachable mid-rollout:

```go
ComputeAllSteps([]string{"p", "d"},
    map[string]int{"p": 1, "d": 4}, map[string]int{"p": 1, "d": 4},
    map[string]RollingUpdateConfig{"p": {0, 1}, "d": {0, 1}})
```

Causes it to get stuck at old 1P/2D, new 0P/1D and never completes (aliveness cap zeroes the drains while the ceiling blocks the adds). This branch then emits RollingUpdateCompleted with no requeue, and the deadlock fallback can't help because we return before scaleDownOld. Suggest re-checking isComplete before emitting, and requeueing when the nil was just a blocked tick.

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -236,54 +249,91 @@ func getTargetReplicas(ds *disaggregatedsetv1.DisaggregatedSet, roleName string)
...
+// the minUnit-multiplier semantics.
+func isRevisionStable(rev disaggregatedsetutils.RevisionRoles, roleNames []string, config map[string]RollingUpdateConfig) bool {
+	// totalSteps for the projection = max role size across the new revision.
+	roleSizes := make(map[string]int, len(roleNames))
+	for _, name := range roleNames {
+		if lws := rev.Roles[name]; lws != nil {
+			roleSizes[name] = int(getLWSReplicas(lws))
+		}
+	}
+	totalSteps := maxRoleSize(roleSizes)
```
BenjaminBraunDev: totalSteps here should use the same basis as the planner and scaleUpNew, max(initialOld, target). Using the current spec makes tolerance too high early in the rollout, this returns true with spec 2P/1D, zero pods ready, target 8P/4D, S=1/U=1.

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -337,79 +412,211 @@ func (executor *RollingUpdateExecutor) scaleDownOld(
...
+					replicas := int(getLWSReplicas(lws))
+					if replicas > 0 && replicas-plannedDrain[name] == 0 {
+						plannedDrain[name] = 0
+					}
 				}
 			}
 		}
 
-		for i, name := range roleNames {
+		revisionDrained := false
```
BenjaminBraunDev: If the orphan-skip zeroes all of the newest revision's planned drains. We probably want to enforced the newest-first drain yes? I think a skip caused by aliveness should break, not continue.

pkg/controllers/disaggregatedset/planner.go
```diff
@@ -14,376 +14,340 @@ See the License for the specific language governing permissions and
...
 	}
-	return true
-}
-
-func computeMinOld(initialOld, currentNew, targetNew RoleReplicaState, config []RollingUpdateConfig) []int {
-	minOld := make([]int, len(initialOld))
-	for i := range initialOld {
-		if initialOld[i] >= targetNew[i] {
-			minOld[i] = max(0, targetNew[i]-config[i].MaxUnavailable-currentNew[i])
+	newSpecStep := min(newProgressStep+newSurge, newTotalSteps)
```
BenjaminBraunDev: The old side is greedy (oldSpecStep gets the full unavail budget) but the new side rebuilds at depth 1+surge. With S=0/MU=2: capacity dips by 2 immediately, then rebuilds one at a time, so the degraded period lasts about twice as long as needed.

I tried this with one ComputeNextStep call: old drained to 3 of 5, zero ready, addBudget 2 per role, planner asks for 1. Was that deliberate? If not, adding the earned lead like max(0, oldProgress - newProgress) to reflect overdrain and overready from our budget would allow us to make up that ground.

test/e2e/disaggregatedset/e2e_test.go
```diff
@@ -569,19 +569,21 @@ var _ = Describe("DisaggregatedSet E2E Tests", Ordered, func() {
...
-					{OldPrefill: 4, OldDecode: 1, NewPrefill: 4, NewDecode: 5},  // step 6: new role0 +1, new role1 +1
-					{OldPrefill: 2, OldDecode: 1, NewPrefill: 4, NewDecode: 5},  // step 7: old role0 -2
-					{OldPrefill: 2, OldDecode: 1, NewPrefill: 5, NewDecode: 7},  // step 8: new role0 +1, new role1 +2
-					{OldPrefill: 2, OldDecode: 1, NewPrefill: 6, NewDecode: 8},  // step 9: new role0 +1, new role1 +1
-					{OldPrefill: 0, OldDecode: 0, NewPrefill: 6, NewDecode: 8},  // step 10: old role0 -2, old role1 -1
+					{OldPrefill: 9, OldDecode: 1, NewPrefill: 1, NewDecode: 1},  // step 1
+					{OldPrefill: 8, OldDecode: 1, NewPrefill: 2, NewDecode: 2},  // step 2
+					{OldPrefill: 7, OldDecode: 1, NewPrefill: 3, NewDecode: 3},  // step 3
+					{OldPrefill: 7, OldDecode: 1, NewPrefill: 3, NewDecode: 4},  // step 4
+					{OldPrefill: 6, OldDecode: 0, NewPrefill: 4, NewDecode: 5},  // step 5
```
BenjaminBraunDev: This table shows OldDecode at 0 from step 5 while OldPrefill keeps draining, but the planner golden for the same config keeps oldD at 1 until the final tick, and the package doc says all roles reach 0 in the same tick. How were these tables generated? Also the intermediates are only logged, so a wrong table passes CI. Can we assert the aliveness/floor invariants over observed states?

BenjaminBraunDev (commented):
Thanks Mathis, just three things block lgtm for me: the deadlock fallback interaction, nil meaning both complete and blocked (with a reachable permanent wedge), and the undisclosed change to what per-role maxSurge/maxUnavailable mean.

One suite-level ask: the deleted TestNRole_UnavailableConstraint (old+new >= target - maxUnavailable at every step) needs a replacement. I think it will help with 2 of my floor violation findings.

ahg-g:
Is this one ready and all comments addressed?

keps/849-DisaggregatedSet-HPA/README.md
```diff
@@ -340,7 +340,7 @@ The DS controller writes back to each scaler's status at the end of every reconc
 
 ### Rolling Update Interaction
 
-`scaler.spec.replicas` is the target for the role's post-rollout steady state. The DS controller feeds it as the new-revision LeaderWorkerSet's target; old revisions continue to drain on the schedule the planner set at rollout start (the pre-existing `initial-replicas` annotation mechanism), independent of the scaler.
+`scaler.spec.replicas` is the target for the role's post-rollout steady state. The DS controller feeds it to the planner and records it on the current revision as `intended-replicas`. If another template update interrupts the rollout, that value freezes as the revision's completed target; the old-side baseline is the per-role maximum intended count across the interrupted revisions, while their current Specs are still summed for physical capacity accounting.
```
hasB4K: For codex: This is absolutely not readable, it feels giberrish. please change that

keps/766-DisaggregatedSet/README.md
```diff
@@ -353,6 +401,13 @@ to implement this enhancement.
 - 2026-03-05: Initial KEP draft
 - 2026-03-22: Updated to reflect N-dimensional roles API
 - 2026-03-23: Renamed "phase" to "role" throughout for semantic clarity
+- 2026-09-14: Updated the rollout contract: replaced strict readiness gating
+  with a bounded pending-work window, defined fractional lockstep and its skew
+  bound, documented availability-safe drain-before-grow ordering, and made
+  whole-revision retirement best effort with a per-role-safe liveness fallback.
+- 2026-09-14: Defined interrupted-rollout baselines as the per-role maximum of
+  immutable revision targets, so deleting a partially rolled-out revision does
+  not shrink the availability baseline.
```
hasB4K: codex: make only one change please.

keps/849-DisaggregatedSet-HPA/README.md
```diff
@@ -457,4 +457,3 @@ Skip the new CRD; instead embed autoscaling target/current fields on the role it
 Same CRD, but with an explicit `spec.targetRef {name, role}` field the user fills in. The DS controller reads whatever scaler the user created and attaches a non-controller ownerRef for GC.
```
hasB4K: codex: add ## Implementation History in this file too.

keps/766-DisaggregatedSet/README.md
```diff
@@ -84,7 +87,10 @@ We propose adding a new CRD called `DisaggregatedSet` that acts as a higher-leve
 
 **Risk**: The N-dimensional rolling update algorithm adds complexity that could lead to stuck rollouts.
 
-**Mitigation**: The algorithm is designed with safety invariants (scale up before scale down, coordinated drain) and the controller is stateless, allowing manual intervention by scaling replicas directly if needed.
+**Mitigation**: The executor enforces hard per-role surge, availability, and
+pending-work bounds. It applies only availability-safe old-replica drains
+before growing the new revision, prefers whole-revision retirement, and is
+stateless so reconciliation can safely resume after a restart.
```
hasB4K: codex: this is not clear - keep the tone (and formatting) of the KEP. It should not be gibberish

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -43,11 +43,26 @@ const (
...
-	Client     client.Client
 	Record     events.EventRecorder
 	LWSManager *LeaderWorkerSetManager
 }
 
+// roleRolloutState separates work already issued through Spec from work that
+// has completed and is serving through Ready. The planner advances Spec; Ready
+// only authorizes additional in-flight work and old-replica drains.
+type roleRolloutState struct {
+	InitialOld int
```
hasB4K: why using InitialOld and not IntentedReplica ?

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -43,11 +43,26 @@ const (
...
+	OldSpec    int
+	OldReady   int
```
hasB4K: why OldSpec? oldReady?

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -43,11 +43,26 @@ const (
...
 type RollingUpdateExecutor struct {
-	Client     client.Client
 	Record     events.EventRecorder
 	LWSManager *LeaderWorkerSetManager
 }
 
+// roleRolloutState separates work already issued through Spec from work that
+// has completed and is serving through Ready. The planner advances Spec; Ready
+// only authorizes additional in-flight work and old-replica drains.
+type roleRolloutState struct {
```
hasB4K: why roleRolloutState? do we need this? should we rename this? if so, in what?

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -72,17 +87,24 @@ func (executor *RollingUpdateExecutor) ReconcileRollingUpdateNew(
 	if len(oldRevisions) == 0 {
 		return ctrl.Result{}, nil
 	}
+	// Migrate the legacy initial-replicas value before using Spec as a fallback:
+	// an old revision's Spec may already be partially drained. Objects carrying
+	// neither annotation use their current Spec as the best available baseline.
+	// Once written, intended-replicas is immutable for an old revision.
+	if err := executor.ensureOldIntendedReplicas(ctx, disaggregatedSet, oldRevisions); err != nil {
+		return ctrl.Result{}, err
+	}
```
hasB4K: how does this work? do we write the initial old into intended replicas instead?

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -72,17 +87,24 @@ func (executor *RollingUpdateExecutor) ReconcileRollingUpdateNew(
...
 
 	addedRoles, removedRoles := detectRoleChanges(roleNames, oldRevisions)
 	if len(addedRoles) > 0 || len(removedRoles) > 0 {
 		log.Info("Role changes detected", "added", addedRoles, "removed", removedRoles)
 	}
 
 	if newRevision == nil {
-		return executor.initRollingUpdate(ctx, disaggregatedSet, slice, revision, roleNames, roleConfigs, oldRevisions)
+		return executor.initRollingUpdate(ctx, disaggregatedSet, slice, revision, roleNames, roleConfigs, scalers)
```
hasB4K: what is scalers here?

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -72,17 +87,24 @@ func (executor *RollingUpdateExecutor) ReconcileRollingUpdateNew(
...
 	}
 
-	return executor.ReconcileRollingUpdate(ctx, disaggregatedSet, slice, oldRevisions, *newRevision, scalers)
+	return executor.ReconcileRollingUpdate(ctx, disaggregatedSet, oldRevisions, *newRevision, scalers)
 }
```
hasB4K: why is there no slice anymore?

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -92,89 +114,93 @@ func (executor *RollingUpdateExecutor) initRollingUpdate(
...
-				replicas = int(*roleLWS.Spec.Replicas)
-			}
-			// Address by the LWS's actual name so a legacy slice-0 object (whose name
-			// has no slice segment) is updated rather than missed.
-			if _, err := executor.LWSManager.SetInitialReplicas(ctx, disaggregatedSet.Namespace, roleLWS.Name, replicas); err != nil {
-				log.Error(err, "Failed to set initial-replicas annotation", "lws", roleLWS.Name)
-			}
-		}
-	}
-
```
hasB4K: no longer needed 'cause we are writing on the fly the intended replicas? what happens if someone removed the label?

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -92,89 +114,93 @@ func (executor *RollingUpdateExecutor) initRollingUpdate(
...
 	// Create new LWS objects (one per role) for the target revision with 0
-	// replicas. The next reconcile loop will start scaling them up.
+	// replicas, but persist the count each role is intended to reach. If another
+	// rollout interrupts this one, that target—not the partial Spec—is its
```
hasB4K: that target—not the partial Spec—is    <-  this ios not clear, avoid em dash for fuck sake, remove it in all this PR.

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -92,89 +114,93 @@ func (executor *RollingUpdateExecutor) initRollingUpdate(
...
 
-// ReconcileRollingUpdate executes one step of an in-progress rolling update:
-//  1. Wait for the new revision to stabilize (all roles have ReadyReplicas == Replicas).
-//  2. Compute the next scale step using the planner.
-//  3. Scale up new revision LWS objects.
-//  4. Scale down old revision LWS objects (newest-first, with coordinated drain).
+// ReconcileRollingUpdate executes one step of an in-progress rolling update.
+// Spec counts represent already-issued rollout work and drive the planner.
+// Ready counts represent completed work and bound both pending scale-up and
+// safe old-replica drain. Old revisions are drained newest-first.
```
hasB4K: the old comment was clearer, a step by step is better?

keps/766-DisaggregatedSet/README.md
```diff
@@ -84,7 +87,7 @@ We propose adding a new CRD called `DisaggregatedSet` that acts as a higher-leve
 
 **Risk**: The N-dimensional rolling update algorithm adds complexity that could lead to stuck rollouts.
 
-**Mitigation**: The algorithm is designed with safety invariants (scale up before scale down, coordinated drain) and the controller is stateless, allowing manual intervention by scaling replicas directly if needed.
+**Mitigation**: For each role, the controller respects `maxSurge` and `maxUnavailable` and limits how many new replicas can be waiting to become Ready at the same time. It normally scales all roles in an old revision to zero together. If that would leave the rollout permanently stuck, it may scale one role to zero first, but only when doing so still respects that role's `maxUnavailable` limit. The controller derives the rollout state from the LeaderWorkerSets, so it can safely continue after a restart.
```
hasB4K: why does this need to be changed? is the old sentence false?

keps/766-DisaggregatedSet/README.md
```diff
@@ -159,143 +162,161 @@ type LeaderWorkerSetTemplateSpec struct {
...
+
+| Observation | Old P | Old D | New P | New D |
+|-------------|------:|------:|------:|------:|
+| Initial     | 8 | 4 | 0 | 0 |
+| 1           | 6 | 3 | 2 | 1 |
+| 2           | 4 | 2 | 4 | 2 |
+| 3           | 2 | 1 | 6 | 3 |
+| 4           | 0 | 0 | 8 | 4 |
+
+The observations are planner checkpoints, not a promise that every cluster
```
hasB4K: ffs, for this entire file, please do not go to a new line without reason

test/testutils/disaggregatedset/fixtures/fixtures.go
```diff
@@ -118,9 +121,16 @@ spec:
...
 		image := p.Image
 		if image == "" {
-			image = "registry.k8s.io/pause:3.9"
+			// Default image depends on whether the caller wants slow-pod
+			// scripting (needs a shell) or just an empty runnable pod.
+			if slow {
+				image = "busybox:1.36"
+			} else {
+				image = "registry.k8s.io/pause:3.9"
+			}
```
hasB4K: Do we really need a slow rollout? I think so... but 🤷  it should not be too slow also... tests should be quite fast.

is there a better / easier way to write this?

pkg/controllers/disaggregatedset/planner.go
```diff
@@ -14,376 +14,340 @@ See the License for the specific language governing permissions and
...
-	initialPast := make([]int, numRoles)
-	copy(initialPast, initialOld)
-	initialNew := make([]int, numRoles)
-	steps := []UpdateStep{{Past: initialPast, New: initialNew}}
+	steps := []UpdateStep{initial}
 
-	for i := 0; i < maxSteps; i++ {
-		nextStep := ComputeNextStep(initialOld, currentOld, currentNew, target, config)
-		if nextStep == nil {
+	for range maxSteps {
```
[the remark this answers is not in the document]
hasB4K: No change is needed here: this repository targets Go 1.26, and ranging over an integer has been supported since Go 1.22. The code compiles in the full test run.

pkg/controllers/disaggregatedset/planner_test.go
```diff
@@ -24,1045 +24,632 @@ import (
...
-		name        string
-		sourceRole0 int
-		sourceRole1 int
-		targetRole0 int
-		targetRole1 int
-		config      []RollingUpdateConfig
-		expected    []UpdateStep
+// TestSideSize verifies the step-count helper: number of steps a side has =
+// smallest non-zero role replica count (so the slowest role can express each
+// step boundary). 0 indicates an empty side.
```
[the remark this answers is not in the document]
hasB4K: Fixed. The misleading wording was removed; sideSize is now documented and tested as the maximum role replica count.

pkg/controllers/disaggregatedset/executor.go
hasB4K: Fixed. The unsafe force-retirement fallback was removed. Drains are now capped by maxSafeDrain, and stale Ready counts are capped by Spec, so terminating pods cannot authorize draining the next revision.

pkg/controllers/disaggregatedset/executor.go
hasB4K: Fixed. Completion is checked independently before calling the planner. A nil planner result during an incomplete rollout is now treated as temporarily blocked and requeued.

pkg/controllers/disaggregatedset/executor.go
hasB4K: Fixed. Pending-readiness limits now use roleSize = max(initialOld, target), projected on the same shared step basis as the planner, rather than the current Spec.

pkg/controllers/disaggregatedset/executor.go
hasB4K: Fixed. Once the newest revision that can consume the drain budget is selected, scaleDownOld returns even if coordinated retirement postpones its drain. The budget no longer spills into an older revision.

pkg/controllers/disaggregatedset/planner.go
hasB4K: Fixed. New-side progress now includes progress already earned by the old-side drain. The zero-surge, maxUnavailable=2 case rebuilds two replicas in the next step and is covered by unit and e2e tests.

test/e2e/disaggregatedset/e2e_test.go
hasB4K: Fixed. The e2e now records each distinct Spec state and asserts per-role surge ceilings, Ready-based availability floors, monotonic progress, and coordinated retirement. It allows one transient incomplete observation for the two non-atomic LWS updates, but the next Spec state must finish retirement. Exact step sequences remain in planner unit tests.

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -38,223 +39,240 @@ const (
...
-func buildPlannerState(
+// committedReadyReplicas returns the Ready capacity that can authorize another
+// scale-down. Status can briefly report more Ready replicas than Spec after a
+// previous scale-down; those excess replicas are already terminating and must
+// not be counted again.
+func committedReadyReplicas(lws *leaderworkersetv1.LeaderWorkerSet) int {
+	if lws == nil {
+		return 0
+	}
+	return max(0, min(int(lws.Status.ReadyReplicas), int(getLWSReplicas(lws))))
```
yankay: Could this reuse Ready capacity already consumed by a previous scale-down? With target=6, `maxUnavailable=1`, old Spec/Ready=`4/3`, and new=`3/3`, scaling old to 3 still leaves `min(3, 3)=3` if status lags, allowing another drain to 2. If both drains remove Ready replicas before replacements become Ready, only 4 remain, below the floor of 5. Could we account for pending drains and cover this with a consecutive-reconcile regression test?

hasB4K: yeah, nice catch!
Fixed: https://github.com/kubernetes-sigs/lws/pull/907/changes/5d7095719e74bccb88409fdaf3277c79845bb6f7

ahg-g:
Can we pls update https://github.com/kubernetes-sigs/lws/tree/main/site/content/en/docs/concepts/disaggregatedset (which will show up at https://lws.sigs.k8s.io/docs/examples/disaggregatedset/) with a Rollouts section that clearly documents the DisaggSet rollout semantics?

ahg-g:
> Can we pls update https://github.com/kubernetes-sigs/lws/tree/main/site/content/en/docs/concepts/disaggregatedset with a Rollouts section that clearly documents the DisaggSet rollout semantics?

Not necessarily the algorithm in detail, but users facing docs that allow them to understand what to expect when they kickoff an update, how to monitor it and what they can/can't do during a rolling update.

ahg-g:
> > Can we pls update https://github.com/kubernetes-sigs/lws/tree/main/site/content/en/docs/concepts/disaggregatedset with a Rollouts section that clearly documents the DisaggSet rollout semantics?
> 
> Not necessarily the algorithm in detail, but users facing docs that allow them to understand what to expect when they kickoff an update, how to monitor it and what they can/can't do during a rolling update.

Actually we should add the user-facing docs after this PR is released.

api/disaggregatedset/v1/disaggregatedset_types.go
```diff
@@ -37,7 +37,12 @@ const (
 	// Applied to LWS and Service objects in the same namespace as the DisaggregatedSet.
 	RevisionLabelKey string = "disaggregatedset.x-k8s.io/revision"
 
-	// InitialReplicasAnnotationKey stores the initial replica count at rollout start.
+	// InitialReplicasAnnotationKey records the baseline used by the controller as
+	// initialOld when the revision becomes old. After this revision's rollout
+	// completes, the value matches its replica count. If a new revision is created
```
hasB4K: If it's old, the old revision is deleted anyway

keps/766-DisaggregatedSet/README.md
```diff
@@ -159,143 +162,141 @@ type LeaderWorkerSetTemplateSpec struct {
...
+Within each side, the planner uses discrete linear interpolation. Every role uses the same progress fraction. The resulting replica counts are rounded up to whole numbers.
+
+Each side measures progress on its own fractional scale. For one side, `roleReplicaCounts` is the list of replica counts for its roles. `positiveRoleReplicaCounts` is the same list without roles whose replica count is zero.
 
 ```
-newAtStep(i) = ceil(i * target / totalSteps)    // scale up: 0 → target
-oldAtStep(i) = initialOld - floor(i * initialOld / totalSteps)  // scale down: initialOld → 0
+fractionalStepCount      = max(roleReplicaCounts)
+smallestReplicaFraction  = 1 / max(roleReplicaCounts)
+largestReplicaFraction   = 1 / min(positiveRoleReplicaCounts)
```
hasB4K: Replica is always positive.

keps/766-DisaggregatedSet/README.md
```diff
@@ -159,143 +162,141 @@ type LeaderWorkerSetTemplateSpec struct {
...
 ```
 
-Where, per role:
+`fractionalStepCount` is the number of equal fractional steps between the start and end of one side. Fractional step `0` is the start and fractional step `fractionalStepCount` is the end, so there are `fractionalStepCount + 1` positions. A fractional step is not a reconcile iteration. The controller may remain at one fractional step or advance across more than one fractional step in a reconcile.
+
+`smallestReplicaFraction` is the distance covered by one fractional step. It comes from the largest role because one replica is the smallest fraction of that role. In an `8P/4D` side, Prefill is the largest role, so one Prefill replica represents `1/8` of the rollout and creates eight equal fractional steps. `largestReplicaFraction` is the width of the coordination window. It comes from the smallest non-zero role because one replica is the largest fraction of that role. Decode is the smallest role in this example, so one Decode replica represents `1/4` of the rollout. The controller therefore allows at most `1/4` difference between role progress.
```
hasB4K: We need to mention maxSurge / maxUnavailable here.

hasB4K: What happen when the window is small? Is it still faster? Test it.

keps/766-DisaggregatedSet/README.md
```diff
@@ -159,143 +162,141 @@ type LeaderWorkerSetTemplateSpec struct {
...
-changes either old or new replicas (Property 1), and surge-blocked scale-ups
-may require intermediate drain iterations, the reconcile count is typically
-higher than `totalSteps` (see Example 1: `totalSteps = 3` but 7 reconcile
-iterations).
+The distance between adjacent columns is `1/8`, the `smallestReplicaFraction`. The window is two columns wide, or `2/8 = 1/4`, the `largestReplicaFraction`. Decode=2 maps to step 5 and defines the window. Prefill is at step 6, so it may advance to step 7 but no further until Decode advances. The window is then recalculated.
 
-**Key Properties**:
+This moving window is the fractional-lockstep guarantee. Roles can move by different replica counts, and they do not have to occupy the same fractional step. A planner step starting inside the window remains inside it. If observed state is already outside the window, the planner does not reverse applied work; it stops the leading roles and lets lagging roles catch up. Per-role limits can trim different parts of a candidate step, so the planner reapplies the window after those limits rather than cancelling every drain when one role cannot move.
```
hasB4K: > the planner does not reverse applied work; it stops the leading roles and lets lagging roles catch up.
How does that work?

keps/766-DisaggregatedSet/README.md
```diff
@@ -159,143 +162,141 @@ type LeaderWorkerSetTemplateSpec struct {
...
+```
+pendingAllowance = projected(role, MaxSurge + MaxUnavailable)
+newSpec - newCommittedReady <= pendingAllowance
+```
+
+This bounded window is what permits pipelining across slow pod starts. It does not grant every role the unscaled `MaxSurge + MaxUnavailable` sum. If independent pending bounds would separate role progress by more than `largestReplicaFraction`, faster roles wait at that coordination boundary.
+
+#### Reconcile ordering and completion
+
+One plan may contain both an old-side drain and new-side growth. Before returning the step, the planner limits old targets using committed Ready capacity and limits new targets using the surge and pending-readiness ceilings. It reapplies the coordination window after independent limits trim role targets. The executor then applies the floor-safe old drain before growing the new revision, so the two API updates cannot create a transient surge violation. If the bounded plan is a no-op, the controller requeues rather than mistaking the no-op for completion.
```
hasB4K: not clear

keps/766-DisaggregatedSet/README.md
```diff
@@ -159,143 +162,141 @@ type LeaderWorkerSetTemplateSpec struct {
...
+|-------------|------:|------:|------:|------:|
+| Initial     | 8 | 4 | 0 | 0 |
+| 1           | 6 | 3 | 2 | 1 |
+| 2           | 4 | 2 | 4 | 2 |
+| 3           | 2 | 1 | 6 | 3 |
+| 4           | 0 | 0 | 8 | 4 |
+
+The observations are planner fractional steps, not a promise that every cluster will expose exactly this sequence. Readiness, API observations, and interrupted updates may introduce additional reconciles.
+
+The pending window changes the slow-start case materially. After observation 1, suppose the new `2P/1D` is still unready. The next reconcile may still issue up to `4P/2D` because that is the proportional pending allowance. It may drain only availability that is actually committed; it cannot count those unready replicas, or stale Ready status above an already-reduced Spec, toward the floor. Once Ready advances, pending capacity opens and the pipeline continues.
```
hasB4K: It may drain only availability that is actually committed; it cannot count those unready replicas, or stale Ready status above an already-reduced Spec, toward the floor. Once Ready advances, pending capacity opens and the pipeline continues.

not clear, you are refering to MaxSurge/MaxUnavailable?

keps/766-DisaggregatedSet/README.md
```diff
@@ -353,6 +354,8 @@ to implement this enhancement.
 - 2026-03-05: Initial KEP draft
 - 2026-03-22: Updated to reflect N-dimensional roles API
 - 2026-03-23: Renamed "phase" to "role" throughout for semantic clarity
+- 2026-09-14: Updated the rolling-update contract to cover fractional lockstep, readiness and availability bounds, how old revisions are removed, and how an interrupted rollout remembers the replica counts it was meant to reach.
+- 2026-09-21: Updated interrupted rollouts to process one old revision at a time, removing revisions with no Ready replicas first and parking the remaining revisions.
```
hasB4K: merge into one.

keps/766-DisaggregatedSet/README.md
```diff
@@ -307,7 +308,7 @@ Headless Services are automatically created for each role per revision. This all
 
 ### Controller Architecture
 
-The controller is stateless—all state is derived from observed resources. An `initial-replicas` annotation tracks the starting replica count for rolling updates. Owner references on managed LeaderWorkerSets and Services ensure proper garbage collection.
+The controller is stateless: all state is derived from observed resources. The `initial-replicas` annotation preserves each revision's replica target across rolling updates. Owner references on managed LeaderWorkerSets and Services ensure proper garbage collection.
```
hasB4K: In the code: consider `initial-replicas` or 0 not set (for safety)

pkg/controllers/disaggregatedset/scaler_manager.go
```diff
@@ -56,6 +56,48 @@ func NewScalerManager(c client.Client, r events.EventRecorder) *ScalerManager {
...
+	return desiredReplicasByRole
+}
+
+// Every desired role should normally have a resolved target after
+// ScalerManager.Reconcile. Currently, a target can remain unresolved only when
+// an External role's generated scaler name is occupied by an object this
+// DisaggregatedSet does not own, so the scaler can be neither created nor
+// adopted. This check is a safety mechanism: callers must pause instead of
+// guessing a target.
+func unresolvedReplicaTargetRoles(roleNames []string, desiredReplicasByRole map[string]int) []string {
```
hasB4K: Is this really useful?

pkg/controllers/disaggregatedset/executor.go
```diff
@@ -373,128 +449,221 @@ func (executor *RollingUpdateExecutor) scaleDownOld(
...
+	parkedReadyReplicas RoleReplicaState,
+	snapshot rolloutSnapshot,
+) bool {
+	anyAliveAfter, anyRetired, canRetire := false, false, true
+	for i, name := range roleNames {
+		lws := roles[name]
+		if lws == nil || getLWSReplicas(lws) == 0 {
+			continue
+		}
+		replicas := int(getLWSReplicas(lws))
```
ahg-g: **[High] Permanent deadlock in `A → B → C` rollouts when the active old revision (`B`) has one Ready role and one unready (`0` Ready) role**

`canRetire` (L552) and `neededForRetirement` (L600) compare `int(getLWSReplicas(lws))` (`spec.replicas`) against `maxSafeDrain(snapshot[i])`. However, `maxSafeDrain` bounds how many **committed Ready** replicas can be removed above `availabilityFloor`. Retiring an LWS (`drain == spec.replicas`) only removes `committedReadyReplicas(lws)` Ready replicas—any unready replicas (`spec.replicas - committedReadyReplicas(lws)`) contribute `0` to `committedReady`.

#### Reproduction (`4P / 4D`, `MaxSurge = 1, MaxUnavailable = 0`, `availabilityFloor = 4`, `surgeCeiling = 5`)

Suppose a rollout from **A → B** is interrupted because `B.Prefill` has a broken image (`Ready = 0`), while `B.Decode` has `1` Ready pod, and the user deploys **C** to fix it:
    - **Rev A (oldest, parked)**: `Prefill: Spec=3, Ready=3 (Initial=4)`; `Decode: Spec=3, Ready=3 (Initial=4)`
    - **Rev B (active old)**: `Prefill: Spec=1, Ready=0 (Initial=4)`; `Decode: Spec=1, Ready=1 (Initial=4)`
    - **Rev C (new target)**: reaches `Prefill: Spec=1, Ready=1`; `Decode: Spec=1, Ready=1`                       

On the next reconcile:                                                                                                                  
    1. `selectRevisionToDrain` picks **B** (`revisionIsFullyUnready(B)` is `false` because `B.Decode.Ready == 1`), with `parkedReadyReplicas = [3, 3]` from **A**.
    2. `maxSafeDrain(Prefill)` is `max(0, (3 + 0 + 1) - 4) = 0`, and `maxSafeDrain(Decode)` is `max(0, (3 + 1 + 1) - 4) = 1`, so `ComputeNextStep` plans `drain = [0, 1]` on **B**.
    3. `coordinateRevisionDrain` checks `canRetire`: for `B.Prefill`, `replicas (1) <= maxSafeDrain (0)` evaluates to `false`—even though `B.Prefill` has **0 Ready replicas** and retiring it loses **0** availability.
    4. `coordinateRevisionDrain` cancels `B.Decode`'s `1 → 0` drain (`drain` becomes `[0, 0]`) and tries the liveness escape hatch (L594–L609) to grow `C.Prefill` by `neededForRetirement = 1`.
    5. **However, `C.Prefill` cannot grow past `1`** because:
       - `phaseTarget = NewTargetReplicas (4) - parkedReadyReplicas[0] (3) = 1` (L602–L606), AND
       - `hardNewReplicaLimits` caps `C.Prefill` at `surgeCeiling (5) - OldSpecReplicas (3 + 1 = 4) = 1` (`B.Prefill`'s unready spec replica is consuming the `MaxSurge = 1` slot).                                           
    6. `coordinateRevisionDrain` returns `blocked = true` (`drain = [0, 0]`, `targetNew = [1, 1]`). `C.Prefill` waits for `B.Prefill` to drain, and `B.Prefill` waits for `C.Prefill` to grow—**permanent deadlock**.
                                                                                                        
#### Suggested Fix

Check `committedReadyReplicas(lws)` against `maxSafeDrain(snapshot[i])` when determining whether retiring a role is safe, and allow unready replicas on `activeRevision` to drain without consuming `maxSafeDrain`:
                  
```
replicas := int(getLWSReplicas(lws))                                                                                            
readyReplicas := committedReadyReplicas(lws)
anyAliveAfter = anyAliveAfter || replicas > drain[i]                                                                                                                                                                                                                        
anyRetired = anyRetired || drain[i] == replicas                                                                                                                                                                                                                             
canRetire = canRetire && readyReplicas <= maxSafeDrain(snapshot[i])
```
And on line 600:

```go
neededForRetirement := committedReadyReplicas(lws) - maxSafeDrain(snapshot[i])              
```

pkg/utils/disaggregatedset/utils.go
```diff
@@ -220,19 +230,19 @@ func (revisions RevisionRolesList) GetTotalReplicasPerRole(role string) int {
 	return total
 }
 
-func (revisions RevisionRolesList) GetTotalInitialReplicasPerRole(role string) int {
-	total := 0
+func (revisions RevisionRolesList) GetMaxInitialReplicasPerRole(role string) int {
```
ahg-g: **[Medium] Retaining drained (`0`-replica) old revisions until `complete == true` inflates `InitialOldReplicas` and throttles cold-start recovery**                                                                                                                                 

Because `reconcileSlice` now delays `cleanupDrainedLWS` until `complete == true`, old revisions that have already been drained to `Spec = 0` remain in `oldRevisions` for the remainder of the rollout. This causes two issues:                                                     

1. **Phantom `InitialOldReplicas` baseline from drained/aborted revisions (`utils.go` L233–L246)**:                                                                                                                                                                                 
       When Revision B is created at `startingReplicas = 0`, its `initial-replicas` annotation is immediately stamped with B's desired target (e.g. `10`). If Revision A had `2` replicas, B was created for `10` (and only reached `0` or `1` replica before being superseded by       
  Revision C at `2` or `6` replicas), B drains to `0` immediately, **but `GetMaxInitialReplicasPerRole` still reads `initial-replicas = 10` from B's `0`-replica LWS**:                                                                                                                 
       - `surgeCeiling` becomes `10 + MaxSurge` instead of `2 + MaxSurge`.                                                                                                                                                                                                              
       - `leastAdvancedStep` treats `currentOld = 2` on A as already `80%` drained (`(10 - 2) / 10`).                                                                                                                                                                                   
       - If C targets `6` with `MaxUnavailable = 1`, `availabilityFloor` becomes `min(10, 6) - 1 = 5` even though the cluster only ever had `2` Ready pods.                                                                                                                             

2. **Cold-start recovery after `fullyUnready` drain is throttled by `pendingReadinessCeiling` (`executor.go` L189–L195)**:                                                                                                                                                          
       If Revision A (`8P / 4D`) is completely unready (`Ready = 0`) and the user deploys Revision B (`8P / 4D`, `MaxSurge = 1, MaxUnavailable = 0`), Reconcile 2 drains A to `0P / 0D` (`fullyUnready == true`). On Reconcile 3+, `OldSpecReplicas == 0` (zero old pods exist), yet    
  because A's `0`-replica LWS objects still exist in `oldRevisions`, B is kept on the rolling-update path where `hardNewReplicaLimits` caps B at `pendingReadinessCeiling = NewReadyReplicas + pendingAllowance` (`2P / 1D`), forcing a sequential batched cold start instead of        
  scaling directly to `availabilityFloor` (`8P / 4D`).                                                                                                                                                                                                                                  


#### Suggested Fix for `GetMaxInitialReplicasPerRole` 

Skip drained (`0`-replica) revisions when computing the active baseline if any non-zero old revision remains (or fall back to the max only if all are `0`):


```
    func (revisions RevisionRolesList) GetMaxInitialReplicasPerRole(role string) int {
        initial := 0
        for _, rev := range revisions {                                                                                                                                                                                                                                                 
            if lws := rev.Roles[role]; lws != nil {
                if getLWSReplicas(lws) == 0 {                                                                                                                                                                                                                                           
                    continue                                                                                                                                                                                                                                                            
                }
                initialReplicas, ok := GetInitialReplicas(lws)
                if ok {
                    initial = max(initial, int(initialReplicas))
                } else {
                    initial = max(initial, getLWSReplicas(lws))
                }
            }
        }
        return initial
    }
```

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
ahg-g: **[Medium] `deleteInForeground` triggers duplicate `RollingUpdateCompleted` / `LWSDeleted` events while foreground GC is running**

In a real cluster (unlike `fake.Client`), `deleteInForeground` (`metav1.DeletePropagationForeground`) does not immediately remove the `LeaderWorkerSet` from the API server: it sets `metadata.deletionTimestamp` and adds the `foregroundDeletion` finalizer while GC deletes child StatefulSets/Services.

1. Setting `deletionTimestamp` on the owned LWS immediately triggers a reconcile on `DisaggregatedSet`.
2. Because neither `GetRevisionRolesList` nor `cleanupDrainedLWS` checks `lws.DeletionTimestamp.IsZero()`, `len(oldRevisions) > 0` stays `true` while the `0`-replica LWS is terminating.
3. Every reconcile during foreground GC re-enters `ReconcileRevisionTransition` (emitting a duplicate `RollingUpdateCompleted` event at `executor.go:175`), re-runs `cleanupDrainedLWS` (calling `Delete` again and emitting a duplicate `LWSDeleted` event), and skips `reconcileSteadyState`.

    ```
            for _, lws := range roles {
                if !lws.DeletionTimestamp.IsZero() {
                    continue
                }
                log.Info("Deleting drained LWS", "name", lws.Name)
                if err := r.LWSManager.deleteInForeground(ctx, lws); err != nil {
                    return fmt.Errorf("failed to delete LWS %s: %w", lws.Name, err)
                }
                r.Record.Eventf(disaggregatedSet, nil, corev1.EventTypeNormal, EventReasonLWSDeleted,
                    "Delete", "Deleted drained LWS %s", lws.Name)
            }
    ```

*(Consider also skipping `0`-replica LWS objects with `!lws.DeletionTimestamp.IsZero()` in `LWSManager.GetRevisionRolesList` so `len(oldRevisions)` immediately reflects that all old revisions are already terminating.)*

Issue #849, opened 2026-05-07
[DisaggregatedSet] HPA support per role in DisaggregatedSet


hasB4K:
**What would you like to be added**:
We need a way to delegate the replicas number per role (either Prefill / Decode in this issue) to a small CRD that exposes a `/scale` API in k8s.

**Why is this needed**:
The idea would be to be able to have KEDA or something else being able to scale-up/scale-down replicas per role. In a Disaggregated Deployment, it's usually easier to scale-up/down per role, since the metrics are easier.

**Completion requirements**:
Now let's say that you are starting a rolling update version_A->version_B:
* DisaggregatedSet would need to freeze the old replicas count for version_B, and it will read the replicas count in that delegated CRD.
* Auto scaling up/down might need to be paused during rolling update, or at least we only need to scale up/down the B replicas per role.
* The planner uses the computed replicas in that `DisaggregatedSetRoleScaler`

I'm not exactly sure of the API, I need to think this through. But it's my next big task after my holidays.

This enhancement requires the following artifacts:

- [x] Design doc
- [x] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

Edwinhr716:
Is it not currently possible to autoscale using LWS directly? You can set the scaleTargetRef to be the LeaderWorkerSet we create https://lws.sigs.k8s.io/docs/examples/hpa/#step-2-create-hpa 

The only annoying part is that you have to create a new HPA object when going through rolling update

Edwinhr716:
@hasB4K can you assign this issue to yourself?

svasilinets:
> The only annoying part is that you have to create a new HPA object when going through rolling update

well, as I understand problem that  LWS  names are generated by operator a `tratata-6ad7c921-prefill` because the name includes a revision itself. So you can't easily go around it.

hasB4K:
> > The only annoying part is that you have to create a new HPA object when going through rolling update
> 
> well, as I understand problem that LWS names are generated by operator a `tratata-6ad7c921-prefill` because the name includes a revision itself. So you can't easily go around it.

@svasilinets yes absolutely, we had this discussion offline with @Edwinhr716, and that's indeed the reason.

PR #906, opened 2026-06-29
[docs] Add DisaggregatedSet placement policy KEP

BenjaminBraunDev:
#### What this PR does / why we need it
Adds KEP documentation for DisaggregatedSet Placement Policy feature. Allowing for optional preference or enforcement of colocation all same-slice resources on the same domain, and/or the spread of other-slice resources across separate domains.

(Some links reference the slices KEP #879, which is not yet merged as of writing this.)

Issue: #848

PR #910, opened 2026-07-03
KEP-849: Per-role HPA support for DisaggregatedSet

hasB4K:
#### What type of PR is this?

  /kind feature
  /kind documentation

Adds KEP-849 proposing per-role autoscaling for DisaggregatedSet through a new lightweight `DisaggregatedSetRoleScaler` CRD that exposes the `/scale` subresource. HPA, KEDA, or any `/scale`-aware controller can then drive replicas for a single role (e.g., prefill or decode) independently.

**Motivation**: pointing an HPA directly at the underlying LeaderWorkerSet breaks because LWS names include a revision hash (e.g., `my-llm-6ad7c921-prefill`), so the HPA orphans on every rolling update.

**Design**:
- A `DisaggregatedSetRoleScaler` targets exactly one (DisaggregatedSet, role) pair via `targetRef` and exposes `/scale`.
- `DisaggregatedRoleSpec` gains a `Scaling` sub-struct with `Mode`: `Static` (default, today's behavior) or `External`.
- When a role is `External`, the DS controller reads the desired replica count from the scaler and writes back `status.replicas` and `status.selector` so the HPA feedback loop closes across rollouts.
- Fully opt-in and per-role: DisaggregatedSets without any scaler behave exactly as they do today.

**Rolling-update interaction** is specified: the scaler drives only the new-revision target, old revisions keep their existing snapshot-and-drain behavior, and a monotonicity guard prevents an HPA scale-down from reversing a rollout in progress.

Refs #849

This PR only contains the KEP (design document); the implementation will follow in separate PR(s) once the design is approved.

```release-note
NONE
```


keps/849-DisaggregatedSet-HPA/README.md
```diff
@@ -0,0 +1,448 @@
...
+    //   - WaitingForScaler is NOT a scaler condition (see DS conditions)
+    //   - TargetMissing: True when the referenced DS or role does not exist
+    // +listType=map
+    // +listMapKey=type
+    // +optional
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+```
+
+**Uniqueness**: A validating webhook rejects the creation of a `DisaggregatedSetRoleScaler` if another scaler with the same `targetRef` already exists in the namespace. As a backstop against races, the scaler controller sets a `Conflicting` status condition and refuses to serve reads.
```
ahg-g: I am not sure we can rely on this since technically one can disable the webhook and create multiple instances of the scaler resource with the same targetRef. We need to define conflict resolution if multiple instances exist that point to the same target, for example the newest one wins.

keps/849-DisaggregatedSet-HPA/README.md
```diff
@@ -0,0 +1,448 @@
...
+
+**Owner reference**: On first reconcile, the DS controller adds an owner reference from the DS to the scaler with `Controller=false, BlockOwnerDeletion=false`. This ensures the scaler is garbage-collected if the DS is deleted, without transferring ownership away from the user.
```
ahg-g: I am not sure this is a conventional approach. Related, why not have the DS controller create the `DisaggregatedSetRoleScaler` resource automatically and manage it if the mode is set to `External`?

keps/849-DisaggregatedSet-HPA/README.md
```diff
@@ -0,0 +1,448 @@
...
+
+### Interaction with DisaggregatedSet Slices
+
+[KEP-846](/keps/846-disaggregatedset-slices) adds `spec.slices: int` to DisaggregatedSet, replicating the whole role topology into N independent copies. Each slice rolls independently, and `spec.roles[].spec.replicas` becomes a *per-slice* count (total pods per role = `replicas × slices`). LWS names gain a slice segment (`<ds>-<slice>-<revision>-<role>`).
+
+This has three unresolved implications for scaler-driven roles:
+
+1. **Scope of `scaler.spec.replicas`.** Three possible shapes:
+   - **Aggregate** — value is total across slices; controller divides among slices. Matches HPA's usual mental model.
+   - **Per-slice scaler CR** — one scaler per (DS, role, slice). Explicit; combinatorial (`roles × slices` scalers per DS).
```
Edwinhr716, line 364: Just to double check: By this do you mean that we can scale at the role level? So if I have two slices, each with a prefill and decode, I'll have to setup four different targetRefs?

keps/849-DisaggregatedSet-HPA/README.md
hasB4K: Yeah, this all targetRefs thing was over-engineered, thanks for your review there!
I followed your advice, DisaggregatedSetRoleScaler will be now created automatically 😉

keps/849-DisaggregatedSet-HPA/README.md
hasB4K: This no longer apply since DisaggregatedSetRoleScaler is now created automatically 😉

keps/849-DisaggregatedSet-HPA/README.md
hasB4K: I completely removed the targetRefs, it was over-engineered on my end.

However, yes, in this use case you would have 2 (Prefill+Decode) * 2 slices = 4 DisaggregatedSetRoleScaler.

I believe that this would be the simpler approach, but I think it's broken if each slice follow the policy placement feature of https://github.com/kubernetes-sigs/lws/pull/906, especially for NVL72...

NVL72 doesn't mean that's Prefill and Decode CANNOT communicate between racks, it's just less efficient (but even for a large KVCache, it might be fine). So we could:
1. "rebalance" the prefill and decode ratio/replicas per slice? This looks "ideal" but also very complex, and dangerous.
2. have a special "slice" that could help with the surge, that would be independent of the rest and that would not care of placement policy: that would be simpler in theory than a rebalance. But this would only work if the initial replicas of both prefill and decode are roughly always good.

I feel that the rebalancing is the way to go, it might happen quite rarely. But since all of this is speculative, and looks complex, I don't want to implement that now 😉

ahg-g:
/lgtm

Thanks!

PR #912, opened 2026-07-03
webhook: reject maxSurge=0+maxUnavailable=0 in DisaggregatedSet

### LINKED ISSUE #869: DisaggregatedSet webhook does not reject maxSurge=0 + maxUnavailable=0, causing infinite reconcile loop
**What happened**:

The LWS validating webhook rejects `maxSurge=0 + maxUnavailable=0` (when replicas > 0), but the DisaggregatedSet validating webhook does not. A DisaggregatedSet created with this combination is accepted; the controller's create-LWS call is then rejected by the LWS webhook on every reconcile, indefinitely (with controller-runtime's exponential back-off, observed intervals on the repro: 0s, 1s, 2s, 5s, 10s, 21s, 41s, 82s, ...).

User-visible symptoms:

- `.status` stays `{}`
- no Events on the DS object
- no child `LeaderWorkerSet` ever created
- only the controller-manager log shows the cause:

```
admission webhook "vleaderworkerset.kb.io" denied the request:
spec.rolloutStrategy.rollingUpdateConfiguration.maxUnavailable:
  Invalid value: 0: must not be 0 when `maxSurge` is 0
```

The DisaggregatedSet webhook should reject this spec up front, with the same field-level error the LWS webhook already produces.

`repro.yaml`:

```yaml
apiVersion: disaggregatedset.x-k8s.io/v1
kind: DisaggregatedSet
metadata:
  name: ds-zero-surge
spec:
  roles:
  - name: prefill
    spec:
      replicas: 2
      rolloutStrategy:
        rollingUpdateConfiguration:
          maxSurge: 0
          maxUnavailable: 0
      leaderWorkerTemplate:
        size: 1
        leaderTemplate:
          spec:
            containers:
            - name: w
              image: nginx:1.29
        workerTemplate:
          spec:
            containers:
            - name: w
              image: nginx:1.29
  - name: decode
    spec:
      replicas: 2
      rolloutStrategy:
        rollingUpdateConfiguration:
          maxSurge: 0
          maxUnavailable: 0
      leaderWorkerTemplate:
        size: 1
        leaderTemplate:
          spec:
            containers:
            - name: w
              image: nginx:1.29
        workerTemplate:
          spec:
            containers:
            - name: w
              image: nginx:1.29
```

```
$ kubectl apply -f repro.yaml
disaggregatedset.disaggregatedset.x-k8s.io/ds-zero-surge created

$ sleep 30; kubectl get lws -l disaggregatedset.x-k8s.io/name=ds-zero-surge
No resources found in default namespace.

$ kubectl get ds ds-zero-surge -o json | jq .status
{}

$ kubectl -n lws-system logs -l control-plane=controller-manager -c manager \
    | grep ds-zero-surge | tail -1
ERROR Reconciler error ... admission webhook "vleaderworkerset.kb.io" denied the request:
  spec.rolloutStrategy.rollingUpdateConfiguration.maxUnavailable: Invalid value: 0:
  must not be 0 when `maxSurge` is 0
```

Root cause: the LWS webhook already enforces this rule (`pkg/webhooks/leaderworkerset_webhook.go:173`):

```go
if maxUnavailableValue == 0 && maxSurgeValue == 0 && *lws.Spec.Replicas != 0 {
    allErrs = append(allErrs, field.Invalid(maxUnavailablePath, maxUnavailable,
        "must not be 0 when `maxSurge` is 0"))
}
```

The DisaggregatedSet webhook (`pkg/webhooks/disaggregatedset/`) does not mirror it.

Suggested fix: for each role with `replicas > 0`, reject when both `maxSurge` and `maxUnavailable` **resolve to 0** after percentage-to-int scaling (use `intstr.GetScaledValueFromIntOrPercent`, matching the LWS webhook behaviour). This means `0%` should also be rejected, not just the literal integer `0`. An all-zero DisaggregatedSet (every role with `replicas: 0`) is legitimate and should still be accepted, matching the LWS behaviour.

A separate but related improvement worth tracking elsewhere: when the DS controller fails to create or update a child LWS due to admission rejection, mirror the error into a `Degraded` condition on the DisaggregatedSet so users have a single place to look.

**Environment**:

- Kubernetes version (use `kubectl version`): v1.35.0 (kind v0.31.0, single node)
- LWS version (use `git describe --tags --dirty --always`): `main` (after #836, includes the unified controller binary)
- Cloud provider or hardware configuration: kind on Linux/amd64
- OS: Ubuntu 22.04
- Install tools: helm + manifests built from source
- Others: lws image `localhost:5004/lws:main`

/kind bug
/area disaggregatedset

Dasmat13:
## What this PR does

Fixes #869.

The LWS validating webhook already enforces that `maxSurge=0 + maxUnavailable=0`
is rejected when `replicas > 0`. The DisaggregatedSet validating webhook was
missing the same check, so a DisaggregatedSet with this combination was admitted;
the controller then tried to create a child LWS on every reconcile, which the LWS
webhook rejected each time, causing an infinite back-off loop with no user-visible
error on the DS object.

## Changes

**`pkg/webhooks/disaggregatedset/disaggregatedset_webhook.go`**
- Refactored `validateRoleRolloutStrategy` to extract a `rucPath` local for the
  `rollingUpdateConfiguration` child, making paths consistent.
- Added `validateRoleMaxSurgeUnavailable()` which mirrors the LWS webhook check
  ([leaderworkerset_webhook.go L171](https://github.com/kubernetes-sigs/lws/blob/main/pkg/webhooks/leaderworkerset_webhook.go#L171)):
  - Uses `intstr.GetScaledValueFromIntOrPercent` so percentage values like `"0%"`
    are also caught.
  - Roles with `replicas == 0` (or unset) are exempt — an all-zero DisaggregatedSet
    is legitimate, matching LWS behaviour.
- Added `validatePositiveIntOrPercentDS` and `isNotMoreThan100PercentDS` helpers for
  per-field validation before the combined zero check.

**`pkg/webhooks/disaggregatedset/disaggregatedset_webhook_test.go`**
- Updated the `"valid with partition set to 0"` test to add `MaxUnavailable: 1`
  (the test was previously implicitly relying on the absence of the combined check).
- Added four new `TestValidateCreate` cases:
  - ❌ `maxSurge=0 + maxUnavailable=0`, replicas > 0, int literal
  - ❌ `maxSurge=0% + maxUnavailable=0%`, replicas > 0, percentage form
  - ✅ `maxSurge=0, maxUnavailable=1` (only one zero — valid)
  - ✅ `maxSurge=0 + maxUnavailable=0`, replicas = 0 (zero-replica exemption)
- Added one new `TestValidateUpdate` case for the same rule on updates.

## Reproducer (from issue)

```yaml
apiVersion: disaggregatedset.x-k8s.io/v1
kind: DisaggregatedSet
spec:
  roles:
  - name: prefill
    spec:
      replicas: 2
      rolloutStrategy:
        rollingUpdateConfiguration:
          maxSurge: 0
          maxUnavailable: 0   # ← now rejected at admission
      leaderWorkerTemplate: { ... }
  - name: decode
    spec: { ... }
```

Before this PR: admitted → infinite reconcile loop.
After this PR: rejected at admission with:
```
spec.roles[0].spec.rolloutStrategy.rollingUpdateConfiguration.maxUnavailable:
  Invalid value: 0: must not be 0 when `maxSurge` is 0
```

### DIFF
--- a/pkg/webhooks/disaggregatedset/disaggregatedset_webhook.go
+++ b/pkg/webhooks/disaggregatedset/disaggregatedset_webhook.go
@@ -19,12 +19,14 @@ package disaggregatedset
 import (
 	"context"
 
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	disaggv1 "sigs.k8s.io/lws/api/disaggregatedset/v1"
 	leaderworkerset "sigs.k8s.io/lws/api/leaderworkerset/v1"
+	"sigs.k8s.io/lws/pkg/webhooks"
 )
 
 // DisaggregatedSetWebhook handles validation for DisaggregatedSet resources.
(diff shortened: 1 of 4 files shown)


Dasmat13:
Hi @yankay and @ahg-g! Just a friendly follow-up on this PR. All CI checks are passing and I believe it is ready for review. Whenever you have time, I'd really appreciate your feedback. Thank you!

config/rbac/role.yaml
```diff
@@ -122,26 +122,33 @@ rules:
...
   verbs:
+  - create
+  - delete
   - get
+  - list
   - patch
   - update
+  - watch
+- apiGroups:
+  - leaderworkersetv1.x-k8s.io
```
ahg-g, line 150: where did the `leaderworkersetv1.x-k8s.io` api group come from? the apigroup used by this project is `leaderworkerset.x-k8s.io`

config/rbac/role.yaml
ahg-g, line 101: are we keeping the rbac roles aligned between kustomize and the helm chart?

ahg-g: also, why do we need to change rbac roles at all in this PR? 

@yankay @Edwinhr716 @BenjaminBraunDev it seems we need a full audit of all latest changes to rbac to tighten the permissions and align it with the chart.

pkg/webhooks/disaggregatedset/disaggregatedset_webhook.go
```diff
@@ -88,15 +91,107 @@ func (w *DisaggregatedSetWebhook) validateRoleRolloutStrategy(role disaggv1.Disa
...
+			maxUnavailablePath,
+			maxUnavailable,
+			"must not be 0 when `maxSurge` is 0",
+		))
+	}
+	return allErrs
+}
+
+// validatePositiveIntOrPercentDS tests if a given value is a valid non-negative int or percentage.
+func validatePositiveIntOrPercentDS(intOrPercent intstr.IntOrString, fldPath *field.Path) field.ErrorList {
```
ahg-g: we have similar functions in the parent pkg lets not duplicate them

ahg-g:
@Dasmat13 thanks for the fix, do you plan to address the RBAC roles issue? I see that your PR at https://github.com/kubernetes-sigs/lws/commit/87cf4a98fa30c92b5ced16f29b775ec4b93d4013 also made changes to RBAC that may not be necessary and probably incorrect (like the introduction of leaderworkersetv1.x-k8s.io as one the apigroups)

PR #913, opened 2026-07-03
webhook: reject LeaderWorkerSet replicas outside [0, 1000000] range

### LINKED ISSUE #908: Reject negative and huge number of replicas in LeaderWorkerSet
https://github.com/kubernetes-sigs/kueue/pull/12715

**Why is this needed**:
Reject negative and huge number of replicas in LeaderWorkerSet, to make sure that bad actor doesn't crash lws controller by constructing bad input

Dasmat13:

LeaderWorkerSet currently accepts any `int32` value for `spec.replicas`, including negative numbers and astronomically large values. A replica count of `math.MaxInt32` (2,147,483,647) causes the controller to compute pod counts that overflow and crash the reconcile loop — this was identified in kubernetes-sigs/kueue#12715 as a source of controller panics in the Kueue/LWS integration.

This PR adds two complementary layers of defence:

1. **OpenAPI schema constraints** (`+kubebuilder:validation:Minimum=0`, `+kubebuilder:validation:Maximum=1000000`) on the `Replicas` field so the API server rejects invalid values before they reach the controller.
2. **Webhook runtime validation** in `generalValidate` that returns a clear error message when `replicas < 0` or `replicas > 1_000_000`, providing defense-in-depth for clients that bypass schema validation.

Regenerated CRDs and Helm chart CRDs reflect the new constraints.

## How to test

Integration tests added in `test/integration/webhooks/leaderworkerset_test.go`:
- `creation with replicas greater than 1000000 should fail`
- `update with invalid replicas should fail (larger than 1000000)`

Run with:
```
make test-integration
```

## Related issues / PRs

Closes #908
Ref: kubernetes-sigs/kueue#12715

### DIFF
--- a/api/leaderworkerset/v1/leaderworkerset_types.go
+++ b/api/leaderworkerset/v1/leaderworkerset_types.go
@@ -119,6 +119,8 @@ type LeaderWorkerSetSpec struct {
 	//
 	// +optional
 	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=1000000
 	Replicas *int32 `json:"replicas,omitempty"`
 
 	// leaderWorkerTemplate defines the template for leader/worker pods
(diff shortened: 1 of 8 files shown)


Edwinhr716:
Thanks!
/lgtm
/approve

PR #909, opened 2026-07-03
Fix cert-manager CA injection for Helm webhooks

1cbyc:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it

The Helm chart enables cert-manager for webhook TLS, but the mutating and validating webhook configurations were missing the `cert-manager.io/inject-ca-from` annotation. Without that annotation, cert-manager does not inject the CA bundle into the webhook configurations, so API server webhook calls can fail with x509 unknown authority errors.

This PR adds the cert-manager CA injection annotation to both Helm-rendered webhook configurations when `.Values.enableCertManager` is enabled, matching the working kustomize flow and pointing at the generated `lws-serving-cert` in the release namespace.

#### Which issue(s) this PR fixes

Fixes #902

#### Special notes for your reviewer

This only affects Helm installs with cert-manager enabled. The annotation is guarded by `.Values.enableCertManager`, so installs that manage webhook certificates another way should be unchanged.

#### Does this PR introduce a user-facing change?

```release-note
Fixed Helm chart webhook TLS configuration to inject the cert-manager CA bundle into mutating and validating webhook configurations when cert-manager is enabled.
```

Issue #902, opened 2026-06-25
Helm installation with enableCertManager: true is not working


EladDolev:
hello there 👋 

when installing using Helm with `enableCertManager: true` the webhooks are not working.
getting the following in CLI when trying to submit LWS resources with kubectl
```
Internal error occurred: failed calling webhook "mleaderworkerset.kb.io": failed to call webhook: Post "https://lws-webhook-service.<ns>.svc:443/mutate-leaderworkerset-x-k8s-io-v1-leaderworkerset?timeout=10s": tls: failed to verify certificate: x509: certificate signed by unknown authority
```
and can see the following in the logs `http: TLS handshake error from 172.16.0.213:48502: EOF` but nothing besides that with logs level 10 and nothing is showing up in the metrics like `certwatcher_read_certificate_errors_total` for example

Prometheus is very happy with cert-manager enabled BTW, though scraping is not working at all with or without cert-manager because of a bug in the chart but that's a different issue

both `lws-metrics-server-cert` and `lws-webhook-server-cert` secrets exist with populated certs and Certificate objects look healthy with no errors reported from cert-manager

when `enableCertManager: false` everything works just fine

happy to provide more details or perform any test that may help

thanks !!

**Environment**:

- Kubernetes version (use `kubectl version`): `v1.33.12-gke.1000000`
- LWS version (use `git describe --tags --dirty --always`): `0.9.0`
- Cloud provider or hardware configuration: GKE
- OS (e.g: `cat /etc/os-release`): COS

ardaguclu:
I think, there is no bug here. The expectation is to patch this https://github.com/kubernetes-sigs/lws/blob/main/config/crd/patches/cainjection_in_leaderworkersets.yaml. Did you do that?

EDITED: I might be wrong. There might be a bug. This may need to be patched on helm chart.

yankay:
Thanks for reporting this! I was able to reproduce the issue. When enableCertManager: true is used, the Helm chart disables the internal certificate manager, but the admission webhook configurations do not get the cert-manager.io/inject-ca-from annotation, so cert-manager never injects the CA bundle.

References:
JobSet: https://github.com/kubernetes-sigs/jobset/blob/main/charts/jobset/templates/webhook/manifests.yaml
Kueue: https://github.com/kubernetes-sigs/kueue/blob/main/charts/kueue/templates/webhook/manifests.yaml

/help-wanted
/good-first-issue

PR #914, opened 2026-07-03
fix: migrate deprecated commonLabels to labels for kustomize v5

### LINKED ISSUE #866: config/default: migrate deprecated commonLabels to labels (kustomize v5)
### What happened

`kubectl kustomize config/default` (kustomize v5+) emits:

```
# Warning: 'commonLabels' is deprecated. Please use 'labels' instead.
```

Only occurrence: `config/default/kustomization.yaml`.

### Migration note (please don't blindly `kustomize edit fix`)

`commonLabels` injects labels into both metadata and selectors. The new `labels:` field is metadata-only unless `includeSelectors: true` is set. To preserve current behavior — and avoid mutating immutable selector fields on already-deployed Deployments/Services — use:

```yaml
labels:
- includeSelectors: true
  pairs:
    app.kubernetes.io/name: lws
    app.kubernetes.io/instance: lws
```

A plain `labels:` block would drop these keys from existing selectors — a breaking upgrade for live clusters.

### Prior art in sibling projects

- **kueue** has migrated: [kubernetes-sigs/kueue#3091](https://github.com/kubernetes-sigs/kueue/pull/3091) ("Fix commonLabels deprecated warning"), merged.
- **jobset** still uses `commonLabels` in `config/default/kustomization.yaml` (same situation as lws); worth tracking there as well.

### Verification

- Diff `kubectl kustomize config/default` before/after — selector blocks identical.
- Warning gone.

Dasmat13:
## Summary
This PR resolves the kustomize v5 deprecation warning for `config/default/kustomization.yaml` (issue #866).

- Replaces the deprecated `commonLabels` field with the kustomize v5 `labels` syntax using `includeSelectors: true`. This preserves the existing behaviour of injecting the labels into both metadata **and** selectors — `includeSelectors: true` is required so selector-injected labels don't mutate immutable `.spec.selector` fields on live clusters.
- Prior art: kueue performed the same migration in kubernetes-sigs/kueue#3091.

## Note on the cert-manager CA injection (issue #902)
The cert-manager CA injection half of the original change (adding the
`cert-manager.io/inject-ca-from` annotation to the webhook configurations)
was already merged upstream via #909, so this PR now contains **only** the
`commonLabels` → `labels` migration. The `webhook.yaml` change is dropped to
avoid duplication/conflict with #909.

## How to verify (issue #866)
Before:
```
kubectl kustomize config/default   # emits: commonLabels is deprecated
```
After:
```
kubectl kustomize config/default   # no deprecation warning
```

## Related issues
Fixes: #866
(issue #902 resolved by #909)

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -9,9 +9,11 @@ namespace: lws-system
 namePrefix: lws-
 
 # Common labels to add to all resources and selectors (matches Helm behavior).
-commonLabels:
-  app.kubernetes.io/name: lws
-  app.kubernetes.io/instance: lws
+labels:
+  - includeSelectors: true
+    pairs:
+      app.kubernetes.io/name: lws
+      app.kubernetes.io/instance: lws
 
 resources:
 - ../crd


yankay:
Thanks for working on this :-)

The Helm cert-manager CA injection change overlaps with https://github.com/kubernetes-sigs/lws/pull/909

Dasmat13:
Hi @yankay, thanks again for the earlier feedback. I've updated the PR to remove the overlapping cert-manager change, leaving only the commonLabels → labels migration. All CI checks are now passing. When you have a chance, I'd appreciate another look. Thanks!

PR #916, opened 2026-07-07
[feat] Add placement policy field and support to DisaggregatedSet

### LINKED ISSUE #848: [DisaggregatedSet] Placement Policy

BenjaminBraunDev:
### Summary
Implements the DisaggregatedSet placement policy from KEP-848 (PR: #906, Issue: #848). Adds a `spec.placementPolicy` that makes the controller inject pod affinity/anti-affinity into the managed LeaderWorkerSet pod templates, so a slice's roles co-locate in a topology domain and a DisaggregatedSet's slices spread across domains. Builds on slices #846.

`spec.placementPolicy.type` + a topology (node-label) key:
  - None (default) — inject nothing
  - ExclusiveSlice — co-locate a slice's roles + spread this set's slices (other sets may share a domain)
  - ExclusiveTopology — the above, plus one slice per domain across all DisaggregatedSets

See KEP-848 for the design, affinity construction, and trade-offs.

Notes for the reviewer
  - Placement is injected at LWS-create time and is deliberately not part of the revision hash, so changing it on an existing DisaggregatedSet takes effect on the next rollout.
  - Validation rejects a non-None policy without topology, and rejects combining it with the LWS exclusive-topology annotation on a role (checked on the role metadata and both pod templates).

Testing: unit tests for the affinity builder, Create wiring (leader + worker templates), and webhook validation; an e2e spec asserting the affinity is injected into the created LWS.

Which issue(s) this PR fixes: Fixes #848

### Release Note
```release-note
DisaggregatedSet supports a `placementPolicy` for topology-aware placement, co-locating a slice's roles and spreading slices across domains via injected pod affinity/anti-affinity.
```


BenjaminBraunDev:
(Currently implemented on top of slices changes: #873 should be merged first)

pkg/utils/disaggregatedset/affinity.go
```diff
@@ -0,0 +1,99 @@
...
+		})
+
+	// Spread this DisaggregatedSet's slices: repel same-set pods of a different slice.
+	// Other DisaggregatedSets (different name) are not matched, so they may share.
+	affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution = append(
+		affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution,
+		corev1.PodAffinityTerm{
+			LabelSelector: &metav1.LabelSelector{MatchExpressions: []metav1.LabelSelectorRequirement{
+				{Key: disaggregatedsetv1.SetNameLabelKey, Operator: metav1.LabelSelectorOpIn, Values: []string{dsName}},
+				{Key: disaggregatedsetv1.SliceLabelKey, Operator: metav1.LabelSelectorOpNotIn, Values: []string{sliceStr}},
```
yankay: Thanks for adding the slice-spread term here. I think this needs one legacy-slice-0 special case: Kubernetes `NotIn` also matches pods where the label is absent, and pre-slices slice-0 pods do not have `disaggregatedset.x-k8s.io/slice`. In a Kind simulation, a new slice-0 pod with placement stayed Pending because its own legacy slice-0 pod matched this anti-affinity term. Could we make the slice-0 case require `slice Exists` before `slice NotIn ["0"]`, while keeping missing-slice pods treated as slice 0 for other slices?

pkg/webhooks/disaggregatedset/disaggregatedset_webhook.go
```diff
@@ -71,6 +74,57 @@ func (w *DisaggregatedSetWebhook) validateRoles(obj *disaggv1.DisaggregatedSet)
...
+
+	return allErrs
+}
+
+// roleHasExclusiveTopology reports whether a role carries the LWS group-level
+// exclusive-topology annotation anywhere it takes effect: the LWS metadata, or the
+// leader/worker pod templates (the LWS pod webhook reads it from the pod, so a
+// template-level annotation would enable LWS exclusive placement too).
+func roleHasExclusiveTopology(role disaggv1.DisaggregatedRoleSpec) bool {
+	if _, ok := role.ObjectMeta.Annotations[leaderworkerset.ExclusiveKeyAnnotationKey]; ok {
```
yankay: Thanks for adding this validation guard. I wonder if this should also check `leaderworkerset.sigs.k8s.io/subgroup-exclusive-topology` in the same metadata/template locations. The pod webhook injects affinity for that annotation too, so combining it with DisaggregatedSet placement seems to create the same overlapping-placement conflict that this validation is trying to prevent.

yankay (commented):
Thanks for the substantial work here. I reviewed both the slices plumbing and the placement-policy pieces, and also ran a Kind simulation against the PR build.

pkg/utils/disaggregatedset/affinity.go
BenjaminBraunDev: Thanks for running the simulation. Just fixed as you suggested: for slice 0 the spread term now requires slice Exists before the NotIn, while slices != 0 keep the current behavior.

pkg/webhooks/disaggregatedset/disaggregatedset_webhook.go
BenjaminBraunDev: Yeah good call. Generalized the check to cover both annotations across the role metadata and leader/worker templates, and the error message now names the annotation it actually found. Also added webhook test cases for the subgroup annotation.

pkg/utils/disaggregatedset/affinity.go
```diff
@@ -0,0 +1,113 @@
...
+	// domain, giving a 1:1 domain-to-slice mapping. slice Exists scopes the term to
+	// DisaggregatedSet-managed pods, so unrelated workloads may still share the domain.
+	if policy.Type == disaggregatedsetv1.PlacementExclusiveTopology {
+		affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution = append(
+			affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution,
+			corev1.PodAffinityTerm{
+				LabelSelector: &metav1.LabelSelector{MatchExpressions: []metav1.LabelSelectorRequirement{
+					{Key: disaggregatedsetv1.SetNameLabelKey, Operator: metav1.LabelSelectorOpNotIn, Values: []string{dsName}},
+					{Key: disaggregatedsetv1.SliceLabelKey, Operator: metav1.LabelSelectorOpExists},
+				}},
```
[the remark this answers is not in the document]
BenjaminBraunDev: done

ahg-g (commented):
can we have an integration test that tests how this placement policy composes with the one from lws if both set at the same time?

BenjaminBraunDev:
> can we have an integration test that tests how this placement policy composes with the one from lws if both set at the same time?

Added an integration test in test/integration/webhooks. To clarify the behavior we decided on: the webhook will reject combining a non-None placementPolicy with either LWS exclusive topology annotation, since both inject affinity at overlapping levels and the slice would never schedule.

PR #918, opened 2026-07-07
Fix SubGroupIndex assignment and TPU variables injection when SubGrou…

### LINKED ISSUE #917: Incorrect `SubGroupIndex` assignment and `TPU_WORKER_HOSTNAMES` injection when `SubGroupSize=1`
**What happened**:
Incorrect `SubGroupIndex` assignment and `TPU_WORKER_HOSTNAMES` injection when `SubGroupSize=1`

When configuring a `LeaderWorkerSet` with TPU workloads where each pod represents an independent 1-slice TPU endpoint (`subGroupPolicy.subGroupSize: 1`), each pod (leader and workers) should be assigned to its own subgroup of size 1 and injected with its own hostname in `TPU_WORKER_HOSTNAMES`. `TPU_WORKER_ID` on the leader should be set to `0` but it is instead set to `-1`

Create a `LeaderWorkerSet` with `size: 3` and `subGroupSize: 1` where pods request TPU resources (`google.com/tpu`):
```yaml
kind: LeaderWorkerSet
metadata:
  name: prefill
  namespace: default
spec:
  replicas: 1
  leaderWorkerTemplate:
    size: 3
    subGroupPolicy:
      subGroupSize: 1
    workerTemplate:
      spec:
        containers:
        - name: lws-member
          image: busybox
          resources:
            limits:
              google.com/tpu: "4"
```

## Observed Behavior
1. **Incorrect `subGroupIndex` labels assigned to pods:**
   - Pod `prefill-0` (Leader, `workerIndex=0`): `leaderworkerset.sigs.k8s.io/subgroup-index: "0"`
   - Pod `prefill-0-1` (`workerIndex=1`): `leaderworkerset.sigs.k8s.io/subgroup-index: "0"` *(Collision with Leader)*
   - Pod `prefill-0-2` (`workerIndex=2`): `leaderworkerset.sigs.k8s.io/subgroup-index: "1"`

2. **Incorrect `TPU_WORKER_HOSTNAMES` injected into worker containers:**
   - Pod `prefill-0` gets `TPU_WORKER_HOSTNAMES=prefill-0.prefill`
   - Pod `prefill-0-1` gets `TPU_WORKER_HOSTNAMES=prefill-0.prefill` *(Points to Leader instead of itself)*
   - Pod `prefill-0-2` gets `TPU_WORKER_HOSTNAMES=prefill-0-1.prefill` *(Points to Worker 1 instead of itself)*
   - Pod `prefill-0-2.prefill` is completely missing from all pods' `TPU_WORKER_HOSTNAMES`.

3. **Incorrect `TPU_WORKER_ID` injected into worker containers:**
   - Pod `prefill-0` gets `TPU_WORKER_ID=-1`

## Expected Behavior
When `size=3` and `subGroupSize=1` (`Size % SubGroupSize == 0`), each pod should be in its own subgroup (`subGroupIndex` 0, 1, 2 respectively), and each pod requesting TPUs should receive its own hostname in `TPU_WORKER_HOSTNAMES`:
- Pod `prefill-0` (`subGroupIndex=0`): `TPU_WORKER_HOSTNAMES=prefill-0.prefill`, `TPU_WORKER_ID=0`
- Pod `prefill-0-1` (`subGroupIndex=1`): `TPU_WORKER_HOSTNAMES=prefill-0-1.prefill`
- Pod `prefill-0-2` (`subGroupIndex=2`): `TPU_WORKER_HOSTNAMES=prefill-0-2.prefill`

## Root Cause Analysis
1. **`getSubGroupIndex` precedence bug (`pkg/webhooks/pod_webhook.go`):**
   ```go
   func getSubGroupIndex(podCount int, subGroupSize int, workerIndex int) string {
       if (podCount-1)%subGroupSize == 0 {
           return fmt.Sprint((workerIndex - 1) / subGroupSize)
       }
       return fmt.Sprint(workerIndex / subGroupSize)
   }
   ```
   When `subGroupSize=1`, `(podCount-1)%1 == 0` is always true. `getSubGroupIndex` takes the `if (podCount-1)%subGroupSize == 0` branch even when `podCount % subGroupSize == 0`. Checking `if podCount % subGroupSize == 0` first (or explicitly handling `subGroupSize == 1`) ensures equal-sized subgroups are created when `Size` is divisible by `SubGroupSize`.

2. **`addTPUVariablesSubGroup` slice calculation bug (`pkg/utils/accelerators/tpu.go`):**
   Because `prefill-0-1` is mislabeled with `subGroupIndex=0`, `addTPUVariablesSubGroup` injects `prefill-0.prefill` into it and shifts subsequent subgroups left by 1.

3. **`tpuWorkerId` calculation bug (`pkg/utils/accelerators/tpu.go`):**
   `tpuWorkerId` should be set to 0 if workerIndex is 0. `workerIndex - 1 % subGroupSize` will result in -1 when `workerIndex=0`

**Environment**:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

YuleiF:
#### What type of PR is this?

/kind bug

This PR fixes a bug in subgroup index assignment and TPU environment variables injection when `subGroupSize = 1` (Issue #917).

**The Bug:**
When `subGroupSize` is set to `1` (e.g., in TPU workloads where each pod represents an independent 1-slice TPU endpoint), the subgroup index calculation incorrectly evaluated the conditional `(podCount-1)%subGroupSize == 0` as true (since any integer modulo 1 is 0). This caused LWS to assume the leader was an extra pod and shift all worker subgroup indices to the left by 1:
- Pod `prefill-0` (Leader, `workerIndex=0`) got `subGroupIndex = 0`
- Pod `prefill-0-1` (`workerIndex=1`) got `subGroupIndex = 0` (colliding with the leader)
- Pod `prefill-0-2` (`workerIndex=2`) got `subGroupIndex = 1`

This incorrect labeling cascaded into `addTPUVariablesSubGroup`, leading to wrong hostnames in `TPU_WORKER_HOSTNAMES` (e.g., `prefill-0-1` pointed to the leader instead of itself).

**The Fix:**
Modified `getSubGroupIndex` in `pkg/webhooks/pod_webhook.go` to ensure that when the total pod count (including the leader) is divisible by the subgroup size (`podCount % subGroupSize == 0`), pods are partitioned into equal-sized subgroups without index shifts.

Added unit tests to `pkg/webhooks/pod_webhook_test.go` and `pkg/utils/accelerators/tpu_test.go` to cover `subGroupSize = 1` configurations.

Fixes #917

None

```release-note
NONE

### DIFF
--- a/pkg/webhooks/pod_webhook.go
+++ b/pkg/webhooks/pod_webhook.go
@@ -247,7 +247,7 @@ func exclusiveAffinityApplied(pod corev1.Pod, topologyKey string) bool {
 }
 
 func getSubGroupIndex(podCount int, subGroupSize int, workerIndex int) string {
-	if (podCount-1)%subGroupSize == 0 {
+	if (podCount-1)%subGroupSize == 0 && podCount%subGroupSize != 0 {
 		// Leader is considered as extra pod, it is part of the first group
 		return fmt.Sprint((workerIndex - 1) / subGroupSize)
 	}
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -14,4 +14,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: us-central1-docker.pkg.dev/k8s-staging-images/lws/lws
-  newTag: release-0.9
+  newTag: main
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -122,6 +122,18 @@ rules:
 - apiGroups:
   - leaderworkerset.x-k8s.io
   resources:
+  - leaderworkersets
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - leaderworkerset.x-k8s.io
+  resources:
   - leaderworkersets/finalizers
   verbs:
   - update
@@ -133,22 +145,3 @@ rules:
   - get
   - patch
   - update
-- apiGroups:
-  - leaderworkerset.x-k8s.io
-  - leaderworkersetv1.x-k8s.io
-  resources:
-  - leaderworkersets
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
-  - leaderworkersetv1.x-k8s.io
-  resources:
-  - leaderworkersets/status
-  verbs:
-  - get
--- a/pkg/utils/accelerators/tpu.go
+++ b/pkg/utils/accelerators/tpu.go
@@ -125,8 +125,9 @@ func addTPUVariablesSubGroup(pod *corev1.Pod) error {
 	}
 	tpuWorkerId := (workerIndex) % subGroupSize
 
-	if pod.Annotations[LeaderRequestsTPUsAnnotationKey] != "true" {
+	if workerIndex != 0 && pod.Annotations[LeaderRequestsTPUsAnnotationKey] != "true" {
 		tpuWorkerId = (workerIndex - 1) % subGroupSize
+		subGroupIndex = (workerIndex - 1) / subGroupSize
 	}
 
 	tpuProcessPortInContainer, tpuProcessPort := podutils.GetEnvVarValueIfInContainer(container, TpuProcessPortName)
--- a/pkg/utils/accelerators/tpu_test.go
+++ b/pkg/utils/accelerators/tpu_test.go
@@ -357,6 +357,50 @@ func TestAddTPUVariablesSubGroup(t *testing.T) {
 		expectedError               bool
 	}{
 		{
+			name: "Leader requests TPU resources, leader pod",
+			pod: &corev1.Pod{
+				Spec: wrappers.MakeLeaderPodSpecWithTPUResource(),
+				ObjectMeta: v1.ObjectMeta{
+					Name:      "test-sample-0",
+					Namespace: "default",
+					Labels: map[string]string{
+						leaderworkerset.WorkerIndexLabelKey:   "0",
+						leaderworkerset.SubGroupIndexLabelKey: "0",
+					},
+					Annotations: map[string]string{
+						leaderworkerset.SubGroupSizeAnnotationKey: "1",
+					},
+				},
+			},
+			expectedTpuWorkerId:         "0",
+			expectedTpuWorkerHostNames:  "test-sample-0.default",
+			expectedTpuName:             "test-sample-0",
+			expectedTpuProcessAddresses: "test-sample-0.default:8476",
+			expectedTpuProcessPort:      "8476",
+		},
+		{
+			name: "Leader requests TPU resources, leader pod, size=2",
+			pod: &corev1.Pod{
+				Spec: wrappers.MakeLeaderPodSpecWithTPUResource(),
+				ObjectMeta: v1.ObjectMeta{
+					Name:      "test-sample-0",
+					Namespace: "default",
+					Labels: map[string]string{
+						leaderworkerset.WorkerIndexLabelKey:   "0",
+						leaderworkerset.SubGroupIndexLabelKey: "0",
+					},
+					Annotations: map[string]string{
+						leaderworkerset.SubGroupSizeAnnotationKey: "2",
+					},
+				},
+			},
+			expectedTpuWorkerId:         "0",
+			expectedTpuWorkerHostNames:  "test-sample-0.default,test-sample-0-1.default",
+			expectedTpuName:             "test-sample-0",
+			expectedTpuProcessAddresses: "test-sample-0.default:8476,test-sample-0-1.default:8476",
+			expectedTpuProcessPort:      "8476",
+		},
+		{
 			name: "Leader requests TPU resources",
 			pod: &corev1.Pod{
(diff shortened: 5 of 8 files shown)

### CODE REVIEW

**Comment by Edwinhr716** [comment]
@YuleiF can you do the EasyCLA authorization?

File: pkg/webhooks/pod_webhook.go [hunk 1]
```diff
@@ -247,7 +247,7 @@ func exclusiveAffinityApplied(pod corev1.Pod, topologyKey string) bool {
 }
 
 func getSubGroupIndex(podCount int, subGroupSize int, workerIndex int) string {
-	if (podCount-1)%subGroupSize == 0 {
+	if (podCount-1)%subGroupSize == 0 && podCount%subGroupSize != 0 {
```
yankay, line 250: HI @YuleiF 

Thanks for taking a look. I think the reason this was not caught is that the new `subGroupSize=1` tests only cover the `leaderRequestsTPUs=true` path.

That path still works because `addTPUVariablesSubGroup` applies the existing `start -= 1` / `end -= 1` compensation when the leader requests TPUs and the subgroup index is greater than 0. The uncovered case is `leaderRequestsTPUs=false + subGroupSize=1`: `TPU_WORKER_ID` is computed from the shifted TPU worker sequence with `(workerIndex-1)%subGroupSize`, but the hostname range is still computed directly from the canonical LWS subgroup label with `start := subGroupSize*subGroupIndex + 1`. After this change, worker-1 has `subGroupIndex=1`, so that formula starts at worker-2.

Could we add a regression test for `leaderRequestsTPUs=false + subGroupSize=1` and make the TPU hostname calculation use an effective TPU subgroup index for that path, for example `(workerIndex-1)/subGroupSize`?

YuleiF: I updated the code and test

yankay:
HI @YuleiF 

Thanks for the update! The new leaderRequestsTPUs=false + subGroupSize=1 coverage addresses my concern, and the fix looks consistent with the existing TPU subgroup offset logic.
One upgrade note: existing Pods will not be automatically recreated or mutated by this webhook change, so already-running subGroupSize=1 TPU workloads may need a recreate/rollout to pick up the corrected labels and TPU env vars.

/lgtm

YuleiF:
Hey @yankay , I noticed that when the leader pod does not have the annotation `leaderworkerset.sigs.k8s.io/leader-requests-tpus`, it will use 
```
		tpuWorkerId = (workerIndex - 1) % subGroupSize
		subGroupIndex = (workerIndex - 1) / subGroupSize
```
Consequently, `tpu_worker_id` evaluates to -1.

I have fixed this in the newest patch. Please re-review the changes. Thanks!

ahg-g:
> Hey @yankay , I noticed that when the leader pod does not have the annotation `leaderworkerset.sigs.k8s.io/leader-requests-tpus`, it will use
> 
> ```
> 		tpuWorkerId = (workerIndex - 1) % subGroupSize
> 		subGroupIndex = (workerIndex - 1) / subGroupSize
> ```
> 
> Consequently, `tpu_worker_id` evaluates to -1.

Thanks @YuleiF , this is not necessarily a problem since the leader pod is not a tpu worker in the first place. In fact, we should probably not set the tpu_worker_id env var at all.

> I have fixed this in the newest patch. Please re-review the changes. Thanks!

ahg-g:
I am not sure why the test continues to fail even though this got fixed in https://github.com/kubernetes-sigs/lws/pull/929

/retest all

ahg-g:
@YuleiF you need to rebase to pick up the vulnerability fix

YuleiF:
> > Hey @yankay , I noticed that when the leader pod does not have the annotation `leaderworkerset.sigs.k8s.io/leader-requests-tpus`, it will use
> > ```
> > 		tpuWorkerId = (workerIndex - 1) % subGroupSize
> > 		subGroupIndex = (workerIndex - 1) / subGroupSize
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Consequently, `tpu_worker_id` evaluates to -1.
> 
> Thanks @YuleiF , this is not necessarily a problem since the leader pod is not a tpu worker in the first place. In fact, we should probably not set the tpu_worker_id env var at all.
> 
> > I have fixed this in the newest patch. Please re-review the changes. Thanks!

Thanks for the reply @ahg-g . It's still a problem even when leader requires tpus, and `tpu_worker_id` will be -1. Somehow `leaderworkerset.sigs.k8s.io/leader-requests-tpus: true` is missing in leader and only added in the workers. I'm not sure if we should fix it by adding `leaderworkerset.sigs.k8s.io/leader-requests-tpus: true` to the leader instead.

YuleiF:
Hi @ahg-g, do you mind taking a look of this PR? Thanks!

Edwinhr716:
@YuleiF we only inject TPU environment variables if the pod requests TPU resources. So if `workerIndex == 0` and it reaches that line, it means that the leader pod is requesting TPU resources, no need to add a new annotation.

Edwinhr716:
> tpu_worker_id will be -1

I'm confused by this. Doesn't the unit test validate that TPU-Worker-ID is 0 if the leader requests resources?

yankay:
This CI failure depends on #937 being merged; once it lands, we can rerun the failed check.

PR #922, opened 2026-07-09
feat(disaggregatedset): per-role autoscaling via DisaggregatedSetRoleScaler

### LINKED ISSUE #849: [DisaggregatedSet] HPA support per role in DisaggregatedSet

hasB4K:
#### What type of PR is this?

/kind feature
/kind api-change

Adds KEP-849 (#910) implementation of per-role external scaling for DisaggregatedSet. A role opts in via scaling.mode: External; the DS controller then auto-creates a DisaggregatedSetRoleScaler named "<ds>-<role>" that exposes /scale for HPA, KEDA, or any /scale-aware controller. Replicas flow: autoscaler → scaler.spec.replicas → LWS.

New CRD DisaggregatedSetRoleScaler with the /scale subresource, aggregate status.selector across revisions, and controller ownerRef for GC. The executor resolves per-role targets from the scaler map; a no-shrink guard prevents the new-revision target from shrinking mid-rollout if HPA writes a smaller value. Webhook warns on External + spec.replicas > 1 and rejects External + spec.slices > 1 (alpha restriction).

Fixes #849 

This feature doesn't support slices > 1 for nows

```release-note
We add the CRD DisaggregatedSetRoleScaler when using External RoleScalingMode.
```


yankay:
Hi @hasB4K 

**A fresh scaler returns HTTP 500 from /scale**

Thank you for working on this. While testing the current head (22753ad) with a real HPA, I encountered what appears to be a blocking issue: https://github.com/yankay/lws/issues/15
KEP-849 uses spec.replicas: nil to mean that no autoscaler has written yet. However, Kubernetes requires the value at specReplicasPath to exist; otherwise, GET /scale returns an error:
https://github.com/kubernetes/kubernetes/pull/55168
https://github.com/kubernetes/kubernetes/issues/111781
The HPA consequently remains AbleToScale=False / FailedGetScale and cannot perform its initial write because it must read Scale first.
Would it make sense to initialize spec.replicas atomically during scaler creation—0 for a fresh role, or the target LWS’s desired spec.replicas for a Static-to-External transition? We may also need to adjust the KEP and WaitingForScaler semantics and add a real-apiserver integration test for this path.
Unless I am missing some context, this seems to block HPA/KEDA from attaching to a fresh scaler. Thanks for taking a look.

pkg/webhooks/disaggregatedset/disaggregatedset_webhook.go
```diff
@@ -45,32 +44,60 @@ var _ admission.Validator[*disaggv1.DisaggregatedSet] = &DisaggregatedSetWebhook
...
+		// almost certainly indicate confusion about which knob takes effect.
+		if role.Spec.Replicas != nil && *role.Spec.Replicas > 1 {
+			warnings = append(warnings, fmt.Sprintf(
+				"role %q sets scaling.mode: External and spec.replicas: %d — spec.replicas is ignored; drive replicas via DisaggregatedSetRoleScaler %q instead",
+				role.Name, *role.Spec.Replicas, obj.Name+"-"+role.Name))
+		}
 	}
 
-	return allErrs
+	// Alpha: External scaling and slices > 1 are incompatible. The scaler
```
Edwinhr716, line 96: @BenjaminBraunDev fyi

Edwinhr716:
/lgtm

Will defer to @yankay for the final approval

pkg/webhooks/disaggregatedset/disaggregatedset_webhook.go
BenjaminBraunDev: Interesting, is there a reason this can't be made compatible with slices?

BenjaminBraunDev: @hasB4K

hasB4K: @BenjaminBraunDev yes, I explain it there: https://github.com/kubernetes-sigs/lws/blob/7072b29/keps/849-DisaggregatedSet-HPA/README.md#interaction-with-disaggregatedset-slices

The main reason is that we need to design a proper solution.

yankay:
Hi @hasB4K,

Thank you for the update. I retested [985c7d4](https://github.com/kubernetes-sigs/lws/commit/985c7d4ae58bfa53ac20ccede6424fd9d6ad7c76) with a real HPA and noticed two remaining cases.

A fresh External role starts at zero, so vanilla HPA cannot bootstrap it even with `minReplicas: 1`:
https://github.com/yankay/lws/issues/16

This appears to expose a mismatch with [KEP-849](https://github.com/kubernetes-sigs/lws/pull/910), rather than require a new scale-from-zero feature. The KEP states that vanilla HPA bootstraps a fresh role from zero through `minReplicas`, but the real HPA remains `ScalingDisabled`. Could we update the implementation and KEP together, perhaps by starting fresh roles at `1` while preserving existing counts, including zero, for Static-to-External transitions?

There is also a group/Pod selector mismatch when `size > 1`:
https://github.com/yankay/lws/issues/17

Since this mainly affects Resource and Pods metrics rather than the expected KEDA external-metric path, I do not think it needs to block this PR. It could be documented or handled as a follow-up.

Would you mind taking another look at the first issue before merge? Thank you again :)

hasB4K:
Thanks @yankay - doing that today

hasB4K:
@yankay - I patched all the bugs you noticed:
* https://github.com/yankay/lws/issues/16 is fixed by https://github.com/kubernetes-sigs/lws/pull/922/changes/b18ffc1c8ee9599ea2533586d9eca046e35fdc1b
* https://github.com/yankay/lws/issues/17 is fixed by https://github.com/kubernetes-sigs/lws/pull/922/changes/b4be61beb34f42dd90072f0bf8d19a5208e11277
* `A fresh scaler returns HTTP 500 from /scale`: should also has been fixed (sorry about this one, it used to work, but I think I broke it after a refactor...)

I have been testing this in a pre-prod environment, and I think we are good to merge. Let me know if you need anything else. 😉

yankay:
Hi @hasB4K,
Thank you for addressing the reported issues. The runtime paths now look correct, and CI is green. I’m happy to approve this PR, with two non-blocking follow-ups:
* Could we add the missing Helm manager RBAC for disaggregatedsetrolescalers and /status? This is tracked in https://github.com/yankay/lws/issues/18.
* It would also be helpful to update the [`DisaggregatedSetRoleScalerSpec.Replicas` comment](https://github.com/kubernetes-sigs/lws/blob/b4be61beb34f42dd90072f0bf8d19a5208e11277/api/disaggregatedset/v1/disaggregatedsetrolescaler_types.go#L32-L35), which still says a fresh role is seeded at `0`, and generated docs that still describe the fresh seed as 0 instead of 1?
Thanks again for your work on this :)

/approve

PR #925, opened 2026-07-13
Add opt-in controller-driven rolling updates (rollout-via-delete)

alexcole:
Rolling updates are paced through the leader statefulset's rollingUpdate.maxUnavailable, an alpha field (MaxUnavailableStatefulSet) unavailable on managed clusters. Without it the statefulset controller recreates leaders strictly one at a time, gated on readiness - for multi-node inference groups where leader-ready means model-loaded, a rollout of N groups takes N cold starts in sequence.

With the leaderworkerset.sigs.k8s.io/rollout-via-delete: "true" annotation, the leader statefulset switches to OnDelete and the lws controller deletes stale leader pods itself, up to the group-level maxUnavailable budget; the statefulset controller recreates them in parallel. The existing partition math is unchanged - the partition moves into an annotation (OnDelete forbids the field) and deletion replaces the partition write as the executor.

Opt-in per object so the controller can be deployed as a no-op and enabled pool by pool; removing the annotation restores upstream statefulset-driven updates without disturbing pods or revisions.

Also resolve percentage budgets like Deployments do: reject only literal 0/0 at admission and floor the resolved maxUnavailable at 1, so maxUnavailable: 20% cannot stall a rollout at small replica counts (e.g. 3 replicas, or after a scale-down mid-rollout).

### DIFF
--- a/api/leaderworkerset/v1/leaderworkerset_types.go
+++ b/api/leaderworkerset/v1/leaderworkerset_types.go
@@ -96,6 +96,22 @@ const (
 	// Enables feature where the group will be restarted after pod failure if and only if
 	// all pods in the group are not pending
 	RecreateGroupAfterStartAnnotationKey string = "leaderworkerset.sigs.k8s.io/experimental-recreate-group-after-start"
+
+	// RolloutViaDelete opts a LeaderWorkerSet into controller-driven rolling updates:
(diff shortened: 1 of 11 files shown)


alexcole:
oops meant to make this on my fork only. not ready for public viewing yet!

but lmk if ya'll are interested in this upstream

Issue #529, opened 2025-05-19
Bug: with older CRD , subgroup missing optional field causing panic


panpan0000:
**What happened**:
with older CRD in system(which tend to missing when helm upgrade)
apply a LWS without subGroupPolicy as below

```
piVersion: leaderworkerset.x-k8s.io/v1
kind: LeaderWorkerSet
metadata:
  name: ...
spec:
  leaderWorkerTemplate:
    leaderTemplate:
     ....
    workerTemplate:
      ...
    subGroupPolicy:
      #subGroupPolicyType: LeaderWorker  ######## without this line
      subGroupSize: 2
```

```
# kubectl  apply -f lws.yaml

Error from server: error when creating "lws.yaml": admission webhook "vleaderworkerset.kb.io" denied the request: panic: runtime error: invalid memory address or nil pointer dereference [recovered]

```

```

[...]

PR #530, opened 2025-05-19
fix: check optional field not nil before reference

panpan0000:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it

1) check optional field not nil which may come from older CRD
2) modify the CRD file in charts sub-folder ( copy that part from `config/crd/bases/leaderworkerset.x-k8s.io_leaderworkersets.yaml` generated by `make artifacts`)

NOTE: the "charts/" folder seems being manually maintained,  I cannot directly override the files under `charts` folder , with kustomize generated `config/`  files.

#### Which issue(s) this PR fixes

Fixes #529 

#### Does this PR introduce a user-facing change?

```release-note

```

PR #932, opened 2026-07-16
fix: prevent nil pointer panics in webhook and controller

### LINKED ISSUE #931: Fix nil-pointer panics in LeaderWorkerSet webhook and controller under nil conditions
Ran into a nil pointer panic when deploying a LeaderWorkerSet where replicas wasn't explicitly set in the spec yaml. It looks like the webhook tries to do some math on the replica pointer without checking if it's nil first. Same thing happens when trying to parse spec.rolloutStrategy.rollingUpdateConfiguration if that isn't defined either. We should add proper nil checks to fallback to defaults.

Dasmat13:
Fixes #931

Added safety checks for optional spec fields. If replicas or rollingUpdateConfiguration (like maxSurge / maxUnavailable / partition) are nil, they will default safely instead of crashing the control plane.


Dasmat13:
Thanks for the nil-safety fixes. While checking the webhook paths, I found one remaining optional-field panic that seems compatible with this PR's scope:

`validateUpdateSubGroupPolicy` is called whenever `subGroupPolicy` is non-nil, but it immediately dereferences the separately optional `SubGroupSize`. Thus an otherwise valid LWS with `subGroupPolicy: {}` panics in the validating webhook. An explicit `subGroupSize: 0` then reaches the modulo operations and panics with divide-by-zero.

Could this PR also:

1. Return a `field.Required` error when `SubGroupSize` is nil.
2. Return a normal invalid-field error for values below one before any modulo operation.
3. Add webhook regression tests for both `{}` and `subGroupSize: 0`.

This is related to closed #529 / #530, which guarded the distinct optional `subGroupPolicy.type` field, but the current `SubGroupSize` path remains unguarded.

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -140,6 +140,10 @@ type LeaderWorkerSetSpec struct {
 	// networkConfig defines the network configuration of the group
 	// +optional
 	NetworkConfig *NetworkConfig `json:"networkConfig,omitempty"`
+
+	// gangScheduling defines the gang scheduling policy of the group.
+	// +optional
+	GangScheduling *GangSchedulingPolicy `json:"gangScheduling,omitempty"`
```
ahg-g: This PR seems to be mixing a few unrelated things

Dasmat13:
Hi @hasB4K @kerthcet, quick ping — latest fixes are in and all CI is green. Would appreciate a review when you get a chance. Thanks

ahg-g (commented):
In general, most of these fields should be validated to have a value in the webhook, we should not need to check and default again in the controllers

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -286,11 +286,18 @@ func SetupIndexes(indexer client.FieldIndexer) error {
 func (r *LeaderWorkerSetReconciler) rollingUpdateParameters(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, sts *appsv1.StatefulSet, revisionKey string, leaderWorkerSetUpdated bool) (stsPartition int32, replicas int32, err error) {
 	log := ctrl.LoggerFrom(ctx).WithValues("leaderworkerset", klog.KObj(lws))
 	ctx = ctrl.LoggerInto(ctx, log)
-	lwsReplicas := *lws.Spec.Replicas
+	lwsReplicas := int32(1)
+	if lws.Spec.Replicas != nil {
```
ahg-g: As suggested before, Replicas should never be nil here, it should be defaulted when the object is admitted. The same goes in all other similar cases, all parameters must have a default and so should never be nil.

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -128,54 +128,54 @@ func (r *LeaderWorkerSetWebhook) generalValidate(lws *v1.LeaderWorkerSet) field.
 	ValidateName := apivalidation.NameIsDNS1035Label
 	allErrs := apivalidation.ValidateObjectMeta(&lws.ObjectMeta, true, apivalidation.ValidateNameFunc(ValidateName), field.NewPath("metadata"))
 	// Ensure replicas and groups number are valid
+	replicas := int32(1)
```
ahg-g: Defaulting should be done in `Default`, this has no effect on the value we actually set and so this validation is misleading and can actually cause problems.

Dasmat13:
Thanks, fixed these now and rebased on latest main. Added update tests too, all tests are passing. PTAL when you get time

ahg-g:
/approve
/lgtm

thanks!

PR #933, opened 2026-07-17
Populate DisaggregatedSet .status from owned LWS objects

### LINKED ISSUE #868: [DisaggregatedSet] .status is never populated by the controller
/kind bug

**What happened**:

The `DisaggregatedSet` controller never writes `.status`. After create, rolling update, scale, and add/remove role, it stays empty:

```
$ kubectl get disaggregatedset ds-basic -o json | jq .status
{}
```

`observedGeneration` is also never set, so clients can't tell whether the controller has seen the latest `.spec`.

Each reconcile should write at minimum:

- `status.observedGeneration`
- `status.replicas` / `readyReplicas` / `updatedReplicas`
- `status.conditions` (`Available` / `Progressing` / `Degraded`)

The CRD already declares the subresource and full schema in `api/disaggregatedset/v1/disaggregatedset_types.go` — this is a controller-side gap.

1. Apply any valid 2-role `DisaggregatedSet`.
2. Wait until pods are `Running`.
3. `kubectl get disaggregatedset <name> -o json | jq .status` → `{}`.

Blocks `additionalPrinterColumns` on the CRD, and leaves "one role unschedulable" silently invisible to users.

**Environment**:

- Kubernetes version (`kubectl version`): v1.35.0 (kind v0.31.0, single-node)
- LWS version (`git describe --tags --dirty --always`): main @ commit `17ee230` (+ 2/N unified controller-manager merge)
- Cloud provider or hardware configuration: local kind cluster (bare-metal host)
- OS (`cat /etc/os-release`): Ubuntu 24.04.4 LTS (Noble Numbat)
- Kernel (`uname -a`): Linux 6.17.0-29-generic x86_64
- Install tools: kind + `make image-build` + `kubectl apply`
- Others: DisaggregatedSet CRD + controller co-located in `lws-controller-manager`

BenAyedMedAla:
/kind bug
/kind api-change

The `DisaggregatedSet` controller reconciles LWS objects for each role but never wrote `.status`: after create, rolling update, scale, or add/remove role, `kubectl get disaggregatedset <name> -o json | jq .status` returns `{}`. This blocks `additionalPrinterColumns` on the CRD and leaves partial/failed rollouts silently invisible to users.

Per review discussion, this PR is scoped to the foundational status reporting only:

- `status.observedGeneration`, mirroring the existing pattern in `LeaderWorkerSetStatus`.
- `status.roleStatuses[]`: per-role `replicas`/`readyReplicas`/`updatedReplicas`, aggregated across all slices and revisions of that role's owned LWS objects. Only LWS at the current target revision contribute to `updatedReplicas` (a draining old-revision LWS is not "updated"). A role removed from `spec.roles` has no entry here, even while its old LWS objects are still draining (documented on the `RoleStatuses` field).
- LWS listing is filtered by controller-owner UID (`metav1.IsControlledBy`), not just name/role labels, so an unrelated LWS that happens to carry matching labels can't inflate the counters.

The `Available`/`Progressing` condition model (desired targets, slices, zero-replica behavior, condition interaction, failure semantics) is being split into a focused follow-up PR, per @yankay's suggestion below — I have that work ready on a branch and will open it once this one lands.

`status.observedGeneration` is new API surface, so I regenerated the CRD manifest and API reference docs via `controller-gen` and `genref` at the versions pinned in the Makefile. The resulting diffs in `config/`, `charts/`, and `site/` are minimal.

Fixes #868 (partially — the condition-model half is tracked in the follow-up PR)

- Added a `LeaderWorkerSetWrapper.UpdatedReplicas()` test helper since it didn't exist yet.
- Fixed several existing test fixtures that built LWS `OwnerReference` literals without `Controller: true` (unlike the real `Create()` path), since the new owner-UID filtering depends on it.

```release-note
DisaggregatedSet: the controller now populates `status.observedGeneration` and per-role `status.roleStatuses` on every reconcile, aggregated from the LWS objects it owns. Previously `.status` was always empty. (Available/Progressing conditions land in a follow-up PR.)
```


pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -114,8 +116,121 @@ func (r *DisaggregatedSetReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+				changed = true
+			}
+			continue
+		}
+		if cond.Status == metav1.ConditionTrue {
+			disaggregatedSet.Status.Conditions[i].Status = metav1.ConditionFalse
+			disaggregatedSet.Status.Conditions[i].LastTransitionTime = newCondition.LastTransitionTime
+			disaggregatedSet.Status.Conditions[i].ObservedGeneration = newCondition.ObservedGeneration
+			changed = true
+		}
```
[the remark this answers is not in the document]
BenAyedMedAla: Good catch, thanks. Fixed in 2bcd8b3: `setDisaggregatedSetCondition` now only refreshes `LastTransitionTime` when a condition's `Status` actually flips true/false; an `ObservedGeneration`-only update (or a no-op) leaves the existing `LastTransitionTime` alone. Added `TestSetDisaggregatedSetCondition_ObservedGenerationOnlyDoesNotResetTransitionTime` and `TestSetDisaggregatedSetCondition_StatusFlipUpdatesTransitionTime` to cover both cases directly.

yankay:
HI @BenAyedMedAla 

Thanks for addressing this status gap—this is a valuable observability improvement. Would you be open to splitting it into two PRs?
This PR could focus on observedGeneration and per-role replica counters aggregated from owned LWS objects. A follow-up could define and implement Available/Progressing, including desired targets, slices, zero-replica behavior, condition interaction, and failure semantics.
This would let the foundational status reporting land first while giving the condition model more focused design discussion and review. What do you think? :)

yankay:
Thanks for the status aggregation work. I found two points to address or clarify before merge:
* LWSManager.List filters by labels but not controller ownership, so an unrelated LWS with matching labels could affect the counters. Could we also filter by the DisaggregatedSet owner UID?
* updateStatus only visits current spec.roles. A removed role disappears from roleStatuses even while its old LWS is still draining. Could we define the intended behavior and add a role-removal test?

The first is a correctness issue; the second mainly needs an explicit contract. Neither requires redesigning the overall aggregation approach. :)

BenAyedMedAla:
Thanks for the review, @yankay! Fixed both in ffa8397:

1. **Owner UID filtering**: `LeaderWorkerSetManager.List` (and `GetRevisionRolesList`, which calls it) now takes the `DisaggregatedSet` object and filters results with `metav1.IsControlledBy`, matching the controller owner reference every LWS this manager creates already carries. An unrelated LWS with matching name/role labels but no matching owner UID is now excluded. I also found that several existing test fixtures were building `OwnerReference` literals without `Controller: true` (unlike the real `Create()` path, which always sets it) — fixed those too, and added an explicit "unowned LWS with matching labels" case to `TestManagerListSliceBucketing` to pin the exclusion directly.

2. **Role-removal contract**: documented explicitly on `RoleStatuses` and `updateStatus` — a role removed from `spec.roles` has no entry in `status.roleStatuses` on the very next reconcile, even though its old LWS objects aren't deleted by this reconcile (nothing currently drains/removes a removed role's leftover LWS — that's a pre-existing lifecycle gap, not something this PR takes on). Added `TestStatusDropsRemovedRoleEvenWhileItsLWSStillDrains` to pin that contract, including confirming the old LWS is still present so the test doesn't accidentally pass because of unrelated cleanup.

Let me know if you'd like the role-removal behavior to go the other way (e.g., keep reporting a draining role until its LWS are gone) — happy to adjust, I don't have a strong opinion beyond matching the "RoleStatuses mirrors spec.roles" framing already in the doc comment.

BenAyedMedAla:
Sounds good, thanks for the suggestion — split done. This PR is now trimmed to just `observedGeneration` + `roleStatuses` (pushed in 7476530): removed the `DisaggregatedSetConditionType`/`Available`/`Progressing` constants and the condition-setting half of `updateStatus`, reverted the `Conditions` doc comment back to its original text since no condition type is implemented here, and updated the PR description to match.

I already have the `Available`/`Progressing` work (including the `LastTransitionTime` fix Copilot flagged, and its tests) sitting on a branch, carried over as-is from before the split. I'll open it as a follow-up PR once this one merges, so its diff is clean against the then-current `main` instead of overlapping with this one.

pkg/controllers/disaggregatedset/lws_manager.go
```diff
@@ -142,26 +142,32 @@ func (manager *LeaderWorkerSetManager) Get(ctx context.Context, namespace, name
...
+// List returns the LWS controlled by disaggregatedSet, filtered to one slice. A
+// slice < 0 matches all slices. Slice 0 also matches legacy (pre-slices) LWS that
+// carry no slice label, so they are reconciled as slice 0. Slice filtering is
+// client-side because "slice label == 0 OR absent" cannot be expressed as a label
+// selector. Results are additionally filtered by controller-owner UID (every LWS
+// this manager creates is owned by its DisaggregatedSet), so an unrelated LWS that
+// happens to carry matching name/role labels — e.g. hand-crafted, or left over from
+// a same-named DisaggregatedSet that was deleted and recreated — cannot be
+// mistaken for one of this DisaggregatedSet's own replicas.
+func (manager *LeaderWorkerSetManager) List(ctx context.Context, disaggregatedSet *disaggregatedsetv1.DisaggregatedSet, slice int, role string) ([]*leaderworkersetv1.LeaderWorkerSet, error) {
```
hasB4K, line 154: Much better! Thank you!

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -134,14 +135,72 @@ func (r *DisaggregatedSetReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 
-	return result, errors.Join(errs...)
+	// Status reflects the state observed above regardless of per-slice errors, so
+	// a role that failed to reconcile is still visible to clients instead of being
+	// silently left out of .status.
+	if statusErr := r.updateStatus(ctx, disaggregatedSet, roleNames, revision); statusErr != nil {
+		return ctrl.Result{}, errors.Join(reconcileErr, fmt.Errorf("failed to update status: %w", statusErr))
+	}
+
+	return result, reconcileErr
```
hasB4K, line 153: This is perfect! Your patch is self-contained and very clean.

hasB4K (commented):
Thank you for this PR 🙂, it fixes an oversight of mine in the DisaggregatedSet implementation, and clearly, the fact that we don't have statuses is clearly a bug! 
/lgtm

I'll let @yankay do the approval!

hasB4K (approved):

BenAyedMedAla:
Gentle bump — this had an LGTM from @hasB4K last week, just needs `/approve` from one of the OWNERS approvers (ahg-g / kerthcet / edwinhr716 / yankay) to merge.

yankay:
HI @BenAyedMedAla 
Thanks for addressing the earlier feedback and keeping this PR focused. The status aggregation looks good to me.
One rare, non-blocking edge case remains: GetForRole/Scale could still mutate a foreign LWS occupying the generated name during rapid same-name DS recreation before GC. Could we track this as a follow-up? :)
/approve

BenAyedMedAla:
Filed as #981 — thanks for catching it!

BenAyedMedAla:
(Following up on my terse comment above for clarity) @yankay — the `GetForRole`/`Scale` foreign-LWS edge case you flagged when approving this PR is now tracked at #981, so it doesn't get lost. Not fixing it here since you'd noted it's rare and non-blocking — just wanted a place to point back to.

Issue #715, opened 2025-12-04
In-place group restart


ahg-g:
**What would you like to be added**:

in-place restart of failure mode instead of deleting all the pods; we can learn from the jobset proposal: https://github.com/kubernetes-sigs/jobset/tree/main/keps/467-InPlaceRestart

**Why is this needed**:

Faster failure handling for large scale setups like wide-ep; 

**Completion requirements**:

This enhancement requires the following artifacts:

- [x] Design doc
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

ahg-g:
@GiuseppeTT @Edwinhr716 @imreddy13

bhks:
@ahg-g ,

I believe one difference to the Jobset API would be how do we handle the RayCluster based model deployment since for multi node model deployment we use ray cluster. 

1. Once the Ray head pod restarts each worker pods need to restart as well while other way around works fine. Each worker of ray can use the same address to restart and join the cluster.
2. There needs to be a sequence understanding between head pod vs the worker pod restart.
3. The reliability of ray cluster depends on the in-memory state rather than an etcd or similar DB. And that creates challenge for us on how to handle this in place restart.
4. Workers restart should just work fine but head pod restart would cause us to restart the whole cluster again. 

My understanding is that we would need to restart each replica as a whole in our case where we have multi node setup and each replica have one ray head and couple of workers.

ahg-g:
/close

I don't think this work is necessary anymore, it adds significant complexity and at the same time the value is marginal compared to recreating the pods.

PR #936, opened 2026-07-23
feat: implement KEP-715 In-Place Group Restart

Dasmat13:
# PR Implementation Summary: KEP-715 In-Place Group Restart

This PR implements the in-place group restart feature natively in LWS using a durable state machine, robust concurrency controls, and a safe Downward API communication channel.

## 1. Controller to Pod Signaling (Downward API)
The LWS controller runs externally to the Pods. To safely signal a group restart, it avoids direct filesystem access entirely.
Instead, it patches the following annotations on every Pod in the group:
- `desired-restart-generation`
- `barrier-open`

Each Pod annotation update is protected by optimistic concurrency. Group-wide fan-out is idempotent and may converge across multiple reconciliation cycles. The Kubelet eventually projects each new annotation set using an atomic Downward API volume refresh into the Pods at `/var/run/lws-api`. This mount purposefully uses no `subPath` so that updates propagate correctly. The agent containers read from this mount, repeatedly reopening the projected paths to detect the atomic `..data` symlink switch, ensuring a Kubernetes-native signaling pathway.

## 2. Pod Injection (Leader and Workers)
Revision-aware leader/worker construction injects the three necessary helper init-containers (`marker`, `agent`, `barrier`) into **both** the Leader Pod Template and the Worker Pod Template. 
This ensures that every member of the group is capable of triggering and fulfilling an in-place container restart. `RestartAllContainers` is a per-pod kubelet action, so local agent presence across all members is strictly required and implemented.

**Credential Isolation:** The helper containers do not use the Kubernetes API. Their service-account credential path is masked with a private `emptyDir`, preventing automatic service-account token injection without disabling credentials for user containers.

## 3. Four-File Transaction Protocol
The internal helper containers (`marker` and `agent`) coordinate state across restarts using a durable 4-file protocol in a local `emptyDir` (`/var/run/lws-state`):
1. `pod-lifecycle-count`
2. `accepted-lifecycle-count`
3. `requested-generation`
4. `observed-generation`

All writes to this local state use temporary-file-plus-rename semantics to avoid partial reads when the marker, agent, or barrier consume the values. The completion proof matches the exact generation. Generation is strictly monotonic because:
- Only an `Idle` → `Quiescing` compare-and-set can create an attempt.
- Only the leader reconciliation increments `DesiredGeneration`.
- The current generation is preserved in durable `Idle` state and is never reset after successful recovery.

## 4. API Safety & Validation (Webhook)
The Validating LWS webhook strictly locks down this feature:
- Enforces the hard K8s API limit of **20 RestartPolicyRules per container** (correctly calculating the sum of workload rules + injected trigger rules for each individual container independently). *Note: This counts rule objects, not individual exit-code values.*
- Validates the Kubernetes API Server version is >= 1.36.
- Prevents end-user workloads from using `exit code 88`, reserving it strictly for the injected `lws-restart-agent`.
- LWS InPlaceGroupRestart feature gate must be enabled.
- StartupPolicy must be LeaderCreated.
- RecoveryTimeout must be positive.
- MaxAttempts and Window must be valid.
- Duplicate role/container triggers are rejected.
- Conflicting helper volume or mount names are rejected.
- Trigger containers must exist in the selected leader or worker role.
- Existing conflicting `restartPolicyRules` must be rejected rather than overwritten (an existing container-level restart policy is preserved; when absent, injection copies the effective Pod restart policy).

*Note on abrupt termination risk*: Because this feature uses `RestartAllContainers`, `preStop` hooks are not executed and `terminationGracePeriodSeconds` is not respected. Applications and sidecars must tolerate abrupt termination. Pod UID, IP address, network namespace, sandbox, attached devices, `emptyDir` contents, and mounted volumes remain preserved.

## 5. Escalation & Failure Shielding
The state machine accurately masks expected terminations during active restart phases. 
**Until every Pod's handled counts are durably updated:**
- The state machine absorbs restart deltas belonging to the current generation.

**Escalation triggers (which immediately force full group recreation):**
1. **Unrecoverable exit:** An additional independent failure on a Pod *after* it acknowledges the active attempt.
2. **RecoveryTimeout expiry:** The `RecoveryTimeout` elapses before all pods acknowledge and become ready. This also serves as a runtime fallback for unsupported kubelets that do not support in-place restarts despite the API server version.
3. **MaxAttempts Exceeded:** The configured `MaxAttempts` is exceeded within the `Window`.
4. **Topology Invalidation:** The group membership or expected size changes, leader UID changes or leader is deleted, controller revision changes, a worker no longer belongs to the current leader, agent protocol version mismatch, stale group-unique hash, or a participant is missing the expected injected helper set.

## 6. Controller-visible Acknowledgements
Local transaction files are consumed only by the injected helpers. The LWS controller never reads Pod files. It derives acknowledgements using API-visible evidence.

During `WaitingForAcknowledgements`:
- Pod UID preservation.
- For the triggering Pod: Marker init-container restart count advanced beyond the *previously handled* marker count.
- For peers: Marker init-container restart count advanced beyond the *baseline captured during Signaling*.
- Current revision and leader ownership still match.
- (Supplementary check: `PodRestartInPlace` is not active.)

During `WaitingForReadiness`:
- Barrier opened.
- Eventual Pod readiness convergence.
- Handled counts patched successfully.

## 7. Worker-to-Leader Initiation
The worker reconciliation path detects a matching local failure and uses a compare-and-set update on the Leader Pod to initiate recovery. This begins the recovery cycle for worker-triggered failures.

**Ray sequencing scope:** The initial implementation coordinates the entire leader-worker group through one restart generation and releases all members through the common group barrier. It does not yet provide role-dependent startup stages, such as waiting for the Ray head application to become Ready before releasing worker applications. Workloads whose workers start concurrently must retry connecting to the head. Role-aware ordered barriers may be added in a future enhancement.


Dasmat13:
Hi @yankay, this PR implements KEP-715 In-Place Group Restart for LWS. Since you've guided me on previous LWS contributions and are familiar with this area, I'd really appreciate your review whenever you have time. Thank you!

Dasmat13:
Hi @yankay 
friendly ping to see if you have any feedback or thoughts on this KEP-715 implementation. Ready for review.

Dasmat13:
Hi @ardaguclu @kerthcet @ahg-g, quick ping on this PR. CI is green, would appreciate a review when you get a chance. Thanks

ahg-g:
Thanks @Dasmat13  for working on this, we learned from the JobSet that in-place restart complexity doesn't justify the value add. I left a comment on https://github.com/kubernetes-sigs/lws/issues/715; I don't think this work feature is necessary, so I am closing this PR for now, but please let us know if you think otherwise .

Dasmat13:
Got it, makes sense
Thanks @ahg-g for the context and for taking a look.

PR #941, opened 2026-07-28
Helm: add priorityClassName parameter

EladDolev:
#### What type of PR is this?

/kind feature

adds the ability to set priorityClassName with an empty default.
very useful for controllers, especially if you want to run a single replica.

```release-note
Helm: adds the ability of setting priorityClassName
```


ahg-g:
@EladDolev I think you need to rebase to get Trivy to pass

PR #946, opened 2026-07-29
fix: preserve missing-node reconcile errors

ShiroKSH:
## Summary

- Preserve the API error when the node hosting an exclusive-placement leader is unavailable.
- Prevent creation of a worker StatefulSet with an empty topology node selector.

## Root cause

`topologyValueFromPod` converted a missing Node error to `nil`, so reconciliation applied an empty topology value before creating the worker StatefulSet. Existing worker StatefulSets are not updated by this path.

## Validation

- `go test ./pkg/controllers -run '^TestSetNodeSelectorForWorkerPodsReturnsNotFoundWhenLeaderNodeIsMissing$' -count=1`
- `go test ./pkg/controllers -count=1`
- `go test ./api/... ./pkg/... ./cmd/... -count=1`
- `go vet ./...`

### DIFF
--- a/pkg/controllers/pod_controller.go
+++ b/pkg/controllers/pod_controller.go
@@ -313,18 +313,13 @@ func (r *PodReconciler) setNodeSelectorForWorkerPods(ctx context.Context, pod *c
 }
 
 func (r *PodReconciler) topologyValueFromPod(ctx context.Context, pod *corev1.Pod, topologyKey string) (string, error) {
-	log := ctrl.LoggerFrom(ctx)
-
 	nodeName := pod.Spec.NodeName
 	ns := pod.Namespace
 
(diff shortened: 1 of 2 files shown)


Dasmat13:
lgtm

looks good returning the err here makes sense so we dont end up creating worker sts with an empty topology selector when node object isn't fetched yet.

yankay:
HI @ShiroKSH 
Thanks for the fix. Could we add an integration test covering both the missing-Node case and recovery? The current test only covers the helper. :)

ShiroKSH:
> HI @ShiroKSH Thanks for the fix. Could we add an integration test covering both the missing-Node case and recovery? The current test only covers the helper. :)

np

Issue #948, opened 2026-07-29
Support External role scaling for multi-slice DisaggregatedSets


BenjaminBraunDev:
**What would you like to be added:**

Support for combining `spec.slices > 1` with `scaling.mode: External` on DisaggregatedSet roles. Today the webhook rejects this combination, so slices (KEP-846) and per-role autoscaling (KEP-849, implemented in #922) are mutually exclusive. The KEP-849 alpha deliberately scoped autoscaling to single-slice sets and deferred the multi-slice semantics to a follow-up KEP: what the scaler value means across slices, what the scale selector should match, and how the rolling update no-shrink guard behaves when slices roll independently.

**Why is this needed:**

Slices exist to map complete P/D topologies onto accelerator domains, and those are exactly the deployments that want to autoscale roles like prefill against traffic. With the restriction in place, users have to choose between slice-based deployment and autoscaling.

/kind feature

PR #949, opened 2026-07-29
[docs] KEP 948 external scaling slices support

BenjaminBraunDev:
#### What type of PR is this?

/kind documentation

Adds KEP-948, a follow-up to KEP-849 defining how External role scaling composes with spec.slices > 1. The scaler value becomes an aggregate split deterministically across slices, scaler status keeps its current shape, and the no-shrink guard applies per slice. This lifts the alpha restriction that makes slices and per-role autoscaling mutually exclusive.

Part of #948

```release-note
NONE
```

### DIFF
--- a/keps/948-disaggregatedset-slices-external-scaling/README.md
+++ b/keps/948-disaggregatedset-slices-external-scaling/README.md
@@ -1,31 +1,18 @@
 # KEP-948: DisaggregatedSet External Scaling for Multi-Slice Sets
 
-<!--
-This KEP lifts the KEP-849 alpha restriction that a DisaggregatedSet with any
-scaling.mode: External role must have spec.slices == 1, by defining aggregate
-semantics for DisaggregatedSetRoleScaler across slices.
--->
-
 <!-- toc -->
 - [Summary](#summary)
 - [Motivation](#motivation)
   - [Goals](#goals)
   - [Non-Goals](#non-goals)
 - [Proposal](#proposal)
-  - [User Stories](#user-stories)
-    - [Story 1: One HPA per role across a rack-partitioned fleet](#story-1-one-hpa-per-role-across-a-rack-partitioned-fleet)
-    - [Story 2: Mixed static and external roles in a multi-slice set](#story-2-mixed-static-and-external-roles-in-a-multi-slice-set)
-  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
   - [Risks and Mitigations](#risks-and-mitigations)
 - [Design Details](#design-details)
-  - [Replica Scope: Aggregate](#replica-scope-aggregate)
-  - [Distribution Across Slices](#distribution-across-slices)
-  - [Example: Scaling Under Uneven Slice Load](#example-scaling-under-uneven-slice-load)
-  - [Where Resolution Happens](#where-resolution-happens)
-  - [Scaler Status Across Slices](#scaler-status-across-slices)
-  - [No-Shrink Guard Per Slice](#no-shrink-guard-per-slice)
+  - [Aggregate Scope and Distribution](#aggregate-scope-and-distribution)
+  - [Per-Slice Resolution and the No-Shrink Guard](#per-slice-resolution-and-the-no-shrink-guard)
   - [Seeding](#seeding)
   - [Changing the Slice Count](#changing-the-slice-count)
+  - [Example: Scaling Under Uneven Slice Load](#example-scaling-under-uneven-slice-load)
   - [Validation Changes](#validation-changes)
   - [Test Plan](#test-plan)
     - [Unit tests](#unit-tests)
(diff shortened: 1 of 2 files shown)


hasB4K:
Hey there,

Thank you for this PR! A few points:

* This PR change the semantic of a slice, especially with uneven load, it's okay, but it's no longer "one slice is replicated exactly n-times"
* A slice usually has a fixed maximum capacity... It feels that we should scale up if we reach it?
* So we can have a prefill at 0 on one slice? and a decode at 2 on the same slice? How is this handled on the planner side? 

With an example of 10 slices: 3 prefill, and 10 decode:
```
   Slices    Prefill    Decode
  ━━━━━━━━  ━━━━━━━━━  ━━━━━━━━
   0–2             1         1
  ────────  ─────────  ────────
   3–9             0         1

  So you get:

  - 3 complete prefill/decode slices.
  - 7 decode-only slices.
```

During a rollout, all those slices will get rollout at the same time, creating an interruption of service for slice 3 to 9 (so 7 Decode at once).

I think we need a mechanism to scale up and scale down the slices themselves? Otherwise it can be dangerous. 


------
I think we need to have a sliceBudget mechanism with each prefill and decode having his own "how budget I use"

hasB4K:
So I expressed this idea with you online, we should have a `sliceBudget` field. The sliceCount field will be deprecated in the next next release.

Each role should have a `budgetUsedPerReplicas` (or a better name).

We should keep your algorithm to smooth over slices. But the budget mechanism will let you know how many number of slices you'll need.

A 3P10D, with sliceBudget at 72, and a budgetUsedPerReplicas for Prefill at 8 and for Decode at 16 would be:
* slice 0: 1P4D
* slice 1: 1P3D
* slice 2: 1P3D

And yes. The budget is the number of GPUs you want to use in the end.
Using sliceCount would be incompatible with sliceBudget.
When you use sliceBudget, the number of replicas for each role (when you don't use autoscaling), mean the literal number of replicas that we will have splitted across all slices. 

Thank you again to have listened to my concerns/ideas @BenjaminBraunDev

Edwinhr716:
> The budget is the number of GPUs you want to use in the end.
I don't agree with having the budget be the number of GPUs, as there is no guaranteed mapping between slices and number of GPUs

hasB4K:
> > The budget is the number of GPUs you want to use in the end
>
> I don't agree with having the budget be the number of GPUs, as there is no guaranteed mapping between slices and number of GPUs

What do you mean?

@Edwinhr716, `sliceBudget` would be the max number of GPUs/TPUs or whatever that you want to have for one slice.

Edwinhr716:
Right, but the number of GPUs / TPUs a slice can have depends on how many chips each role within the slice is requesting. So will DS now check how many accelerators each slice is requesting by iterating over every pod that is created? Why have that level of granularity?

hasB4K:
> Right, but the number of GPUs / TPUs a slice can have depends on how many chips each role within the slice is requesting. So will DS now check how many accelerators each slice is requesting by iterating over every pod that is created? Why have that level of granularity?

I’m not proposing that DS inspect running pods or derive accelerator usage. `budgetUsedPerReplica` would declaratively specify the total packing cost of one LWS group (for prefill or decode), and `sliceBudget` would use the same abstract unit. DS would only use those values to calculate the required slice count and distribute replicas; the scheduler remains authoritative for actual resource availability. You could have a `sliceBudget` of 32 if you want for instance...

One of the goal of slices is replicated DisaggregatedSet that are identical (or almost identical). Using a `sliceBudget` and `budgetUsedPerReplica` is just here to help implement a budget mechanism to know when we should create a new slice, or delete a new slice. The current approach can have way too many "degenerated" slice like a 0p1d described [here](https://github.com/kubernetes-sigs/lws/pull/949#issuecomment-5182037155).

Basically, when scaling prefill and decode, you want to uniformize as much as you can Prefill and Decode across different slices.

Of course, like shown in the above example, you can have non uniformized slice, like this:

```
A 3P10D, with sliceBudget at 72, and a budgetUsedPerReplicas for Prefill at 8 and for Decode at 16 would be:
*  slice 0: 1P4D
*  slice 1: 1P3D
*  slice 2: 1P3D

Then scale up of 3 Decode could be:
*  sice 0: 1P4D
*  slice 1: 1P4D
*  slice 2: 1P4D
*  slice 3: 0P1D    # this one is fine since it will always be "one" slice that can be like this
```

This example has one partial slice instead of seven. It does not guarantee that every workload shape has only one partial slice, but it lets the controller minimize partial slices while respecting a declared logical capacity.

What I'm trying to express, is that, without a budget mechanism, the HPA feature + slices will be broken. If we want to keep one slice to be "replicated DisaggregatedSet across several domain, like NVL72" - then, you need to have a mechanism that let you know how much you can have per slice (without inspecting pods).

Also, if you want to duplicate your model to as much slices that you can for resiliency, nothing stop you by setting `sliceBudget` to 32 for instance.

Happy to ear any alternatives, but right now, I don't see any. The problem that I'm describing is the reason why I didn't implement HPA for slices in the previous HPA KEP :'(

ahg-g:
should we close this?

BenjaminBraunDev:
> should we close this?

Yeah, this feature is not really necessary and causes more turmoil than it's worth.

PR #950, opened 2026-07-30
[feat] DS external scaling slices support

### LINKED ISSUE #948: Support External role scaling for multi-slice DisaggregatedSets

BenjaminBraunDev:
#### What type of PR is this?

/kind feature

Implements KEP-948: allows `scaling.mode: External` together with `spec.slices` > 1. The DisaggregatedSetRoleScaler value is treated as the role's total across all slices and split deterministically by the controller before the per-slice reconcile loop, the mid-rollout no-shrink guard applies per slice, and fresh scalers seed to the slice count. Removes the KEP-849 alpha webhook restriction. No API fields change.

Fixes #948
Related to #949

#### Testing

Tested e2e on a cluster with HPA: aggregate distribution across slices, slice count changes under an active scaler, and per-slice guard behavior under a mid-rollout scale down.

```release-note
DisaggregatedSet roles with scaling.mode: External can be combined with spec.slices > 1. The DisaggregatedSetRoleScaler replicas value is the desired total across all slices, distributed by the controller.
```

### DIFF
--- a/api/disaggregatedset/v1/disaggregatedsetrolescaler_types.go
+++ b/api/disaggregatedset/v1/disaggregatedsetrolescaler_types.go
@@ -29,11 +29,13 @@ const DisaggregatedSetRoleScalerReady string = "Ready"
 // autoscaler via the /scale subresource. The (DS, role) association is derived
 // from the scaler's controller ownerReference and its role label.
 type DisaggregatedSetRoleScalerSpec struct {
-	// Replicas is the desired replica count for the role. The controller seeds
-	// this at scaler creation — 1 for a fresh role (0 would deadlock vanilla
-	// HPA in ScalingDisabled), or the LWS's current replica count for a
-	// Static→External flip so the role does not silently drain to zero.
-	// External autoscalers overwrite it on their first tick.
+	// Replicas is the desired total replica count for the role, summed across
+	// all slices. The controller seeds this at scaler creation — the slice
+	// count for a fresh role, one group per slice (0 would deadlock vanilla
+	// HPA in ScalingDisabled), or the current replica count summed across the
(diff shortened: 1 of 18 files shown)


yankay:
ref: #949

BenAyedMedAla (commented):
Went through this — the no-shrink guard is done right for multi-slice: it compares each slice's own `targetNew` vs its own `currentNew`, not an aggregate, so a real shrink in one slice can't get blocked by a sibling slice still draining. Easy thing to get wrong here, and it isn't. `distribute()`'s monotonicity test is the right call too — a non-monotone split would look like a spurious scale event to the rollout planner.

One question: `config/rbac/role.yaml` drops what looks like a duplicate `leaderworkersets` block, unrelated to the PR's purpose. Probably fine, but worth a one-line note confirming it's intentional.

Know this is still draft — offering early feedback in case it's useful. Happy to help if it's stalled on bandwidth rather than design.

BenjaminBraunDev:
Closed as per #949 comment

PR #953, opened 2026-07-30
Extend Helm manifest synchronization beyond CRDs

### LINKED ISSUE #952: Extend Helm manifest synchronization beyond CRDs

Add an `update-helm` target that synchronizes generated manifests with the Helm chart, starting with manager RBAC.

Kueue and JobSet already use a declarative `yaml-processor` processing plan for this:

- https://github.com/kubernetes-sigs/kueue/blob/main/hack/processing-plan.yaml
- https://github.com/kubernetes-sigs/jobset/blob/main/hack/processing-plan.yaml

Run `update-helm` from `make verify` so CI detects drift between canonical manifests and Helm templates.

**Why is this needed**:

LWS currently synchronizes CRDs, but other Helm manifests are maintained separately and can drift when resources or permissions change.

**Completion requirements**:

- [ ] Add an extensible `update-helm` workflow.
- [ ] Generate Helm manager RBAC from `config/rbac`.
- [ ] Preserve Helm-specific templating.
- [ ] Run `update-helm` from `make verify`.

- [ ] Design doc
- [ ] API change
- [ ] Docs update

Dasmat13:
Closes #952

### What this PR does
- Adds an `update-helm` Makefile target using a declarative `hack/processing-plan.yaml`.
- Pins and reuses Kueue’s `yaml-processor`.
- Synchronizes generated manager and leader-election RBAC directly from `config/rbac/` into `charts/lws/templates/rbac/`.
- Keeps non-manager RBAC resources as standalone Helm templates.
- Preserves the existing Helm labels, resource identities, and conditional Volcano gang-scheduling rules.
- Integrates `update-helm` into `make verify` so CI detects drift between canonical RBAC and Helm templates.
- Makes `helm-verify` depend on `update-helm` to ensure generated templates are current before validation.
- Includes generator safety checks that remove expected generated files first and verify that all expected outputs and generated markers exist.

### Special notes for reviewers
This follows the synchronization pattern used by JobSet and reuses Kueue’s pinned `yaml-processor`. 

### Does this PR introduce a user-facing change?
NONE


Dasmat13:
Hi maintainers - friendly check-in on this Helm manifest synchronization PR when you get a chance

yankay:
Thanks for working on this. Could we simplify it by following JobSet more closely: reuse Kueue’s pinned yaml-processor, generate directly under templates/rbac/, and preserve the existing Helm labels and RBAC permissions? [yankay/lws#21](https://github.com/yankay/lws/pull/21) shows a smaller approach with semantically unchanged default and Volcano renders, preserving backward compatibility.

yankay:
Since this changes existing Helm-managed RBAC resources, would it be safer to land [kubernetes-sigs/lws#962](https://github.com/kubernetes-sigs/lws/issues/962) first, or add equivalent upgrade coverage here? This would give us a safety net to simplify the implementation following JobSet while preserving backward compatibility.

Dasmat13:
reworked this to align with the jobset pattern and reuse kueue's yaml-processor as suggested by @yankay. the generated files are now directly under templates/rbac/ and preserve the existing labels/rules. ready for another look when you have time.

yankay:
Thanks for preserving the existing Helm permissions here. I wonder if replacing the entire Pod verb list makes synchronization fragile: when I added deletecollection to the Pod RBAC marker, the canonical RBAC included it, but the generated Helm role silently dropped it and produced a duplicate Service rule. Could we avoid hard-coding the verbs, ideally by adjusting the Pod marker so controller-gen naturally generates separate Pod and Service rules.

Dasmat13:
\thanks removed the create verb from the pods rbac marker since lws doesn't create pods. since they have diff verbs now, controller-gen automatically generates them as separate rules.

also cleaned up the overrides from processing-plan.yaml so the yaml-processor syncs them dynamically without dropping future additions like deletecollection. force-pushed the changes.

Dasmat13:
@yankay quick ping, all ci is green now and the rbac sync changes are updated. ready for another look when you get time.

pkg/controllers/pod_controller.go
```diff
@@ -62,7 +62,7 @@ func NewPodReconciler(client client.Client, schema *runtime.Scheme, record event
 
 //+kubebuilder:rbac:groups="",resources=events,verbs=create;watch;update;patch
 //+kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;watch;update;patch
-//+kubebuilder:rbac:groups=core,resources=pods,verbs=create;delete;get;list;patch;update;watch
+//+kubebuilder:rbac:groups=core,resources=pods,verbs=delete;get;list;patch;update;watch
```
yankay, line 95: The removal of pods/create seems unrelated to Helm manifest synchronization. Could we leave it for a follow-up PR to keep this PR focused?

Dasmat13: good point, will revert the pods/create removal to keep this pr focused just on the helm sync. can handle the rbac tweaks in a follow up

yankay:
Verified by applying the Helm manifests before and after this PR to Kind: the only API-visible difference is rule ordering in the manager ClusterRole; resource identities and effective RBAC permissions are unchanged.

Dasmat13:
thanks for checking. i'll revert the pods/create change and fix the small review comments, then push again. will rerun the upgrade test too

Dasmat13:
@yankay updated everything in 986b499. restored pods/create, fixed the make target text and dependency, and updated the PR description too. helm generation and verify are passing now.

yankay:
Could you please note in the PR description that, beyond rule reordering, this grants `pods/create`, since the permission expansion may affect upgrades under restrictive RBAC policies?

Dasmat13:
@yankay updated the pr description

yankay:
HI @Dasmat13 
Thanks for the update. Since #997 removed `pods/create` from the canonical RBAC, could you please rebase onto the latest `main`, run `make update-helm` and `make verify`, commit the generated changes, and remove the stale `pods/create` note from the PR description? I verified locally that regeneration changes `charts/lws/templates/rbac/role.yaml`.

Dasmat13:
Hi @yankay, done — rebased onto the latest main, ran make update-helm and make verify, committed the generated RBAC changes, and removed the stale pods/create note from the PR description. Everything is passing and ready for another look.

yankay:
@Dasmat13  Thanks for the updates. I verified that, apart from manager ClusterRole rule ordering, there are no API-visible or effective RBAC changes.
/approve

yankay:
HI @Dasmat13 

We need rebase it :-)

PR #957, opened 2026-07-31
fix: ignore unrelated StatefulSets in LeaderWorkerSet watch mapper

### LINKED ISSUE #955: LeaderWorkerSet controller enqueues reconciles for unrelated StatefulSets
/kind bug

**What happened**:

`LeaderWorkerSetReconciler.SetupWithManager` watches StatefulSet events and
unconditionally maps every StatefulSet to a LeaderWorkerSet reconcile request:

https://github.com/kubernetes-sigs/lws/blob/eb27a21/pkg/controllers/leaderworkerset_controller.go#L224-L238

```go
Watches(&appsv1.StatefulSet{},
    handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, a client.Object) []reconcile.Request {
        return []reconcile.Request{
            {NamespacedName: types.NamespacedName{
                Name:      a.GetLabels()[leaderworkerset.SetNameLabelKey],
                Namespace: a.GetNamespace(),
            }},
        }
    }))
```

For a StatefulSet unrelated to LWS, the
`leaderworkerset.sigs.k8s.io/name` label is absent. The mapper therefore
returns a reconcile request with an empty name instead of ignoring the event:

```go
reconcile.Request{
    NamespacedName: types.NamespacedName{
        Name:      "",
        Namespace: "<statefulset-namespace>",
    },
}
```

Watching StatefulSets is still necessary because worker StatefulSets are not
directly owned by the LeaderWorkerSet. However, unrelated StatefulSets should
not enqueue LeaderWorkerSet reconciliations.

This appears to be a regression introduced while inlining the mapper in #791.
Before that change, the shared `enqueueLWSRequests` helper explicitly returned
`nil` when the LWS name label was missing:

```go
name := a.GetLabels()[leaderworkerset.SetNameLabelKey]
if name == "" {
    return nil
}
```

The StatefulSet watch mapper should:

- Return `nil` for StatefulSets without a non-empty
  `leaderworkerset.sigs.k8s.io/name` label.
- Return a reconcile request for LWS-managed worker StatefulSets.
- Preserve the namespace and LWS name in the generated request.

1. Create or update any StatefulSet that is unrelated to LWS.
2. The StatefulSet watch mapper reads the missing LWS label as an empty string.
3. It returns a reconcile request with an empty object name.

This can also be reproduced with a unit test by invoking the mapper with an
unlabelled StatefulSet and asserting that it currently returns one request
instead of `nil`.

**Proposed fix**:

Restore the missing empty-label check, preferably in a small named mapper
function, and add unit tests for:

- An unrelated StatefulSet without the LWS label.
- A StatefulSet with an empty LWS label.
- An LWS-managed StatefulSet with a valid LWS label.

I would be happy to work on the fix if the proposed direction is accepted.

**Environment**:

- LWS version: `main` at `eb27a21`

Dasmat13:
Fixes #955

What happened:
LeaderWorkerSetReconciler setup mapped all StatefulSets unconditionally to a reconcile request, returning a request with Name: "" when the leaderworkerset.sigs.k8s.io/name label was absent.

Fix:
Restored missing label check in a named mapper function enqueueLWSRequests to return nil for StatefulSets without the LWS label, and added unit test TestEnqueueLWSRequests.

### DIFF
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -227,15 +227,21 @@ func (r *LeaderWorkerSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&appsv1.StatefulSet{}).
 		Owns(&corev1.Service{}).
 		Watches(&appsv1.StatefulSet{},
-			handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, a client.Object) []reconcile.Request {
-				return []reconcile.Request{
-					{NamespacedName: types.NamespacedName{
-						Name:      a.GetLabels()[leaderworkerset.SetNameLabelKey],
-						Namespace: a.GetNamespace(),
-					}},
-				}
-			})).
+			handler.EnqueueRequestsFromMapFunc(enqueueLWSRequests)).
 		Complete(r)
+}
+
+func enqueueLWSRequests(ctx context.Context, a client.Object) []reconcile.Request {
+	name := a.GetLabels()[leaderworkerset.SetNameLabelKey]
+	if name == "" {
+		return nil
+	}
+	return []reconcile.Request{
+		{NamespacedName: types.NamespacedName{
+			Name:      name,
+			Namespace: a.GetNamespace(),
+		}},
+	}
 }
 
 func SetupIndexes(indexer client.FieldIndexer) error {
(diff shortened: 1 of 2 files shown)


Dasmat13:
Hi @ardaguclu  @yankay  - gentle ping on this fix to ignore unrelated StatefulSets in the watch mapper. Ready for review when you have a moment.

ahg-g:
/lgtm
/approve

Regression caused by https://github.com/kubernetes-sigs/lws/pull/791

/cc @Edwinhr716

Dasmat13:
Thanks @ahg-g, appreciate the review

Issue #960, opened 2026-08-03
[DisaggregatedSet] Add virtual sub-roles


hasB4K:
**What would you like to be added**:

Add optional virtual sub-roles to a DisaggregatedSet role. Sub-roles share one parent LWS configuration and physical LWS, but each has its own replica target or `DisaggregatedSetRoleScaler`. The parent replica count is the sum of its sub-roles, and LWS groups are assigned through labels. Parent and sub-role `-prv` Services remain available during their deprecation period.

Also allow a DisaggregatedSet and a sub-role list to contain one entry, so users can grow a standard one-role deployment incrementally.

**Why is this needed**:

Users need configuration-identical pools, such as short and long context decode pools, to avoid queue interference and decoding bubbles without duplicating role templates.

**Completion requirements**:

- [x] KEP / design doc
- [ ] API and controller implementation
- [ ] Tests and documentation

/kind feature

PR #961, opened 2026-08-03
KEP-960: DisaggregatedSet virtual sub-roles

hasB4K:
#### What type of PR is this?

/kind feature
/kind documentation

Adds KEP-960, proposing virtual sub-roles within a DisaggregatedSet role. Sub-roles share one parent LWS configuration while exposing independent static or external replica targets and routing labels. The parent replica count is the sum of its sub-roles.

This allows configuration-identical pools, such as short- and long-context decode pools, without duplicating role templates.

The initial scope supports ordinal group identity only: admission rejects a role combining `subRoles` with `groupIdentity: Hash`, which is deferred to a follow-up PR. It also adds no sub-role-specific `-prv` Services; existing parent-role Services remain unchanged during their deprecation.

Refs #960

This PR contains only the KEP. Implementation will follow separately after the design is approved.

```release-note
NONE
```

### DIFF
--- a/keps/960-DisaggregatedSet-subroles/README.md
+++ b/keps/960-DisaggregatedSet-subroles/README.md
@@ -1,0 +1,485 @@
+# KEP-960: DisaggregatedSet Sub-Roles
+
+<!--
+This KEP proposes partitioning a DisaggregatedSet role's homogeneous LWS replica
+groups into independently scalable, dynamically labelled sub-roles.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Example](#example)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API](#api)
+    - [Validation](#validation)
+  - [Labels, Identity, and Scaler Names](#labels-identity-and-scaler-names)
+  - [Controller and Assignment](#controller-and-assignment)
+  - [Rolling Updates](#rolling-updates)
+  - [Scaler Semantics](#scaler-semantics)
+  - [Routing and Deprecated PRV Services](#routing-and-deprecated-prv-services)
+  - [Status, Slices, and Compatibility](#status-slices-and-compatibility)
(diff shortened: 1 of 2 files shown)


keps/960-DisaggregatedSet-subroles/README.md
```diff
@@ -0,0 +1,485 @@
...
+
+    // Static desired group count. Defaults to 1 in the controller when
+    // scaling is Static or omitted. Must be absent for External scaling.
+    // +optional
+    // +kubebuilder:validation:Minimum=0
+    Replicas *int32 `json:"replicas,omitempty"`
+
+    // Omitted or Static uses Replicas; External uses a role scaler.
+    // +optional
+    Scaling *RoleScaling `json:"scaling,omitempty"`
```
ahg-g, line 205: Another use case is to create sub roles with different priorities. I wonder if we can shape the api in a way where explicitly adding fields here is not necessary? some patching strategy?

hasB4K: Thanks, I think I understand the idea. Instead of adding every new setting directly to `DisaggregatedSubRoleSpec`, each sub-role could define an overlay on the parent role. This could express differences such as `priorityClassName` without adding a dedicated field to `DisaggregatedSubRoleSpec` each time. Am I understanding this correctly?

I think we can already do this with separate roles by sharing their common configuration through Helm.

However, I do not think that model fits this KEP. The goal here is **only** to have virtual sub-roles: the controller dynamically reassigns existing Pods **by patching labels**. The intended use cases are:

* exposing decode pools based on context length (e.g., short, medium, and long); or
* exposing different pools that share exactly the same configuration.

Different `priorityClassName` values require different Pod templates. Since all virtual sub-roles share **one LWS**, every group receives the same leader/worker templates. An overlay that changes the template would therefore require a separate LWS for each patched sub-role, which is a different model from this KEP. For now, such pools can be modeled as separate parent roles, with Helm used to share their common configuration.

If a future use case requires server-side template sharing or overlays, we can revisit that in a separate KEP.

Does that answer your question?

ahg-g: yes, it does, thanks for the detailed answer.

PR #969, opened 2026-08-07
updating docs for 0.10 release

Edwinhr716:

Fixes #

```release-note

```

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -14,4 +14,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: us-central1-docker.pkg.dev/k8s-staging-images/lws/lws
-  newTag: release-0.10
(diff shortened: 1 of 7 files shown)


config/manager/kustomization.yaml
```diff
@@ -14,4 +14,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: us-central1-docker.pkg.dev/k8s-staging-images/lws/lws
-  newTag: main
+  newTag: release-0.10
```
yankay: Thanks for preparing this. Since this PR targets `main`, could we keep `newTag: main` here and update the release-specific tag together with `RELEASE_BRANCH` in a separate PR targeting `release-0.10`?

Edwinhr716: Fixed

PR #976, opened 2026-08-11
Drop unneeded client cert from metrics ServiceMonitor tlsConfig

### LINKED ISSUE #974: ServiceMonitor includes unneeded mTLS
**What happened**:

With `enablePrometheus=true` and `enableCertManager=true`, the rendered ServiceMonitor sets `cert` and `keySecret` in `tlsConfig`. The metrics server doesn't do client cert auth, so neither is needed.

`cmd/main.go` passes `serverTLSOpts` to the metrics server, and `BuildTLSOptions` (`pkg/config/tls.go`) only sets `MinVersion` and `CipherSuites`. There's no `ClientAuth` or `ClientCAs` anywhere in the repo, so the server never sends a CertificateRequest. Auth is TokenReview/SAR via `filters.WithAuthenticationAndAuthorization`, which the existing `bearerTokenFile` already covers.

`ca` and `serverName` are correct and should stay, the deployment mounts the cert-manager secret at controller-runtime's default metrics `CertDir`, so the server really does serve that cert.

This can cause issues with OpenTelemetry Operator configurations when the target allocator (TA) doesn't have mTLS enabled. The TA only serves real secret values to the collector over mTLS. `TLSConfig.Key` is a `config.Secret` in prometheus/common while `CA` and `Cert` are plain strings, so the collector gets the cert but `key: <secret>`, and the scrape job fails. Turning on TA mTLS is a workaround. Removing `cert`/`keySecret` would fix this properly.

A ServiceMonitor that scrapes using CA verification and a bearer token, which is what the server implements. That means dropping `cert` and `keySecret` from `charts/lws/templates/prometheus/monitor.yaml`, and from `config/components/prometheus/monitor_tls_patch.yaml` where the same block appears.

```console
$ helm template lws oci://registry.k8s.io/lws/charts/lws \
    --set enablePrometheus=true --set enableCertManager=true \
    | yq 'select(.kind == "ServiceMonitor") | .spec.endpoints[0].tlsConfig'
```

**Environment**:

- Kubernetes version (use `kubectl version`): v1.35.6-eks-bca9cf6
- LWS version (use `git describe --tags --dirty --always`): v0.9.0
- Cloud provider or hardware configuration: AWS
- OS (e.g: `cat /etc/os-release`): Bottlerocket
- Install tools: Helm
- Others: controller-runtime v0.24.1, scraped via OpenTelemetry Operator target allocator

anneheartrecord:
#### What type of PR is this?

/kind bug

The metrics server doesn't do client cert auth — auth is TokenReview/SAR via `filters.WithAuthenticationAndAuthorization`, which the existing `bearerTokenFile` already covers. `BuildTLSOptions` in `pkg/config/tls.go` only sets `MinVersion`/`CipherSuites`; there's no `ClientAuth`/`ClientCAs` anywhere in the repo, so the server never requests a client cert.

Despite that, the rendered ServiceMonitor sets `cert`/`keySecret` in `tlsConfig`, in both the Helm chart and the kustomize `PROMETHEUS-WITH-CERTS` patch. This also breaks scraping through the OpenTelemetry Operator target allocator when TA mTLS isn't enabled: the TA only forwards real secret values over mTLS, and since `TLSConfig.Key` is a `config.Secret` in prometheus/common while `CA`/`Cert` are plain strings, the collector gets a literal `<secret>` for the key and the scrape job fails.

This drops `cert`/`keySecret` from both `charts/lws/templates/prometheus/monitor.yaml` and `config/components/prometheus/monitor_tls_patch.yaml`, and updates the matching example in the Helm Prometheus docs. `ca` and `serverName` stay — the server does present that cert, and TokenReview/SAR + bearer token still needs CA verification.

Verified with `helm template lws charts/lws --set enablePrometheus=true --set enableCertManager=true` and a standalone `kustomize build` with the TLS patch applied: both now render `tlsConfig` with only `serverName`/`insecureSkipVerify`/`ca`. `enableCertManager=false` is unaffected (`insecureSkipVerify: true`).

Fixes #974

NONE

```release-note
Removed the unneeded client certificate (`cert`/`keySecret`) from the metrics ServiceMonitor's `tlsConfig` — the metrics server does not perform client cert auth, and the extra fields could break scraping through the OpenTelemetry target allocator.
```

### DIFF
--- a/charts/lws/templates/prometheus/monitor.yaml
+++ b/charts/lws/templates/prometheus/monitor.yaml
@@ -22,13 +22,6 @@ spec:
           secret:
             name: lws-metrics-server-cert
             key: ca.crt
-        cert:
-          secret:
-            name: lws-metrics-server-cert
-            key: tls.crt
-        keySecret:
-          name: lws-metrics-server-cert
-          key: tls.key
         {{- else }}
           insecureSkipVerify: true
         {{- end }}
--- a/config/components/prometheus/monitor_tls_patch.yaml
+++ b/config/components/prometheus/monitor_tls_patch.yaml
@@ -10,10 +10,3 @@
       secret:
         name: metrics-server-cert
         key: ca.crt
-    cert:
-      secret:
-        name: metrics-server-cert
-        key: tls.crt
-    keySecret:
-      name: metrics-server-cert
-      key: tls.key
--- a/site/content/en/docs/manage/prometheus.md
+++ b/site/content/en/docs/manage/prometheus.md
@@ -63,13 +63,6 @@ metrics:
         secret:
           name: lws-metrics-server-cert
           key: ca.crt
-      cert:
-        secret:
-          name: lws-metrics-server-cert
-          key: tls.crt
-      keySecret:
-        name: lws-metrics-server-cert
-        key: tls.key
 ```
 
 The secrets must reference the cert manager generated secrets.


charts/lws/templates/prometheus/monitor.yaml
```diff
@@ -22,13 +22,6 @@ spec:
           secret:
             name: lws-metrics-server-cert
             key: ca.crt
-        cert:
-          secret:
-            name: lws-metrics-server-cert
-            key: tls.crt
```
ardaguclu, line 28: Metrics server's data is scraped by Prometheus via TLS. Otherwise, Prometheus must use http to scrape the data which is strongly not recommended. 

I don't think this is a good idea to remove those configurations

ardaguclu:
> This also breaks scraping through the OpenTelemetry Operator target allocator when TA mTLS isn't enabled

I think, for production, TLS must be enabled.

charts/lws/templates/prometheus/monitor.yaml
ThatsMrTalbot: The removed configuration does not stop Prometheus using HTTPs, it stops it sending a client cert, since the server does not _validate_ a client cert this config is redundant and breaks some OTEL operator setups (#974).

The field that determines if Prometheus uses HTTP or HTTP is `servicemonitor.spec.endpoints.scheme` which is untouched in this PR:

```
$ kubectl explain servicemonitor.spec.endpoints.scheme
GROUP:      monitoring.coreos.com
KIND:       ServiceMonitor
VERSION:    v1

FIELD: scheme <string>
ENUM:
    http
    https
    HTTP
    HTTPS

DESCRIPTION:
    scheme defines the HTTP scheme to use when scraping the metrics.
```

ardaguclu:
Thank you and you are all right
/hold cancel
/ok-to-test
/lgtm

Issue #894, opened 2026-06-23
[LeaderWorkerSet] Support topology-aware scheduling with Workload API


yankay:
> **Parent tracker:** https://github.com/kubernetes-sigs/lws/issues/891  
> **Depends on:** The LWS Gang foundation defined in https://github.com/kubernetes-sigs/lws/pull/979 and implemented in https://github.com/kubernetes-sigs/lws/pull/982  
> **Blocks:** None

**What would you like to be added**:

Add a general LeaderWorkerSet design track for topology-aware scheduling through the upstream `Workload` / `PodGroup` APIs.

This should cover how an LWS can express topology placement intent for one replica or subgroup, for example:

- keep all pods in an LWS replica within the same rack or block
- place a subgroup within a topology domain
- support required and preferred topology constraints when the upstream Workload API stabilizes

An illustrative upstream `Workload` shape could look like this once the Workload / PodGroup topology constraints API is available:

```yaml
apiVersion: scheduling.k8s.io/v1alpha2
kind: Workload
metadata:
  name: lws-sample
spec:
  podGroupTemplates:
    - name: replica
      schedulingPolicy:
        gang:
          minCount: 4
      schedulingConstraints:
        topology:
          - key: topology.kubernetes.io/rack
```

For an LWS with `replicas: N` and `size: M`, LWS would translate the replica-level placement intent into one runtime `PodGroup` per replica, each copied from this template with `minCount: M` and the same topology constraint. The exact API version and field names should follow the upstream Workload / PodGroup API as it stabilizes.

This is separate from DisaggregatedSet-level TAS, which is tracked in https://github.com/kubernetes-sigs/lws/issues/892.

**Why is this needed**:

Topology-aware scheduling is useful for ordinary LWS workloads too. Multi-node inference can suffer when leader/worker or worker/worker traffic crosses slow topology boundaries.

Existing exclusive topology and affinity-based approaches provide useful static placement behavior, but they do not fully express scheduler-aware, capacity-aware topology placement. LWS should track how to expose upstream TAS once `Workload` topology constraints are ready.

Related:

- https://github.com/kubernetes-sigs/lws/issues/665
- https://github.com/kubernetes-sigs/lws/issues/859
- https://github.com/kubernetes-sigs/lws/issues/892
- https://github.com/kubernetes-sigs/lws/pull/844
- https://github.com/kubernetes-sigs/lws/tree/main/keps/666-gang-scheduling-in-lws
[...]

PR #847, opened 2026-05-06
PoC Integration with Workload and PodGroup API

Edwinhr716:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it
Part of #666 

This PR is a proof of concept of how the integration with Workload and PodGroup APIs would look like. Only covers LWS controller, doesn't cover DisaggregatedSet. Goal is to discover gaps with the alpha versions of PodGroup  and Workload APIs when it comes to inference use cases. Doc describing implementation: https://docs.google.com/document/d/1VqfNB1u8cmrRhMe0DKycX-bfaLgDm5cgHGWdWu-94cM/edit?tab=t.0#heading=h.y6q4i0rhl51e 

#### Which issue(s) this PR fixes

Fixes #

#### Does this PR introduce a user-facing change?

```release-note

```

PR #979, opened 2026-08-12
KEP-666: Gang Scheduling

### LINKED ISSUE #666: [RFC] Support kubernetes gang scheduling as a pod group provider

pacoxu:
#### What type of PR is this?
/kind feature
/kind documentation

follow up based on https://github.com/kubernetes-sigs/lws/pull/844

sync with latest v1.37 implementation
- and there is a draft implementation based on https://github.com/kubernetes-sigs/lws/pull/982.

Fixes https://github.com/kubernetes-sigs/lws/issues/666

```release-note

```


yankay:
Thanks for updating this KEP. @pacoxu, could you also take a look at the related [[JobSet KEP](https://github.com/kubernetes-sigs/jobset/pull/1253)](https://github.com/kubernetes-sigs/jobset/pull/1253)? It would help keep the shared `spec.scheduling` and `workloadbuilder` patterns aligned. Thanks! :)

ahg-g:
@yankay @Edwinhr716  does this supersede the other open PRs on the same topic? if so, can we close them and focus on this one?

ahg-g:
I am referring to https://github.com/kubernetes-sigs/lws/pull/844 and https://github.com/kubernetes-sigs/lws/pull/847

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,913 @@
...
+
+```go
+// api/leaderworkerset/v1/leaderworkerset_types.go
+type LeaderWorkerSetSpec struct {
+    // ... existing fields ...
+
+    // Scheduling defines Workload-Aware Scheduling for this LWS.
+    // Alpha; guarded by the WorkloadAwareScheduling feature gate.
+    // +optional
+    Scheduling *LeaderWorkerSetSchedulingConfiguration `json:"scheduling,omitempty"`
```
mm4tt: PTAL at https://github.com/kubernetes-sigs/lws/issues/894#issuecomment-5396495086.


I think we should aim to keep this API compatible with CPG from the beginning. Since LWS is inherently a hierarchical workload controller, we cannot model it with a flat PG. While this is fine for an alpha/MVP, we will eventually need CPG support, so we should design the API in a way that makes this possible in the future.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1059 @@
...
+| `GenericWorkload` | Beta, default `false` | Operators must explicitly enable it in kube-apiserver and kube-scheduler. |
+| `gang.minCount` | Mutable | LWS can support elastic size changes without rejecting all size updates. |
+| Workload templates | Existing entries are updateable where their fields allow it; entries cannot be added or removed | Phase 1 uses stable leaf templates for the selected level. Phase 2 compiles the same LWS API into nested composite and leaf templates. |
+| PodGroup protection | PodGroups have deletion protection | LWS owns PodGroups independently of leader Pods and follows ordered cleanup. |
+| Workload-aware preemption ([KEP-5710][kep5710]) | Beta behavior under `GenericWorkload`; no separate feature gate | The PodGroup priority is authoritative and every member Pod must have the same effective priority. |
+| `TopologyAwareWorkloadScheduling` ([KEP-5732][kep5732]) | Alpha, default `false` in `release-1.37` | Topology constraints require a separate cluster prerequisite. The KEP targets Beta, but the 1.37 release-branch gate did not graduate. |
```
kannon92, line 146: IMO I would go forward with implementing this feature assuming TAS and DRAWorklaodResourceClaims are possible.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1059 @@
...
+type LeaderWorkerSetSpec struct {
+    // ... existing fields ...
+
+    // Scheduling defines Workload-Aware Scheduling for this LWS.
+    // Alpha; guarded by the WorkloadAwareScheduling feature gate.
+    // +optional
+    Scheduling *LeaderWorkerSetScheduling `json:"scheduling,omitempty"`
+}
+
+type LeaderWorkerSetScheduling struct {
```
kannon92, line 280: Should we add support for ResourceClaims at top level? 

For JobSet we did. It was mostly shorthand but we could see a benefit of people specifiyng everything at level 1.

kannon92:
@pacoxu this seems pretty good. Why is it draft?

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Agree. Let me update.

pacoxu:
> @pacoxu this seems pretty good. Why is it draft?

Updated. KEP is ready for review, I think.

I am working on a demo implementation in https://github.com/kubernetes-sigs/lws/pull/982 and want to make sure the KEP works as expected.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: I’d like to support both, but the relevant upstream APIs are still evolving and may introduce breaking changes. 

I’d prefer to revisit the concrete integration after the v1.38 KEP freeze. 

In the meantime, we can make sure KEP-666 does not prevent future TAS or `DRAWorkloadResourceClaims` support. Should the actual integration stay in this KEP, or be handled in a follow-up LWS KEP?

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1059 @@
...
+   `leaderworkerset.sigs.k8s.io/exclusive-topology` until their combined
+   placement and failure semantics are tested.
+10. The selected scheduler provider supports every requested field and active
+   level.
+11. Shared ResourceClaims have matching references in every member pod
+   template that consumes them.
+12. All pods mapped to the same leaf PodGroup have the same effective
+    `priorityClassName`. The default flat replica mode therefore requires the
+    leader and workers to match; leader/worker leaf mode may use different
+    priorities. LWS copies each leaf's common value into its Workload template.
```
yankay: HI @pacoxu 

I may be missing something, but it seems that different leader and worker priorities may not be implementable with Kubernetes 1.37 as currently written. Kubernetes requires identical `priorityClassName` and `priority` across all templates in a Workload, so the API server would reject this role-mode Workload.

Could we require one effective priority for the entire Workload in Phase 1 and defer mixed-role priorities until upstream supports them?

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1059 @@
...
+2. Each policy or disruption union at an active level selects exactly one
+   variant after defaulting.
+3. In Phase 1, exactly one scheduling level is active: top-level fields,
+   replica-level fields, or `replica.leader` / `replica.worker` leaves. Phase
+   2 may combine levels only when both the LWS and Kubernetes CPG gates are
+   enabled.
+4. The scheduling hierarchy cannot be added, removed, or switched between
+   levels after LWS creation. Policy variants and immutable constraints cannot
+   change. Mutable generated leaf `gang.minCount` values continue to follow
+   LWS cardinality.
```
yankay: Thanks for documenting the update semantics here. It looks like fields such as scheduling constraints, resource claims, and disruption mode are already treated as immutable. Would it make sense to do the same for the effective `priorityClassName` while scheduling is enabled, since it is derived from mutable pod templates but compiled into an immutable Workload template?

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Good catch, This is an alpha limitation. 

https://github.com/kubernetes/enhancements/blob/c4f439c2dd4acb928094660be0ea771bf63f2b76/keps/sig-scheduling/6012-composite-podgroup-api/README.md?plain=1#L897-L901

> For the **Alpha** release, we enforce a strict single-priority constraint: all
> member groups and pods within a single group hierarchy tree **must share the exact
> same priority and PriorityClassName**. Support for differing group-level
> priorities under basic scheduling policies will be explored for the **Beta**
> release.

I intentionally kept priority on the leader/worker pod templates so that, if KEP-6012 Beta later supports differing group-level priorities, LWS can relax validation without changing its API shape.

But this may be updated later, I just updated it according v1.37 limitation.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1102 @@
...
+
+When an LWS has a registered controller owner and the root delegates runtime
+group management:
+
+1. LWS follows the controller-owner chain and discovers the root Workload.
+2. The parent supplies
+   `scheduling.k8s.io/group-template-name` on the child LWS to select the
+   Workload template.
+3. If the LWS leaf belongs below a runtime CompositePodGroup, the parent also
+   supplies `scheduling.k8s.io/parent-composite-podgroup`.
```
yankay: It may be `scheduling.k8s.io/parent-compositepodgroup`, ref to 
https://github.com/kubernetes/kubernetes/blob/release-1.37/pkg/apis/scheduling/annotation_key_constants.go#L33

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1102 @@
...
+4. LWS uses `NewBuilderFromExistingWorkload` and creates the selected runtime
+   replica or role groups from the persisted template.
```
yankay: Thanks for clarifying the Phase 2 design. I wonder if we could state here that delegated role mode is not supported in Phase 1, since one `group-template-name` selects only one flat template. Would it make sense for Phase 1 admission to reject this combination until CompositePodGroup support is implemented? :)

keps/666-gang-scheduling-in-lws/README.md
kannon92: Your API supports TopologyAwareScheduling though.

`schedulingConstraints`

keps/666-gang-scheduling-in-lws/kep.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - "/keps/407-gang-scheduling"
+  - "/keps/766-DisaggregatedSet"
+
+# The target maturity stage in the current dev cycle for this KEP.
+stage: alpha
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v0.10.0"
```
andreyvelich: should be updated, right?

```suggestion
latest-milestone: "v0.11.0"
```

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1102 @@
...
+
+The validating webhook enforces:
+
+1. The LWS `WorkloadAwareScheduling` gate is enabled.
+2. Each policy or disruption union at an active level selects exactly one
+   variant after defaulting.
+3. In Phase 1, exactly one scheduling level is active: top-level fields,
+   replica-level fields, or `replica.leader` / `replica.worker` leaves. Phase
+   2 may combine levels only when both the LWS and Kubernetes CPG gates are
+   enabled.
```
andreyvelich: Do we have a use-case when leader + workers should use basic scheduling?
If not, shall we enforce that leader + worker CPG must always define gang scheduling?

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1102 @@
...
+
+### Non-Goals
+
+- Materializing `CompositePodGroup` objects in the first phase. The user API
+  includes the hierarchy now, but nested levels are enabled in a later phase
+  because `CompositePodGroup` is alpha in Kubernetes 1.37.
+- Extending this KEP's three-level LWS hierarchy with the additional role and
+  slice levels needed by [DisaggregatedSet][kep766].
+- Combining `startupPolicy: LeaderReady` with a gang that contains both the
+  leader and workers.
```
andreyvelich, line 112: What would be the recommendation to use startupPolicy with gang scheduling for both leader and workers resources?

keps/666-gang-scheduling-in-lws/kep.yaml
andreyvelich: Are we going to support LWS scheduling API under feature gate?

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1102 @@
...
+    // +listMapKey=name
+    ResourceClaims []schedulingv1alpha3.WorkloadPodGroupResourceClaim `json:"resourceClaims,omitempty"`
+
+    // Leader defines the level-3 leader PodGroup.
+    // +optional
+    Leader *LeaderWorkerSetPodGroupScheduling `json:"leader,omitempty"`
+
+    // Worker defines the level-3 worker PodGroup.
+    // +optional
+    Worker *LeaderWorkerSetPodGroupScheduling `json:"worker,omitempty"`
```
andreyvelich: IIRC, we discussed with @mm4tt before on CPG APIs, that it is recommended to have dedicated struct for each object, even if they have the same fields. So we can extend the API further if needed without breaking it.


```suggestion
    // Leader defines the level-3 leader PodGroup.
    // +optional
    Leader *LeaderWorkerSetLeaderScheduling `json:"leader,omitempty"`

    // Worker defines the level-3 worker PodGroup.
    // +optional
    Worker *LeaderWorkerSetWorkerScheduling `json:"worker,omitempty"`
```

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1102 @@
...
+### Defaulting and Validation
+
+Defaulting is controller-side so the stored LWS preserves user intent:
+
+- `spec.scheduling` absent means the new upstream integration is disabled for
+  that LWS. No upstream Workload or PodGroup is created.
+- `spec.scheduling: {}` selects replica mode and defaults to
+  `replica.schedulingPolicy.gang: {}`. In Phase 1, LWS lowers that composite
+  intent to one PodGroup per replica with `gang.minCount` equal to
+  `spec.leaderWorkerTemplate.size`.
```
andreyvelich: Should this be `.spec.replicas * .spec.leaderWorkerTemplate.size`?

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1102 @@
...
+11. Shared ResourceClaims have matching references in every member pod
+   template that consumes them.
+12. All pod templates represented by an LWS-managed Workload must use one
+    Workload-wide effective `priorityClassName`, irrespective of the selected
+    scheduling level. LWS copies this common value into every PodGroupTemplate
+    and CompositePodGroupTemplate. Different class names are rejected even if
+    their PriorityClass objects have the same numeric priority. Mixed-role
+    priorities are deferred until [upstream Workload validation][workload-validation]
+    supports them.
+
```
andreyvelich: Shall we also restrict `schedulingGroup` in the PodTemplate?

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1102 @@
...
+The Workload is referenced through `spec.workloadRef`, not a second
+ownerReference, matching `workloadbuilder.NewPodGroup`. A leader Pod cannot
+own an object that must exist before the leader, and must not be the owner
+across a same-name restart.
+
+On scale-down or rollout cleanup, LWS deletes member pods first, then the
+PodGroup, and waits for the protection finalizer. Deleting the LWS relies on
+owner-based garbage collection, with reconciliation providing ordered
+best-effort cleanup.
```
andreyvelich, line 805: Shall we also delete the Workload object like we do for suspended JobSet: https://github.com/kubernetes-sigs/jobset/tree/main/keps/969-WAS-integration#workload-lifecycle

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1102 @@
...
+        # ...
+```
+
+LWS compiles one Workload:
+
+```yaml
+apiVersion: scheduling.k8s.io/v1beta1
+kind: Workload
+metadata:
+  name: inference
```
andreyvelich: We might need to add hash to the Workload object too.
What if users have another CRD that creates Workload API with the same name: `inference`?

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1102 @@
...
+   variant after defaulting.
+3. In Phase 1, exactly one scheduling level is active: top-level fields,
+   replica-level fields, or `replica.leader` / `replica.worker` leaves. Phase
+   2 may combine levels only when both the LWS and Kubernetes CPG gates are
+   enabled.
+4. The scheduling hierarchy cannot be added, removed, or switched between
+   levels after LWS creation. Policy variants and immutable constraints cannot
+   change. The effective Workload-wide `priorityClassName`, derived from the
+   leader and worker pod templates, is also immutable while `spec.scheduling`
+   is present. Mutable generated leaf `gang.minCount` values continue to follow
```
andreyvelich: @mm4tt @pacoxu @yankay @kannon92 Do we expect that `priorityClassName` for PodGroup will derive from the underlying PodTemplate? I think, we haven't discussed this as part of JobSet KEP.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1102 @@
...
+    SchedulingConstraints *schedulingv1alpha3.WorkloadCompositePodGroupSchedulingConstraints `json:"schedulingConstraints,omitempty"`
+
+    // DisruptionMode defines how replica groups may be disrupted.
+    // Immutable after creation.
+    // +optional
+    DisruptionMode *schedulingv1alpha3.WorkloadCompositePodGroupDisruptionMode `json:"disruptionMode,omitempty"`
+
+    // Replica defines level-2 scheduling for each LWS replica.
+    // +optional
+    Replica *LeaderWorkerSetReplicaScheduling `json:"replica,omitempty"`
```
andreyvelich, line 312: Within the LWS scope, replicas are always homogeneous, right?

I am asking if we ever see use-cases when users want to configure different scheduling parameters for each replica.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1102 @@
...
+- Each replica (and each role leaf or CPG in later modes) adds scheduling
+  objects, and revision-specific names increase that count during surge.
+- Supporting both upstream and third-party providers increases validation and
+  test complexity.
+- Embedding alpha building-block types means the LWS alpha API may need to
+  track upstream package changes before graduation.
+- Phase 1 must maintain a lowering adapter because LWS and replica fields use
+  composite building blocks while the controller materializes leaf PodGroups.
+
+## Alternatives
```
andreyvelich, line 1351: As another alternative we can add @tenzen-y's suggestion to implement WAS via LWS labels to avoid API breaking changes before WAS go to GA (if any).

andreyvelich (commented):
Thanks @pacoxu!
This looks great, I left a few comments/suggestions.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1102 @@
...
+### Goals
+
+- Support upstream Kubernetes gang scheduling for an LWS replica.
+- Represent the LWS, replica, and leader/worker scheduling levels in the API
+  from the first release.
+- Represent one LWS as one Workload and, in the default first-phase mode, each
+  active replica as one PodGroup.
+- Use the standard `v1alpha3` WAS building blocks and `workloadbuilder`
+  translation library.
+- Create scheduling objects in the strict order Workload, PodGroup, then Pod.
```
andreyvelich: Same question as here: https://github.com/kubernetes/enhancements/pull/6295
Does the creation order really matter for WAS?

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Updated

keps/666-gang-scheduling-in-lws/kep.yaml
pacoxu: Updated.

pacoxu: Yes, probably WorkloadAwareScheduling feature gate

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Added.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Removed as this may be not a goal of the KEP. 

The API server rejects a PodGroup whose workloadRef points at a missing Workload. The controller, not the API, is responsible for creating the PodGroup before its member Pods.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1400 @@
...
+The conversion contract is:
+
+| LWS-owned input | Flat target | Composite target |
+| --- | --- | --- |
+| Composite policy on LWS or replica | Convert Basic directly; lower an empty Gang to a pod Gang with LWS-derived `minCount`; reject explicit `minGroupCount` | Convert field-by-field to `WorkloadCompositePodGroupSchedulingPolicy` |
+| Pod policy on leader or worker | Convert field-by-field to `WorkloadPodGroupSchedulingPolicy` | Same leaf conversion |
+| Constraints and disruption | Convert to the upstream pod-group variant | Convert to the upstream composite or pod-group variant according to the materialized node |
+| Resource claim | Convert to `WorkloadPodGroupResourceClaim` on the selected flat or leaf group | Convert only on a leaf; composite placement is rejected |
+
+The alpha feature gate does not waive compatibility for JSON already stored in
```
mm4tt: This is probably the most significant architectural divergence from what we are doing across other controller integrations (e.g., Job in KEP-5547 and JobSet in KEP-969), and I'd like to understand the rationale a bit better.

A few thoughts:
1. Overall, this is classic chicken-and-egg problem: for the reusable building blocks to graduate to `v1`, we need controllers in the ecosystem to adopt them and prove their stability. Conversely, controllers are reluctant to adopt them until they are `v1`... 
2. `batch/v1` Job is also a stable `v1` API, yet we embedded the `scheduling.k8s.io/v1alpha3` building blocks directly into `JobSchedulingConfiguration`.
3. Since this integration in LWS is guarded by the `WorkloadAwareScheduling` feature gate, why don't the same compatibility and gate-guarding arguments that worked for the `v1` Job API apply to LWS?
4. We already have practically complete designs for multiple other controllers adopting these building blocks (JobSet, TrainJob, Deployment, StatefulSet). Because of this broad alignment and validation across the ecosystem, from the SIG Scheduling side we can consider promoting the scheduling building blocks to v1 (sic!) in 1.38.

Given this, could we reconsider embedding the upstream building blocks directly rather than maintaining owned structs and an explicit conversion layer in LWS?

mm4tt (commented):
Thanks @pacoxu for driving this!

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1400 @@
...
+  CPGs -> leader/worker PodGroups`) in the controller.
+
+This is a future goal because CompositePodGroup is alpha in Kubernetes 1.37.
+Alpha APIs are difficult to support for many users, so nested CPG creation
+stays behind a separate LWS gate. The user-facing field already reserves the
+hierarchy; Phase 1 only creates Workload and flat PodGroups.
+
+### Non-Goals
+
+- Extending this KEP's three-level LWS hierarchy with the additional role and
```
ahg-g: This is ok, but can we clarify how in the future we intend to support that, or at least how this proposal takes it into account? 

I have two cases in mind:
1) Scheduling a minimum prefill to decode ratio for slice.
2) Ensuring that the prefill and decode replicas of a slice are scheduled in the same domain.

Would these cases be addressed by CompositePodGroup?

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1400 @@
...
+    // +listType=map
+    // +listMapKey=name
+    ResourceClaims []LeaderWorkerSetResourceClaim `json:"resourceClaims,omitempty"`
+}
+
+// LeaderWorkerSetCompositeSchedulingPolicy selects the policy for a group of
+// child groups. Exactly one member must be set.
+type LeaderWorkerSetCompositeSchedulingPolicy struct {
+    // +optional
+    Basic *LeaderWorkerSetBasicSchedulingPolicy `json:"basic,omitempty"`
```
ahg-g: pls document what Basic is intended to capture here especially that it is defined as an empty struct now.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1400 @@
...
+    Gang *LeaderWorkerSetPodGangSchedulingPolicy `json:"gang,omitempty"`
+}
+
+type LeaderWorkerSetBasicSchedulingPolicy struct{}
+
+type LeaderWorkerSetCompositeGangSchedulingPolicy struct {
+    // MinGroupCount is a positive minimum number of child groups.
+    // +optional
+    // +kubebuilder:validation:Minimum=1
+    MinGroupCount *int32 `json:"minGroupCount,omitempty"`
```
ahg-g: At level 2, what does a group represent? either worker or leader? so the max here can be 2?

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1400 @@
...
+// may be disrupted. Exactly one member must be set.
+type LeaderWorkerSetDisruptionMode struct {
+    // +optional
+    Single *LeaderWorkerSetSingleDisruptionMode `json:"single,omitempty"`
+    // +optional
+    All *LeaderWorkerSetAllDisruptionMode `json:"all,omitempty"`
+}
+
+type LeaderWorkerSetSingleDisruptionMode struct{}
+type LeaderWorkerSetAllDisruptionMode struct{}
```
ahg-g: pls document the semantics of each, this is important especially that both are empty structs now.

keps/666-gang-scheduling-in-lws/README.md
ahg-g: +1 to embedding the upstream as long as they have gone through enough iterations.

ahg-g (commented):
thanks @pacoxu , do you think we can add diagrams to show what WAS objects get created in the various supported scenarios?

Another thing to keep in mind is support for disaggSet, I see we explicitly put that out of scope, which is fair, but I think we at least need to discuss how it may look like in the future.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Thanks @mm4tt and @ahg-g. Agreed and updated. LWS now directly embeds the upstream scheduling building blocks while retaining only the LWS-specific hierarchy. The duplicated types and conversion layer have been removed.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: I added a future DisaggregatedSet hierarchy. CPG topology can provide co-location, while separate role minima can represent fixed prefill/decode counts. A single minGroupCount cannot directly express a weighted ratio.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Agreed and updated. Basic now means independent scheduling and I add a table to describe the details.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Yes. `minGroupCount` always counts immediate child groups, not pods. So it should be either 1 or 2. Updated.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Agreed and updated. Also in the table.

pacoxu:
> thanks @pacoxu , do you think we can add diagrams to show what WAS objects get created in the various supported scenarios?

I added diagrams for the Phase-1 modes, the Phase-2 hierarchy.

> Another thing to keep in mind is support for disaggSet, I see we explicitly put that out of scope, which is fair, but I think we at least need to discuss how it may look like in the future.

Add a new section for the future DisaggregatedSet integration.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Yes. Basic is useful when users want WAS features without gang admission, or as a structural parent for different leader and worker policies. So we should not require every replica CPG to use Gang.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: `LeaderReady` will be rejected when the gang contains both leader and workers because it would create a startup deadlock. This is mentioned now. I add a table to describe all scanerios.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Good catch. Updated.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Yes. LWS replicas are homogeneous, and the replica scheduling configuration applies uniformly to every replica. Per-replica scheduling overrides may be out of scope?

keps/666-gang-scheduling-in-lws/README.md
pacoxu: IIUC, yes. LWS derives it from the underlying pod templates. All templates in one Workload must have the same effective priorityClassName; otherwise the LWS is rejected.

Let me check jobset KEP later.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Agreed and updated.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: LWS currently has no first-class suspend semantics. Scaling replicas to zero is still a scaling operation, rather than suspension. If LWS introduces spec.suspend in the future, we can align that lifecycle with JobSet.

ahg-g:
Thanks, looks good to me!

@mm4tt  @yankay any final thoughts?

kannon92:
cc @wojtek-t

keps/666-gang-scheduling-in-lws/README.md
andreyvelich: Yes, that makes sense to me.

keps/666-gang-scheduling-in-lws/README.md
andreyvelich: @kannon92 @mimowo @tenzen-y For the LWS <-> Kueue integration, did we ever have a discussion to introduce `.suspend` field into LWS API?

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1470 @@
...
+    // may share. This is supported only while level 1 is selected as a flat
+    // PodGroup. It is mutually exclusive with Replica because a root CPG
+    // cannot own a pod-level resource claim; nested configurations put claims
+    // on replica or role leaves.
+    // Immutable after creation.
+    // +optional
+    // +kubebuilder:validation:MaxItems=4
+    // +listType=map
+    // +listMapKey=name
+    ResourceClaims []schedulingv1alpha3.WorkloadPodGroupResourceClaim `json:"resourceClaims,omitempty"`
```
andreyvelich, line 378: CPG doesn't support RC today, so we should remove it from the API.

```suggestion
```

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1470 @@
...
+    // ResourceClaims lists dynamic resource claims shared by replica members.
+    // A flat representation supports this while replica is the selected leaf. It is
+    // mutually exclusive with leader and worker because a CPG cannot own a
+    // pod-level resource claim; nested configurations put claims on the leaves.
+    // Immutable after creation.
+    // +optional
+    // +kubebuilder:validation:MaxItems=4
+    // +listType=map
+    // +listMapKey=name
+    ResourceClaims []schedulingv1alpha3.WorkloadPodGroupResourceClaim `json:"resourceClaims,omitempty"`
```
andreyvelich, line 412: same

```suggestion
```

keps/666-gang-scheduling-in-lws/README.md
kannon92: I don't think that was ever discussed. LWS uses scheduling gates which relies on the pod-integration.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1470 @@
...
+KEP.
+
+### Observability
+
+Users can inspect:
+
+- the Workload and its `controllerRef`;
+- the PodGroups selected by the active Phase-1 level and their `workloadRef`;
+- in Phase 2, each CPG's parent link and the complete root-to-leaf chain;
+- `PodGroup.status.conditions[type=PodGroupInitiallyScheduled]`;
```
aliceco01: @pacoxu this condition is terminal: per KEP-4671 it never reverts to False "even if pods are subsequently evicted and group constraints are no longer met", and the scheduler enforces that with an explicit guard in `updatePodGroupCondition`. So it answers "was this group ever placed", not "is this replica healthy now". Since this KEP also reuses the replica PodGroup across a leader restart (line 797), the condition will read True for a recreated group whose members are all Pending.

Suggest either wording this bullet as an initial-placement signal only, or adding a note that runtime group health has no PodGroup-level source today and that LWS derives it from pods. I'm proposing a runtime condition upstream (kubernetes/kubernetes#136334) that would give `WorkloadSchedulingReady` a real source later.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1470 @@
...
+`uid-hash` is the same deterministic DNS-safe short hash of `lws.UID` used in
+the Workload name. `name-prefix` truncates `lws.Name` as needed so the complete
+name satisfies the target API's length limit. Reconciliation always verifies
+the controller ownerReference and expected labels before reusing a computed
+name; a hash collision is surfaced rather than adopted.
+
+The UID component prevents a newly created, same-name LWS from colliding with
+deletion-protected groups that belong to the previous UID. Revision-aware
+suffixes avoid collisions while old and new replicas of one LWS coexist during
+a rolling update. A leader restart within the same UID and revision reuses the
```
aliceco01: Reusing the group across a leader restart is a change from today's Volcano provider, where the PodGroup is owned by the leader Pod and is destroyed and recreated with it under `RecreateGroupOnPodRestart`. Two things I could not find covered for the failure path:

1. During the foreground deletion of the old members, the reused PodGroup's scheduler cache state still counts terminating members as scheduled. Members are only dropped on the pod delete event (`RemovePodGroupMember` in `pkg/scheduler/eventhandlers.go`; `podgroupstate.go` has no deletion-timestamp handling). So `scheduled + feasible >= minCount` can admit part of the new generation on the strength of pods that are about to disappear, and the new generation is no longer gang-placed as a unit. Should the controller wait for the old members to be gone before creating the new ones, or is that window acceptable?

2. `pendingPodsInGroup` (`pkg/controllers/pod_controller.go:440`) suppresses recreation under `RecreateGroupAfterStart` whenever any member is Pending. Under a native gang that is waiting for capacity, Pending is a steady state, so recreation is suppressed for as long as capacity is missing. Does the guard need a different signal once groups are native? The condition proposed in kubernetes/kubernetes#136334 would be one candidate.

keps/666-gang-scheduling-in-lws/README.md
pacoxu:
```
    // ResourceClaims is valid only when this level is compiled as a flat
    // PodGroup (no nested replica). CompositePodGroup cannot own pod-level
    // claims; admission rejects this field together with Replica.
```

CPG can't own ResourceClaims, but these fields are only for the flat PodGroup case (already mutually exclusive with nested children). I'd keep them for whole-LWS / replica gangs and just make the comments/validation clearer. Same for the replica-level field.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Same as above.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1470 @@
...
+`scheduling.k8s.io/v1alpha3` reusable controller building blocks for policy,
+constraints, disruption, and resource claims. This keeps the scheduling
+vocabulary aligned with Job, JobSet, and other workload controllers while LWS
+retains ownership of how its LWS, replica, leader, and worker levels are
+selected. kube-scheduler consumes the resulting `v1beta1` Workload and PodGroup
+objects. The feature is guarded by the LWS `WorkloadAwareScheduling` gate,
+delivered through the LWS Configuration API, and selected via the existing
+`gangSchedulingManagement.schedulerProvider` setting. The upstream
+`kubernetes` provider also requires Kubernetes `GenericWorkload` on both
+kube-apiserver and kube-scheduler.
```
yankay: Would it make sense to also include `kube-controller-manager` in the `GenericWorkload` prerequisites? :-)

yankay:
> Thanks, looks good to me!
> 
> @mm4tt @yankay any final thoughts?

Thanks, @pacoxu @ahg-g .
LGTM too. Let's move forward.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Updated

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Good catch.  Updated as `as an  initial-placement signal only;`.


A live PodGroup-level health condition would be useful later if upstream adds one.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Agreed.

1. We'll wait until old members are fully gone before creating the new generation. Overlapping terminating and new pods in the same PodGroup is not acceptable. 

2. Native gang does make `pendingPodsInGroup` the wrong AfterStart signal, but this KEP does not redefine `RecreateGroupAfterStart` — that can be a follow-up?

keps/666-gang-scheduling-in-lws/README.md
andreyvelich: Yeah, but once we support CPG in LWS in the phase 2, we should always create CPG and PG, isn't?
This is why we didn't introduce ResourceClaim on the Level 1 and Level 2 in JobSet, to avoid breaking changes in the phase 2: https://github.com/kubernetes-sigs/jobset/tree/main/keps/969-WAS-integration#api

Could you clarify here please @kannon92 @mm4tt?

pacoxu: Got it. So here we may follow jobset design then, only job level pod group has the resource claim.

pacoxu: Updated.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1335 @@
...
+
+This KEP targets the Kubernetes 1.37 APIs, not the earlier `v1alpha2` design:
+
+| Area | Kubernetes 1.37 state | LWS consequence |
+| --- | --- | --- |
+| Workload and PodGroup runtime APIs | `scheduling.k8s.io/v1beta1` | LWS creates and watches `v1beta1` objects. |
+| Reusable controller API blocks | `scheduling.k8s.io/v1alpha3` | The LWS-owned hierarchy directly uses `WorkloadPodGroup*` and `WorkloadCompositePodGroup*` types. LWS-specific validation chooses the structural level and preserves flat/composite representation. |
+| `workloadbuilder` | Shipped in `k8s.io/component-helpers/scheduling/schedulingv1/workloadbuilder` | LWS uses the release implementation for validation, Workload compilation, and PodGroup materialization. |
+| `GenericWorkload` | Beta, default `false` | Operators must explicitly enable it in kube-apiserver, kube-controller-manager, and kube-scheduler. kube-controller-manager runs the PodGroup protection controller. |
+| Gang minima (`minCount`, `minGroupCount`) | Mutable, but both must remain positive | LWS can support elastic size and replica changes; zero replicas use an unused template placeholder of `1`. |
```
wojtek-t: minPodGroupCount (used in CompositePodGroup) is still immutable in 1.37 (when CPG is in Alpha).
It will become mutable in 1.38 as part of beta release though.

wojtek-t: Also - can you clarify what do you mean by "zero replicas use an unused template placeholder of 1." ?

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1335 @@
...
+| Workload templates | Existing entries are updateable where their fields allow it; entries cannot be added or removed | A single-level shape always uses stable leaf templates for the selected level. A multi-level shape, admitted only in Phase 2, compiles into nested composite and leaf templates. |
```
wojtek-t: minCount is the only mutable field in the PodGroupTemplates for now.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1335 @@
...
+   requires one effective priority across the entire Workload.
+
+Delivery is split into two phases:
+
+- **Phase 1:** LWS creates only Workload and PodGroup objects. Admission
+  permits exactly one active level: whole LWS, replica, or leader/worker
+  leaves. This single-level shape is always lowered to flat PodGroup
+  templates. An empty `spec.scheduling` selects replica mode and Gang
+  scheduling. Leader and worker leaves are admitted independently in this
+  phase; coordinating them as a gang of groups requires Phase 2.
```
wojtek-t: Can you clarify what do you mean by "empty spec.scheduling selects replica mode"?

Are you effectively saying that the feature from now on is a new default behavior (assuming feature gate is on)?

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1335 @@
...
+   This allows workers to request an NVLink domain or shared DRA claim without
+   forcing the leader to use the same resources. Kubernetes 1.37 still
+   requires one effective priority across the entire Workload.
+
+Delivery is split into two phases:
+
+- **Phase 1:** LWS creates only Workload and PodGroup objects. Admission
+  permits exactly one active level: whole LWS, replica, or leader/worker
+  leaves. This single-level shape is always lowered to flat PodGroup
+  templates. An empty `spec.scheduling` selects replica mode and Gang
```
wojtek-t: To clarify, by empty you mean that it's set to `{}` right?

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1335 @@
...
+the controller derives template names, instances, membership, and parent
+links. This matches the JobSet phased approach of shipping flat PodGroups
+before CPG materialization ([JobSet KEP-969][jobset-kep969]).
+
+[jobset-kep969]: https://github.com/kubernetes-sigs/jobset/pull/1253
+
+### Generated WAS Object Shapes
+
+Runtime groups are owned by LWS and reference their Workload template through
+`workloadRef` unless a parent controller owns the Workload.
```
wojtek-t: It should reference the template even if the parent controller owns the Workload.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1335 @@
...
+   replica levels reject the field in every shape, including when that level
+   is the selected Phase-1 flat PodGroup. A replica-wide claim covering both
+   roles is not expressible. A flat shape also rejects explicit composite
+   `minGroupCount` and all other fields that require a runtime CPG. In a
+   Phase-2 composite shape, replica-level `minGroupCount` must be `1` or `2`
+   because its immediate children are the leader and worker PodGroups.
+6. In delegated flat mode, one `group-template-name` can select only one
+   flat template. Whole-LWS and replica modes are supported, but
+   leader/worker mode is rejected until a composite shape can select a root
+   CPG and materialize both role descendants.
```
wojtek-t, line 530: Hmm - I'm trying to understand what it would even mean in that case.

We would be saying that a parent controller owns a Workload and come (potentially hierarchical) CPG(s) but the pods are LWS-owned and we wand leader/worker structure so two PodGroups with separate templates.

As of now, the API isn't even prepared for that kind of evolution and my current thinking is that we don't even want to make the generic API to support that.

I think that in phase 2 this problem automatically disappears and we shouldn't try to address the flat mode and simply use the structured mode in such case.

@mm4tt @kannon92 - in case you disagree

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1335 @@
...
+11. The selected scheduler provider supports every requested field and active
+   level.
+12. Every ResourceClaim entry has a valid name and selects exactly one of
+    `resourceClaimName` or `resourceClaimTemplateName`. Shared claims have
+    matching references in every member pod template that consumes them.
+13. All pod templates represented by an LWS-managed Workload must use one
+    Workload-wide effective `priorityClassName`, irrespective of the selected
+    scheduling level. LWS copies this common value into every PodGroupTemplate
+    and CompositePodGroupTemplate. Different class names are rejected even if
+    their PriorityClass objects have the same numeric priority. Once
```
wojtek-t, line 555: This can potentially be relaxed in the future, because we generally compare the integer values resolved from the PriorityClassName.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1335 @@
...
+    `spec.scheduling` is set, an update that changes the effective class of any
+    represented pod template is rejected because the compiled Workload
+    priority fields are immutable. Mixed-role priorities are deferred until
+    the upstream Workload API supports them.
```
wojtek-t, line 559: Question: do you have real life usecases for mixed priorities? If so, can you please describe them or point me to them?

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1335 @@
...
+
+- a controller ownerReference to the LWS, never to the leader Pod;
+- labels for LWS name, active level, optional group index, role, and template
+  revision;
+- `spec.workloadRef.workloadName` and the selected `templateName` (`lws`,
+  `replica`, `leader`, or `worker`);
+- an inline copy of the resolved template fields.
+
+The Workload is referenced through `spec.workloadRef`, not a second
+ownerReference. A leader Pod cannot own an object that must exist before the
```
wojtek-t: I would suggest to have an additional ownerRef (with controller=false) also set on the PodGroup.

This would mimic what we did in Job:
https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/5547-integrate-workload-with-job#ownerreferences-relationship

as it facilitates better object lifecycle.
And will help you avoid leaking resources under some races (which I can imagine if you do rollouts/scaleup/... right before deleting LWS itself).

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1335 @@
...
+The annotations are controller-to-controller linkage. Reject a missing
+template, invalid owner chain, or missing parent CPG. Block pods until the
+delegated Workload and required parent instance exist.
+
+In Phase 2, `group-template-name` selects the LWS root composite template; LWS
+then materializes that CPG and its replica and role descendants, so delegated
+leader/worker mode needs only one annotation.
+
+Use the Kubernetes 1.37 KEP-6089 linkage names. The implementation-sync
+([kubernetes/enhancements#6244][kep6089-sync]) is still open and the keys are
```
wojtek-t: nit: it was merged last week

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1335 @@
...
+└── Workload (owned by DisaggregatedSet)
+    └── slice CompositePodGroup (owned by DisaggregatedSet)
+        ├── prefill LWS root CPG → per-replica groups
+        └── decode LWS root CPG → per-replica groups
+```
+
+Slice topology and `minGroupCount` can require both role subtrees; each LWS
+root supplies that role's replica minimum. A flat CPG cannot express a
+weighted prefill-to-decode ratio (`minGroupCount` is an unweighted child
+count). Ratio, scale, and rollout semantics belong in a DisaggregatedSet
```
wojtek-t, line 875: We will need more policies in the upstream APIs to support those features. But once we have the foundations, these should be much simpler to add.

keps/666-gang-scheduling-in-lws/README.md
```diff
@@ -0,0 +1,1335 @@
...
+- in Phase 2, each CPG's parent link and the complete root-to-leaf chain;
+- `PodGroup.status.conditions[type=PodGroupInitiallyScheduled]`, as an
+  initial-placement signal only;
+- pod events and `spec.schedulingGroup`;
+- LWS events and a new `WorkloadSchedulingReady` condition.
+
+`PodGroupInitiallyScheduled` is a terminal initial-placement signal: once
+True it does not revert, even if members are later evicted. Reusing the
+replica PodGroup across a leader restart can therefore leave it True while
+the replacement generation is Pending. It is not replica health.
```
wojtek-t, line 901: We were chatting about having runtime-oriented condition too. It was removed from the initial scope, but it's something that is still planned. So stay tuned.

wojtek-t (commented):
Great KEP!

I added couple comments/questions but these are pretty minor - I'm overall very supportive the KEP in the current shape.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: I see some use cases in community like volcano/kthena: prioritize decode over prefill when scaling up. https://github.com/volcano-sh/kthena/pull/1227. Not sure if this is a general case.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Yes.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Updated.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Updated according to the job example.

keps/666-gang-scheduling-in-lws/README.md
pacoxu: At `replicas: 0`, whole-LWS minCount would be `replicas * size = 0`. The API cannot store 0, so the unused template keeps `minCount: 1` as a placeholder (not a scheduling quota) and LWS creates no PodGroup or CPG. That `1` is unrelated to replica index `0`.

keps/666-gang-scheduling-in-lws/README.md
wojtek-t: The problem that I have with this usecase is that it's not aligned with the current semantics of the priority that is used not only for scheduling, but also for preemption.
Because technically the more decodes I add, the more prefills I may need to preempt in this usecase (which isn't necessarily what I want).

I think this usecase isn't necessarily about different priority, but rather "relatively priority *within* a single workload".

So my point is that, supporting different priorities within workload would not solve this usecase (arguably it would cause more pain than gain...)

wojtek-t:
Thanks a lot for applying the feedback.
This LGTM.

yankay:
Thanks @pacoxu 
/lgtm

keps/666-gang-scheduling-in-lws/README.md
pacoxu: Agree. The kthena example is autoscaler scale-up ordering (decode before prefill), not PriorityClass.

Different priorities inside one Workload would also enable preemption, which that use case does not want.

So this is not a good motivation for mixed-role priorities. LWS will keep one Workload-wide priorityClassName.

Let me collect more feedback from the ecosystem like Dynamo, LLM-d and other AI serving or training projects that use LWS or similar logics.

pacoxu: After checking Dynamo, Kthena, and RBG: no clear use case for different priorities within one Workload. The PD “decode first” case is autoscaler ordering; Dynamo/Grove prefer ratio/weight for balance in the future, and intra-workload priority mainly adds preemption complexity. I think LWS should keep one Workload-wide priorityClassName for now.

wojtek-t: Thanks a lot for looking into it - the is consistent with my current understanding of no clear usecases for different priorities within a single workload.

I acknowledge the request for ordered scale-up though adding more policies is something I would like to happen (though it will not fit 1.38 timeframe).

ahg-g:
@pacoxu pls fix the failures.

/approve

pacoxu:
Rebased with latest main branch to see if the test can pass.

PR #982, opened 2026-08-14
KEP-666: Phase 1 Workload-aware gang scheduling

pacoxu:
## Summary
Phase 1 of [KEP-666](https://github.com/kubernetes-sigs/lws/tree/main/keps/666-gang-scheduling-in-lws): add alpha `spec.scheduling` and compile one active LWS level into a Kubernetes 1.37 Workload + flat PodGroups.

- Gated by `WorkloadAwareScheduling` (default off). Select provider with `gangSchedulingManagement.schedulerProvider`.
- Empty `spec.scheduling: {}` defaults to one gang PodGroup per replica (`minCount = size`).
- `kubernetes` provider owns Workload/PodGroups and creates them before pods. Legacy Volcano is unchanged when `spec.scheduling` is unset.
- Admission: exactly one level, immutable hierarchy, `LeaderReady` + mixed-role gang rejected.

## Out of scope
CompositePodGroup, DisaggregatedSet as parent, Hash identity, same-revision leader-recreation isolation, kubernetes e2e.

## Test plan
- [x] unit: lowering, validation, kubernetes provider lifecycle
- [x] integration: replica Workload/PodGroup before STS; delegated template; whole-LWS scale to 0
- [ ] Hash + `spec.scheduling` (reject or implement)
- [ ] kubernetes e2e with `GenericWorkload=true`


api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -177,6 +183,92 @@ type LeaderWorkerSetSpec struct {
...
+	// +optional
+	Leader *LeaderWorkerSetLeaderScheduling `json:"leader,omitempty"`
+
+	// worker defines scheduling for the worker PodGroup.
+	// +optional
+	Worker *LeaderWorkerSetWorkerScheduling `json:"worker,omitempty"`
+}
+
+// LeaderWorkerSetLeaderScheduling defines scheduling for the leader PodGroup.
+type LeaderWorkerSetLeaderScheduling struct {
```
ahg-g, line 253: Do we really need to define distinct types for each of the leader and worker?

pkg/schedulerprovider/volcano_provider.go
```diff
@@ -46,6 +49,58 @@ func NewVolcanoProvider(client client.Client) *VolcanoProvider {
 	}
 }
 
+// ReconcileScheduling pre-creates LWS-owned PodGroups for spec.scheduling.
+func (v *VolcanoProvider) ReconcileScheduling(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, replicas int32, revision string) error {
```
ahg-g, line 53: Without this PR, how did we create and manage volcano resources?

pkg/schedulerprovider/volcano_provider.go
```diff
@@ -46,6 +49,58 @@ func NewVolcanoProvider(client client.Client) *VolcanoProvider {
...
+	if lws.Spec.Scheduling == nil {
+		return nil
+	}
+	if errs := ValidatePhaseOneWorkload(ctx, nil, lws); len(errs) > 0 {
```
ahg-g: Validation should be done in the webhook upon create/update events, not after the fact. Why do we validate again here?

pkg/schedulerprovider/phase_one.go
```diff
@@ -0,0 +1,561 @@
...
+			if !matched {
+				allErrs = append(allErrs, field.Invalid(path.Index(i), claim.Name, "must have a matching reference in every member pod template"))
+			}
+		}
+	}
+	return allErrs
+}
+
+func buildFlatWorkload(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet) (*schedulingv1beta1.Workload, error) {
+	if errs := ValidatePhaseOneWorkload(ctx, nil, lws); len(errs) > 0 {
```
ahg-g: Why do we validate again here? wasn't the validation done already by the webhook?

pkg/schedulerprovider/phase_one.go
```diff
@@ -0,0 +1,561 @@
...
+	leaderworkerset "sigs.k8s.io/lws/api/leaderworkerset/v1"
+)
+
+// SchedulingMode is the LWS hierarchy level materialized as PodGroups.
+type SchedulingMode string
+
+const (
+	SchedulingModeLWS     SchedulingMode = "lws"
+	SchedulingModeReplica SchedulingMode = "replica"
+	SchedulingModeRole    SchedulingMode = "role"
```
ahg-g: pls document each of them and the associated semantics

pkg/schedulerprovider/phase_one.go
```diff
@@ -0,0 +1,561 @@
...
+		},
+		AllowedDisruptionModes: []workloadbuilder.DisruptionModeOption{
+			workloadbuilder.SingleMode,
+			workloadbuilder.AllMode,
+		},
+	}
+}
+
+// SchedulingModeFor selects the active scheduling level. Empty scheduling and an
+// explicitly empty replica node both select replica mode.
```
ahg-g: pls document that only one mode at most is supposed to be active otherwise an error is returned

pkg/schedulerprovider/volcano_provider.go
```diff
@@ -46,6 +49,58 @@ func NewVolcanoProvider(client client.Client) *VolcanoProvider {
 	}
 }
 
+// ReconcileScheduling pre-creates LWS-owned PodGroups for spec.scheduling.
```
ahg-g, line 52: do we have an integration test covering this case?

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -130,6 +150,93 @@ func (r *LeaderWorkerSetWebhook) ValidateUpdate(ctx context.Context, oldLws, new
...
+		return nil
+	}
+
+	var allErrs field.ErrorList
+	// Disabling the gate prevents new opt-ins, but existing scheduled objects
+	// must remain updateable so they can scale down and be deleted safely.
+	if !r.FeatureGates.Enabled(features.WorkloadAwareScheduling) && (oldLws == nil || oldLws.Spec.Scheduling == nil) {
+		allErrs = append(allErrs, field.Forbidden(path, "requires the WorkloadAwareScheduling feature gate"))
+	}
+	if r.SchedulerProvider == "" {
```
ahg-g, line 171: can we default this to kubernetes in `Default`?

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -130,6 +150,93 @@ func (r *LeaderWorkerSetWebhook) ValidateUpdate(ctx context.Context, oldLws, new
...
+		allErrs = append(allErrs, field.Required(path, "requires a configured scheduler provider"))
+	} else if r.SchedulerProvider == schedulerprovider.Volcano {
+		allErrs = append(allErrs, validateVolcanoScheduling(lws, path)...)
+	} else if r.SchedulerProvider != schedulerprovider.Kubernetes {
+		allErrs = append(allErrs, field.NotSupported(path, r.SchedulerProvider, []string{string(schedulerprovider.Kubernetes), string(schedulerprovider.Volcano)}))
+	}
+	if r.SchedulerProvider == schedulerprovider.Kubernetes && r.RESTMapper != nil {
+		for _, resource := range []string{"Workload", "PodGroup"} {
```
ahg-g, line 180: pls document that here we validate that the upstream kubernetes scheduling apis are enabled

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -130,6 +150,93 @@ func (r *LeaderWorkerSetWebhook) ValidateUpdate(ctx context.Context, oldLws, new
...
+			if _, err := r.RESTMapper.RESTMapping(schema.GroupKind{Group: schedulingv1beta1.GroupName, Kind: resource}, schedulingv1beta1.SchemeGroupVersion.Version); err != nil {
+				allErrs = append(allErrs, field.Forbidden(path, fmt.Sprintf("scheduling.k8s.io/v1beta1 %s API is not available: %v", resource, err)))
+			}
+		}
+	}
+	if r.SchedulerProvider == schedulerprovider.Kubernetes {
+		templateName, delegated := lws.Annotations[schedulerprovider.GroupTemplateNameAnnotation]
```
ahg-g, line 192: pls document what exactly we are validating here. Also document if these annotations are something the user could set.

pkg/schedulerprovider/phase_one.go
```diff
@@ -0,0 +1,561 @@
...
+	}
+
+	var allErrs field.ErrorList
+	var oldMode SchedulingMode
+	if oldLWS != nil && oldLWS.Spec.Scheduling != nil {
+		oldMode, err = SchedulingModeFor(oldLWS)
+		if err == nil && oldMode != mode {
+			allErrs = append(allErrs, field.Forbidden(path, "cannot switch the active scheduling level after creation"))
+		}
+		if oldPriority, oldConsistent := workloadPriorityClassName(oldLWS); oldConsistent {
```
ahg-g, line 352: This is validating that priority can't be changed only in the case the leader and worker have the same priority class, otherwise changing it is allowed if the leader and worker set different priorities, is this intended?

pkg/schedulerprovider/phase_one.go
```diff
@@ -0,0 +1,561 @@
...
+func workloadPriorityClassName(lws *leaderworkerset.LeaderWorkerSet) (string, bool) {
+	workerPriority := lws.Spec.LeaderWorkerTemplate.WorkerTemplate.Spec.PriorityClassName
+	if lws.Spec.LeaderWorkerTemplate.LeaderTemplate == nil {
+		return workerPriority, true
+	}
+	leaderPriority := lws.Spec.LeaderWorkerTemplate.LeaderTemplate.Spec.PriorityClassName
+	return leaderPriority, leaderPriority == workerPriority
+}
+
+func validatePhaseOneSemantics(lws *leaderworkerset.LeaderWorkerSet, mode SchedulingMode) field.ErrorList {
```
ahg-g, line 406: pls document phase one semantics so one can check the code against the intended behavior.

pkg/schedulerprovider/phase_one.go
ahg-g: it seems we validate that the priority for both leader and worker should be the same, pls document this here.

pkg/schedulerprovider/phase_one.go
```diff
@@ -0,0 +1,561 @@
...
+	schedulingv1beta1 "k8s.io/api/scheduling/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/component-helpers/scheduling/schedulingv1/workloadbuilder"
+	"k8s.io/utils/ptr"
+
+	leaderworkerset "sigs.k8s.io/lws/api/leaderworkerset/v1"
+)
+
+// SchedulingMode is the LWS hierarchy level materialized as PodGroups.
```
ahg-g, line 35: can we add unit tests that cover this file?

pkg/schedulerprovider/interface.go
```diff
@@ -28,15 +29,54 @@ import (
...
 )
 
 const (
-	PodGroupNameFmt = "%s-%s-%s"
+	PodGroupNameFmt     = "%s-%s-%s"
+	RolePodGroupNameFmt = "%s-%s-%s-%s"
+	LWSGroupNameFmt     = "%s-lws"
+
+	GroupTemplateNameAnnotation       = "scheduling.k8s.io/group-template-name"
+	ParentCompositePodGroupAnnotation = "scheduling.k8s.io/parent-compositepodgroup"
```
ahg-g, line 50: pls document what sets these annotations, the user or the controller?

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -177,6 +188,20 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	if lws.Spec.Scheduling != nil {
+		if r.SchedulerProvider == nil {
+			err := fmt.Errorf("spec.scheduling requires a configured scheduler provider")
+			return ctrl.Result{}, r.failWorkloadScheduling(ctx, lws, schedulerprovider.ReasonUnsupportedProviderCapability, err)
+		}
+		if err := r.SchedulerProvider.ReconcileScheduling(ctx, lws, replicas, revisionutils.GetRevisionKey(revision)); err != nil {
+			log.Error(err, "Reconciling workload-aware scheduling prerequisites")
+			return ctrl.Result{}, r.failWorkloadScheduling(ctx, lws, schedulerprovider.ReconcileErrorReason(err), err)
+		}
+		if err := r.updateWorkloadSchedulingCondition(ctx, lws, metav1.ConditionTrue, "SchedulingPrerequisitesReady", "Workload and PodGroups are ready"); err != nil {
```
ahg-g: what does it mean that workload and podgroups are "ready"? also, does this apply to valcano?

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -464,6 +556,10 @@ const (
 	// is true when the lws is in upgrade process after the (leader/worker) template is updated. If only replicas is modified, it will
 	// not be considered as UpdateInProgress.
 	LeaderWorkerSetUpdateInProgress LeaderWorkerSetConditionType = "UpdateInProgress"
+
+	// LeaderWorkerSetWorkloadSchedulingReady reports whether all upstream
+	// scheduling prerequisites for the current LWS generation exist.
```
ahg-g: Ready has a stronger meaning than "exist/created", is "WorkloadSchedulingReady" as a condition type also used by the Job API?

pkg/schedulerprovider/kubernetes_provider.go
```diff
@@ -0,0 +1,646 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	leaderworkerset "sigs.k8s.io/lws/api/leaderworkerset/v1"
+)
+
+const (
+	// Kubernetes identifies the upstream scheduling.k8s.io provider.
+	Kubernetes ProviderType = "kubernetes"
+
+	WorkloadSchedulingAnnotationKey = "leaderworkerset.sigs.k8s.io/workload-aware-scheduling"
```
ahg-g, line 51: pls document each of these annotation/label keys and who is expected to set them and when.

pkg/schedulerprovider/kubernetes_provider.go
```diff
@@ -0,0 +1,646 @@
...
+		meta.OwnerReferences = refs
+		return true
+	}
+	meta.OwnerReferences = append(meta.OwnerReferences, desired)
+	return true
+}
+
+func (p *KubernetesProvider) reconcileWorkload(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet) (*schedulingv1beta1.Workload, error) {
+	if templateName := lws.Annotations[GroupTemplateNameAnnotation]; templateName != "" {
+		workload, err := p.findDelegatedWorkload(ctx, lws)
```
ahg-g, line 383: what does delegated workload mean?

pkg/schedulerprovider/kubernetes_provider.go
```diff
@@ -0,0 +1,646 @@
...
+		}
+		owner = metav1.GetControllerOf(parent)
+	}
+	if selected == nil {
+		return nil, fmt.Errorf("no parent Workload matches the LWS controller-owner chain")
+	}
+	return selected, nil
+}
+
+func updateMutableWorkloadFields(ctx context.Context, c client.Client, current, desired *schedulingv1beta1.Workload) error {
```
ahg-g, line 523: pls document the semantics of this function

pkg/schedulerprovider/kubernetes_provider.go
```diff
@@ -0,0 +1,646 @@
...
+
+	persisted, err := p.findOwnedWorkload(ctx, lws)
+	if err != nil {
+		return nil, workloadAPIError(ReasonWorkloadCreateFailed, err)
+	}
+	if persisted != nil {
+		if persisted.Name != desiredWorkload.Name {
+			return nil, NewReconcileError(ReasonInvalidSchedulingConfiguration, fmt.Errorf("owned Workload %s/%s does not have the expected UID-qualified name %q", persisted.Namespace, persisted.Name, desiredWorkload.Name))
+		}
+		if err := updateMutableWorkloadFields(ctx, p.client, persisted, desiredWorkload); err != nil {
```
ahg-g, line 408: pls document what we are trying to update here, why would the workload we got not have the expected values

pkg/schedulerprovider/kubernetes_provider.go
```diff
@@ -0,0 +1,646 @@
...
+			return nil, NewReconcileError(ReasonInvalidSchedulingConfiguration, err)
+		}
+		return persisted, nil
+	}
+
+	if err := p.client.Create(ctx, desiredWorkload); err == nil {
+		return desiredWorkload, nil
+	} else if !apierrors.IsAlreadyExists(err) {
```
ahg-g, line 416: `findOwnedWorkload` should have already handled the case of the workload existing, and so here even if the workload already exists was true (because of some race condition), if we return an error, another reconcile will kick in and `findOwnedWorkload` will find it and successfully reconcile eventually; all this to say that the logic below is not needed I think.

pkg/schedulerprovider/kubernetes_provider.go
```diff
@@ -0,0 +1,646 @@
...
+type KubernetesProvider struct {
+	client client.Client
+}
+
+func NewKubernetesProvider(c client.Client) *KubernetesProvider {
+	return &KubernetesProvider{client: c}
+}
+
+// ReconcileScheduling enforces Workload -> PodGroup -> Pod creation order.
+func (p *KubernetesProvider) ReconcileScheduling(ctx context.Context, lws *leaderworkerset.LeaderWorkerSet, replicas int32, revision string) error {
```
ahg-g, line 139: pls document the logic in bullet points, summarize what ReconcileScheduling aims to achieve.

pkg/schedulerprovider/kubernetes_provider.go
```diff
@@ -0,0 +1,646 @@
...
+}
+
+type runtimePodGroup struct {
+	name                string
+	templateName        string
+	labels              map[string]string
+	allowMinCountUpdate bool
+}
+
+func phaseOneRuntimeGroups(lws *leaderworkerset.LeaderWorkerSet, replicas int32, revision string) ([]runtimePodGroup, error) {
```
ahg-g: nit: the name "runtimeGroup" is confusing

pkg/schedulerprovider/kubernetes_provider.go
```diff
@@ -0,0 +1,646 @@
...
+			templateName:        lwsWorkloadTemplateName,
+			labels:              baseLabels(mode),
+			allowMinCountUpdate: true,
+		}}, nil
+	case SchedulingModeReplica:
+		groups := make([]runtimePodGroup, 0, replicas)
+		for groupIndex := int32(0); groupIndex < replicas; groupIndex++ {
+			index := strconv.FormatInt(int64(groupIndex), 10)
+			labels := baseLabels(mode)
+			labels[leaderworkerset.GroupIndexLabelKey] = index
```
ahg-g, line 251: We recently supported a new mode of deployment where the groups are not associated with a specific ordinal index. The parameter in LWS to enable that mode of deployment is called [GroupIdentity](https://github.com/kubernetes-sigs/lws/blob/main/api/leaderworkerset/v1/leaderworkerset_types.go#L177); when `GroupIdentity` is `hash`, the leaders are created using a Deployment not a StatefulSet and the LWS replicas don't have an index. We should support this mode of deployment here.

ahg-g (commented):
This is great!

A couple of high-level comments:
1) Please ensure we have good documentation of the logic (most of the detailed comments are related to this)
2) Please ensure we have good unit and integration tests coverage.

ahg-g:
@pacoxu  do you think we can get the mostly documentation related comments resolved today/tomorrow to catch the release (planned this week)?

api/leaderworkerset/v1/leaderworkerset_types.go
pacoxu: We split them for a future role-specific field. For current struct, we can merge if we want a smaller API.

pkg/schedulerprovider/volcano_provider.go
pacoxu: https://github.com/kubernetes-sigs/lws/tree/main/keps/407-gang-scheduling
IIUC, webhook stamps group-name, pod controller creates a Volcano PodGroup owned by the leader. Unchanged when `spec.scheduling` is unset. New path only runs for typed `spec.scheduling`.

pkg/webhooks/leaderworkerset_webhook.go
pacoxu: This is a controller configurition.

https://github.com/kubernetes-sigs/lws/blob/f841942ef07e37d1dbf1cdaac1997823d5d77cbb/pkg/config/validation.go#L45-L48

I prefer to keep the current behavior, as it is not a new LWS field.

pkg/schedulerprovider/kubernetes_provider.go
pacoxu: Good catch. I miss this.  

Currently,  this PR now rejects `spec.scheduling` with `groupIdentity: Hash` at admission.

I may add the support tomorrow either as a follow-up commit here or a separate PR.

api/config/v1alpha1/configuration_types.go
```diff
@@ -37,6 +37,11 @@ type Configuration struct {
 	// GangSchedulingManagement is configuration for gang scheduling management.
 	GangSchedulingManagement *GangSchedulingManagement `json:"gangSchedulingManagement,omitempty"`
 
+	// FeatureGates is the set of alpha or beta LWS features enabled for the
+	// controller manager. Unknown feature gates are rejected during startup.
+	// +optional
+	FeatureGates map[string]bool `json:"featureGates,omitempty"`
```
yankay, line 44: Thanks @pacoxu 

Could we follow JobSet and Kueue by using `k8s.io/component-base/featuregate` to manage feature gates? This would reuse Kubernetes’ standard validation and versioning mechanisms.

pkg/schedulerprovider/phase_one.go
```diff
@@ -0,0 +1,561 @@
...
+		}},
+	}
+}
+
+// ValidatePhaseOneWorkload validates spec.scheduling.
+func ValidatePhaseOneWorkload(ctx context.Context, oldLWS, lws *leaderworkerset.LeaderWorkerSet) field.ErrorList {
+	path := field.NewPath("spec", "scheduling")
+	mode, err := SchedulingModeFor(lws)
+	if err != nil {
+		return field.ErrorList{field.Invalid(path, lws.Spec.Scheduling, err.Error())}
```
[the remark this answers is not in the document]
ahg-g: pls address to make sure we clearly reject hash based until the followup PR.

api/leaderworkerset/v1/leaderworkerset_types.go
ahg-g: distinct is fine, I was just curious

ahg-g (commented):
Thanks, all good from my end, we can followup with a separate PR for hash mode support. Please see the last comment from @yankay

aliceco01 (commented):
Two issues in the kubernetes provider, details inline.

pkg/schedulerprovider/kubernetes_provider.go
```diff
@@ -0,0 +1,659 @@
...
+	desiredOwner := metav1.NewControllerRef(lws, leaderworkerset.GroupVersion.WithKind("LeaderWorkerSet"))
+	for i := range groups.Items {
+		group := &groups.Items[i]
+		if !controllerReferencesEqual(metav1.GetControllerOf(group), desiredOwner) {
+			continue
+		}
+		if _, keep := desired[group.Name]; keep {
+			continue
+		}
+		if _, inUse := inUseGroups[group.Name]; !inUse {
```
aliceco01, line 612: An old-revision PodGroup can be deleted here while the leader StatefulSet still needs it, and nothing recreates it. This is separate from the same-revision isolation item under Out of scope: here the group is gone, not shared.

Sequence, replica mode with the default restartPolicy:

1. A rollout is in progress, or `rollingUpdateConfiguration.partition` is set, so replica 0 stays at revision A while `desiredPodGroups` only covers revision B.
2. Replica 0 restarts. The leader is deleted with foreground propagation, so for a moment no pod references the replica 0 PodGroup for revision A.
3. An LWS reconcile in that window deletes it here. Reconciles do fire around then (the worker StatefulSet deletion enqueues one), so this is timing-dependent.
4. The leader StatefulSet recreates ordinal 0 at revision A, since ordinals below the partition keep the old revision, and the pod webhook points it at the deleted group.
5. Nothing recreates that group, so the leader and its workers wait in PreEnqueue for a PodGroup that never appears. During an LWS-driven rollout that lasts until the rollout reaches replica 0, and the replica counts against maxUnavailable meanwhile. With a user-set partition it lasts until the partition moves.

The trigger is a race, but the end state is deterministic: replaying the provider calls in a unit test against c76a58b leaves the referenced PodGroup NotFound.

One possible fix: `CreatePodGroupIfNotExists` already runs for every leader before the worker StatefulSet is created, and is a no-op for this provider. It could create the leader's referenced PodGroup from the Workload template when it is missing, which would also cover a manual delete. I can do that as a follow-up with the test if you would rather not hold the release for it.

pkg/schedulerprovider/kubernetes_provider.go
```diff
@@ -0,0 +1,659 @@
...
+				if selected != nil {
+					return nil, fmt.Errorf("multiple parent Workloads match the LWS controller-owner chain: %s/%s and %s/%s", selected.Namespace, selected.Name, candidate.Namespace, candidate.Name)
+				}
+				selected = candidate
+			}
+		}
+
+		parent := &unstructured.Unstructured{}
+		parent.SetGroupVersionKind(schema.GroupVersionKind{Group: gv.Group, Version: gv.Version, Kind: owner.Kind})
+		if err := p.client.Get(ctx, types.NamespacedName{Namespace: lws.Namespace, Name: owner.Name}, parent); err != nil {
```
aliceco01, line 510: Two RBAC gaps in delegated mode. The integration test cannot show them, because envtest runs with admin credentials:

1. This walk GETs every owner in the chain, even after a matching Workload was found at the first hop. The ClusterRole only covers LWS's own kinds and a few built-ins, so for a third-party parent (the `ParentJob` in the unit test, or JobSet or TrainJob) this returns Forbidden, `ReconcileScheduling` fails with `ParentWorkloadNotReady`, and the reconcile returns before the leader StatefulSet is created. The integration test uses a ControllerRevision parent, which LWS can read. Stopping at the first hop that matches would avoid it.
2. CompositePodGroup is listed as out of scope, but the webhook accepts the parent annotation and the check at line 160 reads a CompositePodGroup through the cached client, with no `compositepodgroups` rule in the ClusterRole. A cached Get waits for its informer to sync, which cannot happen without list and watch. With no reconcile timeout and one worker, one LWS that gets this far would stall reconciliation of every LWS. Rejecting the annotation until CompositePodGroup is supported would close it; so would adding get, list and watch to the kubebuilder marker, or reading that object with the API reader.

Neither affects an LWS without the delegation annotations.

pkg/schedulerprovider/phase_one.go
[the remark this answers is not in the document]
pacoxu: This should be addressed https://github.com/kubernetes-sigs/lws/pull/982/changes#diff-615863b8f48cf0a300241b0414555a55219ad704a563968c6c7aeb90d87c32e3R331 in `func ValidateGroupIdentity()`.

api/config/v1alpha1/configuration_types.go
pacoxu: Updated: https://github.com/kubernetes-sigs/lws/pull/982/changes/ec6e6f8a0c2af83430722f73b0eea4cdc9b408e7#diff-4914632202b9c7ee5d0ba5f8d143ba5a91887e66c02f35e1456e912092a9346dR28-R30.

pkg/schedulerprovider/kubernetes_provider.go
pacoxu: > Two RBAC gaps in delegated mode. The integration test cannot show them, because envtest runs with admin credentials:
> 
> 1. This walk GETs every owner in the chain, even after a matching Workload was found at the first hop. The ClusterRole only covers LWS's own kinds and a few built-ins, so for a third-party parent (the `ParentJob` in the unit test, or JobSet or TrainJob) this returns Forbidden, `ReconcileScheduling` fails with `ParentWorkloadNotReady`, and the reconcile returns before the leader StatefulSet is created. The integration test uses a ControllerRevision parent, which LWS can read. Stopping at the first hop that matches would avoid it.

Updated.

> 2. CompositePodGroup is listed as out of scope, but the webhook accepts the parent annotation and the check at line 160 reads a CompositePodGroup through the cached client, with no `compositepodgroups` rule in the ClusterRole. A cached Get waits for its informer to sync, which cannot happen without list and watch. With no reconcile timeout and one worker, one LWS that gets this far would stall reconciliation of every LWS. Rejecting the annotation until CompositePodGroup is supported would close it; so would adding get, list and watch to the kubebuilder marker, or reading that object with the API reader.

Removed CPG related code in Phase 1 and keep some todo comments .

pkg/schedulerprovider/kubernetes_provider.go
pacoxu: > can do that as a follow-up with the test if you would rather not hold the release for it.

Great. I think we can fix this in a follow up PR.

PR #983, opened 2026-08-14
Verify controller ownership before GetForRole/Scale/legacy-slice0 mutate an LWS

### LINKED ISSUE #981: DisaggregatedSet: GetForRole/Scale can mutate a foreign LWS occupying the generated name during rapid same-name recreation
/kind bug

**What happened**:

Raised by @yankay during review of #933: `LeaderWorkerSetManager.GetForRole` and `Scale` look up an LWS by its generated name only (`GenerateName(ds.Name, slice, revision, role)` / the legacy name for slice 0), with no check that the found LWS is actually controller-owned by the `DisaggregatedSet` calling them:

```go
func (manager *LeaderWorkerSetManager) GetForRole(ctx context.Context, ds *disaggregatedsetv1.DisaggregatedSet, slice int, revision, role string) (*leaderworkersetv1.LeaderWorkerSet, error) {
	lws, err := manager.Get(ctx, ds.Namespace, disaggregatedsetutils.GenerateName(ds.Name, slice, revision, role))
	...
}

func (manager *LeaderWorkerSetManager) Scale(ctx context.Context, namespace, name string, replicas int) error {
	leaderWorkerSet := &leaderworkersetv1.LeaderWorkerSet{}
	if err := manager.client.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, leaderWorkerSet); err != nil {
	...
}
```

This is the same class of issue #933 fixed for `LeaderWorkerSetManager.List` (used by status aggregation) via `metav1.IsControlledBy` — but `GetForRole`/`Scale` are a separate code path (used by `reconcileRoleSimple`, the create/scale path) that wasn't covered by that fix.

**How to reproduce it**:

1. Create a `DisaggregatedSet` named `foo`.
2. Delete it.
3. Immediately (before the garbage collector removes `foo`'s owned LWS objects — asynchronous, not instant) create a *new* `DisaggregatedSet` also named `foo`.
4. The new controller's `reconcileRoleSimple` calls `GetForRole`, which finds the *old* `foo`'s still-present LWS (same generated name), and may then call `Scale` on it — mutating an LWS that is not actually owned by the new `DisaggregatedSet`.

`GetForRole`/`Scale` should only treat an LWS as "the existing one for this role" if it's actually controller-owned by the calling `DisaggregatedSet` (`metav1.IsControlledBy`), the same way `List` now does. A name match alone isn't sufficient during this race window.

- Marked non-blocking/rare by @yankay when approving #933 — tracked here as a follow-up rather than blocking that PR.
- Reference: https://github.com/kubernetes-sigs/lws/pull/933#issuecomment-5265580514

BenAyedMedAla:
/kind bug

Fixes #981, raised by @yankay during review of #933.

`LeaderWorkerSetManager.GetForRole` and `Scale` looked up an LWS by its generated name only, with no check that the found object is actually controller-owned by the `DisaggregatedSet` calling them. During rapid same-name recreation — a `DisaggregatedSet` named `foo` is deleted and a new one also named `foo` is created before garbage collection removes the old one's LWS objects — the new controller's `reconcileRoleSimple` could find and then `Scale` the *old* `foo`'s still-present LWS, mutating an object it doesn't own. This is the same class of bug #933 fixed for `List` via `metav1.IsControlledBy`; `GetForRole`/`Scale` are a separate code path that fix didn't cover.

While fixing this I found the identical gap in `recreateLegacySlice0`: it fetched a legacy-named LWS by a raw `Get` with no ownership check before deleting it to make way for the slice-aware migration — same root cause, different call site, so I fixed it too rather than leaving an equally real instance of the same bug next to the one the issue names.

#### What changed

- Added `LeaderWorkerSetManager.getOwned`, an ownership-filtered wrapper around `Get` (mirrors `List`'s existing filtering). `GetForRole` now uses it for both the slice-aware and legacy-name lookups; a foreign-owned LWS at either name is treated as absent rather than returned for the caller to mutate.
- `Scale` now takes the `DisaggregatedSet` (instead of a bare namespace) and refuses to patch a same-named LWS that isn't controller-owned by it, returning an error instead of silently mutating a foreign object. Updated all three call sites (`reconcileRoleSimple`, `scaleUpNew`, and the scale-down path in `executor.go`).
- `recreateLegacySlice0` uses `getOwned` instead of a raw `Get` before deleting a legacy-named LWS.

Fixes #981

- Test fixtures across `lws_manager_test.go` and `executor_test.go` that built LWS objects without a matching controller `OwnerReference` needed updating to carry one (matching what the real `Create()` path already always sets), since `Scale`'s new ownership check depends on it. I found these by running the full suite after the change and fixing whatever broke — none of the fixture changes reflect a behavior change, only making the fixtures match what production objects actually look like.
- Added regression tests: `GetForRole`/`Scale`/`recreateLegacySlice0` each ignore a same-named LWS owned by a different `DisaggregatedSet`, and normal operation is unaffected when ownership is intact.
- No API surface change, no CRD/docs regen needed.

```release-note
DisaggregatedSet: GetForRole, Scale, and the legacy slice-0 migration path now verify controller ownership before reading or mutating a same-named LeaderWorkerSet, closing a rare race where a rapidly recreated DisaggregatedSet could mutate a still-draining LWS left over from a previous DisaggregatedSet of the same name.
```

### DIFF
--- a/pkg/controllers/disaggregatedset/lws_manager.go
+++ b/pkg/controllers/disaggregatedset/lws_manager.go
@@ -99,10 +99,24 @@ func (manager *LeaderWorkerSetManager) Create(ctx context.Context, params disagg
 	}
 
 	if err := manager.client.Create(ctx, leaderWorkerSet); err != nil {
-		if apierrors.IsAlreadyExists(err) {
-			return nil
+		if !apierrors.IsAlreadyExists(err) {
+			return fmt.Errorf("failed to create LeaderWorkerSet %s: %w", lwsName, err)
 		}
-		return fmt.Errorf("failed to create LeaderWorkerSet %s: %w", lwsName, err)
+		// Name is taken. If we already own it, a concurrent reconcile of this
+		// same DisaggregatedSet beat us to it — no-op. If it's foreign-owned
+		// (see #981), error instead of silently no-oping: this DS's watches
+		// won't fire again for a foreign object it doesn't own, so a silent
+		// return here could leave the role permanently missing an LWS until
+		// some unrelated trigger causes another reconcile. Returning an error
+		// requeues instead.
+		existing := &leaderworkersetv1.LeaderWorkerSet{}
+		if getErr := manager.client.Get(ctx, types.NamespacedName{Name: lwsName, Namespace: params.DisaggregatedSet.Namespace}, existing); getErr != nil {
+			return fmt.Errorf("failed to get existing LeaderWorkerSet %s after create conflict: %w", lwsName, getErr)
+		}
+		if !metav1.IsControlledBy(existing, params.DisaggregatedSet) {
+			return fmt.Errorf("LeaderWorkerSet %s exists but is not controlled by DisaggregatedSet %s; refusing to adopt it", lwsName, params.DisaggregatedSet.Name)
+		}
+		return nil
 	}
 
 	log := logf.FromContext(ctx)
@@ -110,10 +124,18 @@ func (manager *LeaderWorkerSetManager) Create(ctx context.Context, params disagg
 	return nil
 }
 
-func (manager *LeaderWorkerSetManager) Scale(ctx context.Context, namespace, name string, replicas int) error {
+// Scale patches the LWS named name to replicas, but only if it's actually
+// controller-owned by ds. A same-named LWS that exists but isn't owned by ds
+// — e.g. left over from a same-named DisaggregatedSet that was deleted and
+// recreated before garbage collection ran — is refused rather than mutated;
+// see #981.
+func (manager *LeaderWorkerSetManager) Scale(ctx context.Context, ds *disaggregatedsetv1.DisaggregatedSet, name string, replicas int) error {
 	leaderWorkerSet := &leaderworkersetv1.LeaderWorkerSet{}
-	if err := manager.client.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, leaderWorkerSet); err != nil {
+	if err := manager.client.Get(ctx, types.NamespacedName{Name: name, Namespace: ds.Namespace}, leaderWorkerSet); err != nil {
 		return fmt.Errorf("failed to get LeaderWorkerSet %s for scaling: %w", name, err)
+	}
+	if !metav1.IsControlledBy(leaderWorkerSet, ds) {
+		return fmt.Errorf("LeaderWorkerSet %s exists but is not controlled by DisaggregatedSet %s; refusing to scale it", name, ds.Name)
 	}
 
 	if int(getLWSReplicas(leaderWorkerSet)) == replicas {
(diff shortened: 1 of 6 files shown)


pkg/controllers/disaggregatedset/lws_manager.go
```diff
@@ -110,11 +110,19 @@ func (manager *LeaderWorkerSetManager) Create(ctx context.Context, params disagg
 	return nil
 }
 
-func (manager *LeaderWorkerSetManager) Scale(ctx context.Context, namespace, name string, replicas int) error {
+// Scale patches the LWS named name to replicas, but only if it's actually
+// controller-owned by ds. A same-named LWS that exists but isn't owned by ds
+// — e.g. left over from a same-named DisaggregatedSet that was deleted and
+// recreated before garbage collection ran — is refused rather than mutated;
+// see #981.
+func (manager *LeaderWorkerSetManager) Scale(ctx context.Context, ds *disaggregatedsetv1.DisaggregatedSet, name string, replicas int) error {
```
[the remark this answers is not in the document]
BenAyedMedAla: Good catch — you're right that it's worse than a rare miss: since this DisaggregatedSet doesn't own the foreign object, its owned-object watches never fire for it, so nothing would prompt another reconcile once GC eventually clears the name. Fixed in f4f5105: `Create`'s `AlreadyExists` handling now fetches the existing object and checks `metav1.IsControlledBy` — a benign same-owner race still no-ops as before, but a foreign-owned name now returns an error so the reconcile requeues. Added a test (`errors when the name is taken by a foreign-owned LWS`) covering it.

pkg/controllers/disaggregatedset/lws_manager.go
ahg-g: List already checks for owned objects, so Get behavior should be consistent. Are there calls to Get that don't assume the LWS is owned by the given disaggset? If not, I suggest to just update Get instead of creating a new `getOwned` function.

BenAyedMedAla: Good call, thanks — done in ae34cf8. `Get` now takes the `DisaggregatedSet` and checks `metav1.IsControlledBy` directly, consistent with `List`; `getOwned` is gone and `GetForRole`/`recreateLegacySlice0` just call `Get`.

Checking every caller of `Get` to make the signature change turned up a fourth instance of the same bug I hadn't caught: `executor.go`'s `ensureNewLWSExists` did the identical unguarded check-then-create against a raw `Get` — a foreign-owned LWS at the generated name would have made it wrongly conclude "already exists, nothing to do," permanently skipping creation for that role. Fixed for free by the signature change, so thanks for pushing on this — it made the fix more complete than my original surgical approach would have.

ahg-g:
/lgtm
/approve

Thanks!

Issue #898, opened 2026-06-25
Support targeted relocation of specific LWS groups


wcwuwc:
**What would you like to be added**:
Support targeted relocation of specific LeaderWorkerSet groups without requiring ordinal-ordered disruption.
Today, LWS group identities are tied to the leader StatefulSet ordinals, so the active groups are implicitly `0..replicas-1`. This makes it difficult to relocate or recreate a middle group while keeping higher ordinal groups running.
For example, if an LWS has groups `0`, `1`, and `2`, it should be possible to relocate or recreate group `1` without deleting, recreating, or otherwise disrupting group `2`.
This could be implemented through one of several possible APIs, such as:
- A way to mark specific groups for relocation or recreation.
- A relocation subresource or annotation, for example `relocateGroups: ["1"]`.
- A future group lifecycle API that allows LWS to operate on selected group IDs instead of only relying on `spec.replicas`.
The key requirement is that LWS should be able to move selected groups while keeping unrelated groups running.

**Why is this needed**:
This is needed for cluster defragmentation, especially in GPU clusters.
GPU inference and training workloads often have high startup cost due to model loading, cache warmup, distributed runtime initialization, and device setup. Unnecessary restarts can be very expensive.
In defragmentation or relocation scenarios, the group that needs to move is usually determined by placement, topology, or resource-fragmentation constraints, not by its ordinal. For example, an external defragmentation controller may need to free a specific node or topology domain by relocating the group currently placed there.
With the current contiguous ordinal model:
- Scaling down removes groups from the highest ordinal first.
- A middle group cannot be removed while keeping higher ordinal groups.
- Recreating or relocating a specific group may require touching unrelated groups.
- This can increase startup and restart cost for GPU workloads.
Using StatefulSet semantics inside each group still makes sense because worker indexes need to be stable. The concern is mainly about group-level lifecycle management across groups. Requiring all groups to follow a contiguous ordinal lifecycle makes relocation less flexible.

**Completion requirements**:
This enhancement requires the following artifacts:
- [ ] Design doc
- [ ] API change
- [ ] Docs update

wcwuwc:
## Motivation

Kubernetes clusters need a way to reduce resource fragmentation that accumulates during runtime.

[...]

PR #986, opened 2026-08-18
[docs] Add hash group identity KEP

BenjaminBraunDev:
Adds KEP documentation for LeaderWorkerSet hash group identity: a new `spec.groupIdentity` field with values `Ordinal` (default to current behavior) and `Hash`, where leader pods are managed by a Deployment instead of a StatefulSet, groups have no ordering, and scale down removes unscheduled and not-ready groups before healthy ones. Includes DisaggregatedSet integration so roles can opt in through their inline LeaderWorkerSet spec, including a rolling Ordinal to Hash migration.

Issue: #898

### DIFF
--- a/keps/898-group-identity/README.md
+++ b/keps/898-group-identity/README.md
@@ -1,0 +1,218 @@
+# KEP-898: Hash Group Identity
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Scale down without collateral churn](#story-1-scale-down-without-collateral-churn)
+    - [Story 2: Relocate one specific group](#story-2-relocate-one-specific-group)
+    - [Story 3: Autoscale serving groups with HPA](#story-3-autoscale-serving-groups-with-hpa)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API](#api)
+  - [Leader Deployment](#leader-deployment)
+  - [Group Identity Assignment](#group-identity-assignment)
+  - [Worker StatefulSets and Leader Address](#worker-statefulsets-and-leader-address)
+  - [Rollouts](#rollouts)
+  - [Scale Subresource and HPA](#scale-subresource-and-hpa)
+  - [Unsupported Combinations](#unsupported-combinations)
+  - [Test Plan](#test-plan)
+    - [Prerequisite testing updates](#prerequisite-testing-updates)
+    - [Unit tests](#unit-tests)
+    - [Integration tests](#integration-tests)
+    - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Alternative 1: Smarter victim selection in the leader StatefulSet](#alternative-1-smarter-victim-selection-in-the-leader-statefulset)
+  - [Alternative 2: Controller-managed bare leader pods](#alternative-2-controller-managed-bare-leader-pods)
+  - [Alternative 3: Solve it in consumers](#alternative-3-solve-it-in-consumers)
+<!-- /toc -->
+
+## Summary
+
+This KEP adds a `groupIdentity` field to the LeaderWorkerSet spec with two values. `Ordinal` is the default and keeps today's behavior, where groups are named by the leader StatefulSet's ordinals. `Hash` gives each group a random identity instead of an ordinal and manages leader pods with a Deployment instead of a StatefulSet.
+
+With hash identity, scale down removes unscheduled and not-ready groups before healthy ones, because that is how the ReplicaSet ranks its victims. Everything else about a group stays the same: workers still run in a per-group StatefulSet, rolling updates still move group by group within the `maxSurge` and `maxUnavailable` budgets, and startup policies and exclusive placement work as they do today.
+
+## Motivation
+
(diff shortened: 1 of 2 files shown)


keps/898-group-identity/README.md
```diff
@@ -0,0 +1,233 @@
...
+)
+```
+
+The webhook defaults an empty value to `Ordinal` and rejects updates to it.
+
+### Leader Deployment
+
+In hash mode, the controller owns a Deployment (named after the LeaderWorkerSet) instead of a leader StatefulSet. `maxSurge` and `maxUnavailable` from the LWS rolling update configuration map directly onto the Deployment strategy.
+
+Leader pods carry a `leaderworkerset.sigs.k8s.io/group-ready` readiness gate. The pod controller sets the condition to true once the group's worker StatefulSet is ready, so a leader counts as ready only when its whole group is. This makes the Deployment's availability budget count groups rather than bare leader pods, which is what paces rollouts group by group and what makes the ReplicaSet prefer broken groups at scale down.
```
ahg-g, line 130: Great use of readiness gates!

keps/898-group-identity/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+### Group Identity Assignment
+
+The pod webhook assigns each new leader pod a group key, a SHA1 over the namespace and a random 16 character string. A random input is required because leader pods are created through `generateName`, so at mutating admission time the pod has no name or UID to derive a key from. The key is stored in both the `group-index` and `group-key` labels on the leader and inherited by its workers. Exclusive placement continues to key off `group-key` exactly as in ordinal mode.
```
ahg-g, line 134: This is unfortunate, ideally we just use the hash in the pod's name since it will be easier to lookup and match by operators, but I guess we have no way around it.

keps/898-group-identity/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+### Worker StatefulSets and Leader Address
+
+Each group's worker StatefulSet is named after its leader pod, as today. Since hash-named leaders have no per-pod DNS record, the controller stamps the leader's pod IP onto worker pods through a `leaderworkerset.sigs.k8s.io/leader-address` annotation, surfaced to containers as the `LWS_LEADER_ADDRESS` environment variable. On the leader itself the variable resolves to its own pod IP through the downward API. Groups of size 1 get no readiness gate and no worker StatefulSet.
```
ahg-g: > hash-named leaders have no per-pod DNS record

Are we sure this is true? why don't they a DNS record? I think they should have one.

Also, note that we can set the hostname and subdomain fields: https://kubernetes.io/docs/concepts/workloads/pods/pod-hostname/#hostname-with-pod-s-hostnameoverride

keps/898-group-identity/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+Because the revision includes the field and DisaggregatedSet rolls template changes by replacing whole LeaderWorkerSets, changing a role from `Ordinal` to `Hash` is a normal rolling update rather than a forbidden in-place mutation. NOTE: the DisaggregatedSet revision covers all roles jointly, so changing one role's identity mode rolls the whole slice, the same as any other role template change.
+
+### Unsupported Combinations
+
+Validation rejects hash mode combined with features whose semantics depend on stable ordinals:
+
+- `subGroupPolicy`
+- volume claim templates
+- `networkConfig.subdomainPolicy: UniquePerReplica`
```
ahg-g: Why not? we can create a headless service with the same name as the leader pod and use that as the subdomain name.

keps/898-group-identity/README.md
```diff
@@ -0,0 +1,233 @@
...
+2. The DisaggregatedSet webhook runs the same hash-mode validation per role, so an unsupported combination fails at DisaggregatedSet admission instead of surfacing later as LeaderWorkerSet creation failures in a reconcile loop.
+3. The DisaggregatedSet revision hash includes the field, normalized so an empty value and the CRD default `Ordinal` hash identically. Objects persisted before the field existed keep their revision when the new CRD starts defaulting it, so upgrading the controller does not roll existing DisaggregatedSets.
+
+Because the revision includes the field and DisaggregatedSet rolls template changes by replacing whole LeaderWorkerSets, changing a role from `Ordinal` to `Hash` is a normal rolling update rather than a forbidden in-place mutation. NOTE: the DisaggregatedSet revision covers all roles jointly, so changing one role's identity mode rolls the whole slice, the same as any other role template change.
+
+### Unsupported Combinations
+
+Validation rejects hash mode combined with features whose semantics depend on stable ordinals:
+
+- `subGroupPolicy`
```
ahg-g: why not? this can be supported, we should still treat the leader as worker 0

ahg-g (commented):
Great details in the design section, good job.

keps/898-group-identity/README.md
BenjaminBraunDev: thanks!

keps/898-group-identity/README.md
BenjaminBraunDev: Actually yeah, the webhook can set pod hostname to the group key it generates at admission and subdomain to the LWS headless service, which hash mode already creates with publishNotReadyAddresses. Then we can make LWS_LEADER_ADDRESS be an actual DNS name like ordinal mode instead of a pod IP. I'll make this update. And we can probably use 16 character prefix of the group key for the DNS facing values to avoid +40 chars in length

keps/898-group-identity/README.md
BenjaminBraunDev: group key will essentially act as the identifier we have access to at admission time and becomes the leader's DNS name, so operators can resolve that.

keps/898-group-identity/README.md
BenjaminBraunDev: Yeah, can make it work with group key

keps/898-group-identity/README.md
BenjaminBraunDev: I looked into this and subdomain is immutable and has to be set at admission when don't have the leader name. However, it can be named after the group key, just like with the hostnames.

ahg-g:
/approve
/lgtm

Great!

BenjaminBraunDev:
rebasing to pull the verify-main fix and I think the pull-lws-test-integration-main test fail was a flake

PR #987, opened 2026-08-18
[feat] Add groupIdentity Hash mode with Deployment managed leaders

### LINKED ISSUE #898: Support targeted relocation of specific LWS groups

BenjaminBraunDev:
### Summary
Implements hash group identity from KEP-898 (PR: #986, Issue: #898). Adds `spec.groupIdentity` with values `Ordinal` (default, unchanged behavior) and `Hash`. In hash mode the controller manages leader pods through a Deployment instead of a StatefulSet:
  - scale down removes unscheduled and not-ready groups before healthy ones (ReplicaSet victim ranking)
  - a `group-ready` readiness gate on leader pods paces rollouts and availability by whole groups
  - groups are identified by random 40 character keys in the group-index and group-key labels
  - workers reach their leader by pod IP through `LWS_LEADER_ADDRESS`

DisaggregatedSet roles opt in through their inline LWS spec, and changing a role from Ordinal to Hash rolls out like normal.

Notes for the reviewer
  - `groupIdentity` is immutable on a LeaderWorkerSet. Validation rejects hash mode combined with volumeClaimTemplates and non-zero rolling update partition. `subGroupPolicy` and `subdomainPolicy: UniquePerReplica` are supported, with their identity derived from the group key instead of the leader pod name.
  - The revision normalization treats an empty value and the CRD default Ordinal identically. Without it, the new CRD default would change every existing DisaggregatedSet revision on upgrade and trigger a mass rollout. Covered by tests.

Testing: unit tests; integration specs for webhook behavior in both APIs. Benchmarking: node failure victim selection in both modes, DisaggregatedSet scale down under load, Ordinal to Hash migration.

Which issue(s) this PR fixes: Fixes #898

### Release Note
```release-note
LeaderWorkerSet supports `groupIdentity: Hash`, which manages leader pods with a Deployment so scale down removes unhealthy groups before healthy ones.
```

### DIFF
--- a/api/leaderworkerset/v1/leaderworkerset_types.go
+++ b/api/leaderworkerset/v1/leaderworkerset_types.go
@@ -109,11 +109,9 @@ const (
 	// can tell the identity scheme apart without fetching the LWS object.
 	GroupIdentityAnnotationKey string = "leaderworkerset.sigs.k8s.io/group-identity"
 
-	// LeaderAddressAnnotationKey carries the leader address on worker pods when
-	// GroupIdentity=Hash: the leader's DNS name derived from the group key, or the
-	// leader pod IP for groups admitted before hostnames were assigned. The address
-	// only needs to be stable for one group generation because the whole group is
-	// recreated together.
+	// LeaderAddressAnnotationKey carries the leader's DNS address on worker pods
+	// in both group identity modes, so that pod admission can inject
+	// LWS_LEADER_ADDRESS without recomputing it.
 	LeaderAddressAnnotationKey string = "leaderworkerset.sigs.k8s.io/leader-address"
 )
 
--- a/api/leaderworkerset/v1/leaderworkerset_types.go
+++ b/api/leaderworkerset/v1/leaderworkerset_types.go
@@ -304,11 +304,17 @@ const (
 	// will share. The host names look like:
 	// Replica 0: my-lws-0.my-lws, my-lws-0-1.my-lws
 	// Replica 1: my-lws-1.my-lws, my-lws-1-1.my-lws
+	// With groupIdentity Hash, the leader host name is an 8 character prefix of
+	// the group key and worker host names extend the leader pod name:
+	// Group a1b2c3d4...: a1b2c3d4.my-lws, my-lws-7d9f8b6c4-x2kkp-1.my-lws
 	SubdomainShared SubdomainPolicy = "Shared"
 	// UniquePerReplica will create a headless service per replica
 	// The pod host names look like:
 	// Replica 0: my-lws-0.my-lws-0,my-lws-0-1.my-lws-0, my-lws-0-2.my-lws-0
 	// Replica 1: my-lws-1.my-lws-1,my-lws-1-1.my-lws-1, my-lws-1-2.my-lws-1
+	// With groupIdentity Hash, the per replica service is named with the group
+	// key prefix instead of the replica ordinal:
+	// Group a1b2c3d4...: a1b2c3d4.my-lws-a1b2c3d4, my-lws-7d9f8b6c4-x2kkp-1.my-lws-a1b2c3d4
 	SubdomainUniquePerReplica SubdomainPolicy = "UniquePerReplica"
 )
 
--- a/api/leaderworkerset/v1/leaderworkerset_types.go
+++ b/api/leaderworkerset/v1/leaderworkerset_types.go
@@ -103,7 +103,24 @@ const (
 	// Enables feature where the group will be restarted after pod failure if and only if
 	// all pods in the group are not pending
 	RecreateGroupAfterStartAnnotationKey string = "leaderworkerset.sigs.k8s.io/experimental-recreate-group-after-start"
+
+	// GroupIdentityAnnotationKey is set on leader and worker pod templates when the
+	// LeaderWorkerSet runs with GroupIdentity=Hash so that admission and controllers
+	// can tell the identity scheme apart without fetching the LWS object.
+	GroupIdentityAnnotationKey string = "leaderworkerset.sigs.k8s.io/group-identity"
+
+	// LeaderAddressAnnotationKey carries the leader address (pod IP) on worker pods
+	// when GroupIdentity=Hash. With hash-named leaders there is no per-pod DNS record,
+	// and the address only needs to be stable for one group generation because the
+	// whole group is recreated together.
+	LeaderAddressAnnotationKey string = "leaderworkerset.sigs.k8s.io/leader-address"
 )
+
+// GroupReadyConditionType is the pod readiness gate condition set on leader pods when
+// GroupIdentity=Hash. The pod controller marks it True once the group's worker
+// statefulset is ready, which makes Deployment rollout pacing count whole groups
+// instead of bare leader pods.
+const GroupReadyConditionType corev1.PodConditionType = "leaderworkerset.sigs.k8s.io/group-ready"
 
 // One group consists of a single leader and M workers, and the total number of pods in a group is M+1.
 // LeaderWorkerSet will create N replicas of leader-worker pod groups (hereinafter referred to as group).
@@ -147,7 +164,31 @@ type LeaderWorkerSetSpec struct {
 	// networkConfig defines the network configuration of the group
 	// +optional
 	NetworkConfig *NetworkConfig `json:"networkConfig,omitempty"`
+
+	// groupIdentity determines how group identities are assigned.
+	// Ordinal (default) manages leaders through a StatefulSet: groups are named
+	// <lws>-0..<lws>-N-1 and scale down always removes the highest ordinal.
+	// Hash manages leaders through a Deployment: group names are hash-suffixed,
+	// scale down prefers unscheduled and not-ready groups over healthy ones, and
+	// rollouts are paced by a group readiness gate on the leader pods.
+	// This field is immutable.
+	// +kubebuilder:default=Ordinal
+	// +kubebuilder:validation:Enum={Ordinal,Hash}
+	// +optional
+	GroupIdentity GroupIdentityType `json:"groupIdentity,omitempty"`
 }
+
+// GroupIdentityType defines how group identities are assigned.
+type GroupIdentityType string
+
+const (
+	// GroupIdentityOrdinal names groups by contiguous StatefulSet ordinals.
+	GroupIdentityOrdinal GroupIdentityType = "Ordinal"
+
+	// GroupIdentityHash names groups by hash-suffixed leader pod names managed
+	// through a Deployment.
+	GroupIdentityHash GroupIdentityType = "Hash"
+)
 
 // Template of the leader/worker pods, the group will include at least one leader pod.
 // Defaults to the worker template if not specified. The idea is to allow users to create a
--- a/api/leaderworkerset/v1/leaderworkerset_types.go
+++ b/api/leaderworkerset/v1/leaderworkerset_types.go
@@ -304,17 +304,18 @@ const (
 	// will share. The host names look like:
 	// Replica 0: my-lws-0.my-lws, my-lws-0-1.my-lws
 	// Replica 1: my-lws-1.my-lws, my-lws-1-1.my-lws
-	// With groupIdentity Hash, the leader host name is an 8 character prefix of
-	// the group key and worker host names extend the leader pod name:
-	// Group a1b2c3d4...: a1b2c3d4.my-lws, my-lws-7d9f8b6c4-x2kkp-1.my-lws
+	// With groupIdentity Hash, the leader host name is the lws name plus an 8
+	// character prefix of the group key and worker host names extend the leader
+	// pod name:
+	// Group a1b2c3d4...: my-lws-a1b2c3d4.my-lws, my-lws-7d9f8b6c4-x2kkp-1.my-lws
 	SubdomainShared SubdomainPolicy = "Shared"
 	// UniquePerReplica will create a headless service per replica
 	// The pod host names look like:
 	// Replica 0: my-lws-0.my-lws-0,my-lws-0-1.my-lws-0, my-lws-0-2.my-lws-0
 	// Replica 1: my-lws-1.my-lws-1,my-lws-1-1.my-lws-1, my-lws-1-2.my-lws-1
-	// With groupIdentity Hash, the per replica service is named with the group
-	// key prefix instead of the replica ordinal:
-	// Group a1b2c3d4...: a1b2c3d4.my-lws-a1b2c3d4, my-lws-7d9f8b6c4-x2kkp-1.my-lws-a1b2c3d4
+	// With groupIdentity Hash, the per replica service name matches the leader
+	// host name:
+	// Group a1b2c3d4...: my-lws-a1b2c3d4.my-lws-a1b2c3d4, my-lws-7d9f8b6c4-x2kkp-1.my-lws-a1b2c3d4
 	SubdomainUniquePerReplica SubdomainPolicy = "UniquePerReplica"
 )
 
--- a/api/leaderworkerset/v1/leaderworkerset_types.go
+++ b/api/leaderworkerset/v1/leaderworkerset_types.go
@@ -109,10 +109,11 @@ const (
 	// can tell the identity scheme apart without fetching the LWS object.
 	GroupIdentityAnnotationKey string = "leaderworkerset.sigs.k8s.io/group-identity"
 
-	// LeaderAddressAnnotationKey carries the leader address (pod IP) on worker pods
-	// when GroupIdentity=Hash. With hash-named leaders there is no per-pod DNS record,
-	// and the address only needs to be stable for one group generation because the
-	// whole group is recreated together.
+	// LeaderAddressAnnotationKey carries the leader address on worker pods when
+	// GroupIdentity=Hash: the leader's DNS name derived from the group key, or the
+	// leader pod IP for groups admitted before hostnames were assigned. The address
+	// only needs to be stable for one group generation because the whole group is
+	// recreated together.
 	LeaderAddressAnnotationKey string = "leaderworkerset.sigs.k8s.io/leader-address"
 )
 
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -103,6 +103,8 @@ func NewLeaderWorkerSetReconciler(client client.Client, scheme *runtime.Scheme, 
 //+kubebuilder:rbac:groups=apps,resources=statefulsets,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=apps,resources=statefulsets/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=apps,resources=statefulsets/finalizers,verbs=update
+//+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=apps,resources=deployments/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=apps,resources=controllerrevisions,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=apps,resources=controllerrevisions/status,verbs=get;update;patch
@@ -121,6 +123,10 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 	log := ctrl.LoggerFrom(ctx).WithValues("leaderworkerset", klog.KObj(lws))
 	ctx = ctrl.LoggerInto(ctx, log)
+
+	if lws.Spec.GroupIdentity == leaderworkerset.GroupIdentityHash {
+		return r.reconcileHash(ctx, lws)
+	}
 
 	leaderSts, err := r.getLeaderStatefulSet(ctx, lws)
 	if err != nil {
@@ -225,6 +231,7 @@ func (r *LeaderWorkerSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&leaderworkerset.LeaderWorkerSet{}).
 		Owns(&appsv1.StatefulSet{}).
+		Owns(&appsv1.Deployment{}).
 		Owns(&corev1.Service{}).
 		Watches(&appsv1.StatefulSet{},
 			handler.EnqueueRequestsFromMapFunc(enqueueLWSRequests)).
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -153,10 +153,13 @@ func (r *LeaderWorkerSetReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, err
 	}
 
-	updatedRevision, err := r.getUpdatedRevision(ctx, leaderSts != nil, lws, revision)
-	if err != nil {
-		log.Error(err, "Validating if LWS has been updated")
-		return ctrl.Result{}, err
+	var updatedRevision *appsv1.ControllerRevision
+	if leaderSts != nil {
+		updatedRevision, err = r.getUpdatedRevision(ctx, lws, revision)
+		if err != nil {
+			log.Error(err, "Validating if LWS has been updated")
+			return ctrl.Result{}, err
+		}
 	}
 	lwsUpdated := updatedRevision != nil
 	if lwsUpdated {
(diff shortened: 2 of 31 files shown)


pkg/webhooks/pod_webhook.go
```diff
@@ -99,6 +100,57 @@ func (p *PodWebhook) Default(ctx context.Context, pod *corev1.Pod) error {
...
 	// adding labels for pods
 	if podutils.LeaderPod(*pod) {
+		if pod.Annotations[leaderworkerset.GroupIdentityAnnotationKey] == string(leaderworkerset.GroupIdentityHash) {
+			// Hash-identity leaders are created through a Deployment: the pod name is
+			// not known at admission (generateName), so the group identity is a fresh
+			// random key rather than a name-derived ordinal.
+			var groupUniqueKey string
+			if _, foundGroupKey := pod.Labels[leaderworkerset.GroupUniqueHashLabelKey]; !foundGroupKey {
+				groupUniqueKey = genGroupUniqueKey(pod.Namespace, utilrand.String(16))
+				pod.Labels[leaderworkerset.GroupUniqueHashLabelKey] = groupUniqueKey
```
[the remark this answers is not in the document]
BenjaminBraunDev: Reading a nil map is safe in Go, and default returns early unless pod.Labels already contains the LWS set-name label, so both maps are guaranteed non-nil at every write site. This is the same pattern the ordinal path has always used.

pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -186,6 +194,37 @@ func (r *LeaderWorkerSetWebhook) generalValidate(lws *v1.LeaderWorkerSet) field.
...
+}
+
+// ValidateGroupIdentity rejects the parts of the API surface whose semantics
+// depend on stable StatefulSet identity and are not supported when leaders are
+// Deployment-managed. It is a no-op unless the spec asks for groupIdentity Hash.
+// Exported so the DisaggregatedSet webhook can run the same checks against each
+// role's inline LeaderWorkerSet spec at DisaggregatedSet admission time, where a
+// bad combination would otherwise only surface as LWS creation failures during
+// reconciliation.
+func ValidateGroupIdentity(specPath *field.Path, spec *v1.LeaderWorkerSetSpec) field.ErrorList {
```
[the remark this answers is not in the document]
BenjaminBraunDev: updated

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -96,8 +96,26 @@ const (
...
+	// LeaderWorkerSet runs with GroupIdentity=Hash so that admission and controllers
+	// can tell the identity scheme apart without fetching the LWS object.
+	GroupIdentityAnnotationKey string = "leaderworkerset.sigs.k8s.io/group-identity"
+
+	// LeaderAddressAnnotationKey carries the leader address on worker pods when
+	// GroupIdentity=Hash: the leader's DNS name derived from the group key, or the
+	// leader pod IP for groups admitted before hostnames were assigned. The address
+	// only needs to be stable for one group generation because the whole group is
+	// recreated together.
+	LeaderAddressAnnotationKey string = "leaderworkerset.sigs.k8s.io/leader-address"
```
ahg-g, line 115: We don't need to limit it to hash, lets unify and inject it for all modes.

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -140,8 +158,32 @@ type LeaderWorkerSetSpec struct {
...
+	// Ordinal (default) manages leaders through a StatefulSet: groups are named
+	// <lws>-0..<lws>-N-1 and scale down always removes the highest ordinal.
+	// Hash manages leaders through a Deployment: group names are hash-suffixed,
+	// scale down prefers unscheduled and not-ready groups over healthy ones, and
+	// rollouts are paced by a group readiness gate on the leader pods.
+	// This field is immutable.
+	// +kubebuilder:default=Ordinal
+	// +kubebuilder:validation:Enum={Ordinal,Hash}
+	// +optional
+	GroupIdentity GroupIdentityType `json:"groupIdentity,omitempty"`
```
ahg-g: in hash mode, only LeaderReady StartupPolicy is applicable, if so, please document that and validate it.

BenjaminBraunDev:
@hasB4K Could you take a look at this when you have a moment?

tl;dr this allows LWS to be a "Deployment of StatefulSets" instead of its current "StatefulSet of StatefulSets", this allows for unordered replicas of leader-worker groups, so we can have smarter scale-down vs. always scaling down the highest ordinal (e.g. we can tear down any unhealthy groups first etc.)

pkg/webhooks/pod_webhook.go
```diff
@@ -99,6 +100,57 @@ func (p *PodWebhook) Default(ctx context.Context, pod *corev1.Pod) error {
 	}
 	// adding labels for pods
 	if podutils.LeaderPod(*pod) {
+		if pod.Annotations[leaderworkerset.GroupIdentityAnnotationKey] == string(leaderworkerset.GroupIdentityHash) {
+			// Hash-identity leaders are created through a Deployment: the pod name is
+			// not known at admission (generateName), so the group identity is a fresh
+			// random key rather than a name-derived ordinal.
+			var groupUniqueKey string
+			if _, foundGroupKey := pod.Labels[leaderworkerset.GroupUniqueHashLabelKey]; !foundGroupKey {
```
ahg-g: why do we need `GroupUniqueHashLabelKey` label if `GroupIndexLabelKey` is also set to the generated unique key?

pkg/utils/pod/pod_utils.go
```diff
@@ -144,6 +144,32 @@ func AddLWSVariables(pod *corev1.Pod) error {
 		Name:  leaderworkerset.LwsLeaderAddress,
 		Value: fmt.Sprintf("%s-%s.%s.%s", lwsName, groupIndex, pod.Spec.Subdomain, pod.ObjectMeta.Namespace),
 	}
+	if pod.Annotations[leaderworkerset.GroupIdentityAnnotationKey] == string(leaderworkerset.GroupIdentityHash) {
```
ahg-g: I would change AddLWSVariables func to pass in the leader address so that we don't need to special case here based on group identity.

pkg/controllers/pod_controller.go
```diff
@@ -158,6 +177,28 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 		return ctrl.Result{}, err
 	}
 
+	if hashIdentity {
+		// Workers reach the leader through the DNS record the webhook derived
+		// from the group key. Leaders admitted before hostnames were assigned
```
ahg-g: In what cases would we have leaders admitted without hostnames assigned? I don't think we have such a case; if no such case exists indeed, then I would simplify things here. In fact we should error out if hostname and subdomain are not set.

pkg/controllers/pod_controller.go
```diff
@@ -431,6 +516,10 @@ func constructWorkerStatefulSetApplyConfiguration(leaderPod corev1.Pod, lws lead
 	serviceName := leaderPod.Name
```
ahg-g: The service name is always equal to leaderPod.Subdomain in all cases and all modes, I think here we can just do `serviceNAme = leaderPod.Spec.Subdomain` and remove lines 517 to 522

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -103,8 +103,26 @@ const (
...
 	RecreateGroupAfterStartAnnotationKey string = "leaderworkerset.sigs.k8s.io/experimental-recreate-group-after-start"
+
+	// GroupIdentityAnnotationKey is set on leader and worker pod templates when the
+	// LeaderWorkerSet runs with GroupIdentity=Hash so that admission and controllers
+	// can tell the identity scheme apart without fetching the LWS object.
+	GroupIdentityAnnotationKey string = "leaderworkerset.sigs.k8s.io/group-identity"
+
+	// LeaderAddressAnnotationKey carries the leader address on worker pods when
+	// GroupIdentity=Hash: the leader's DNS name derived from the group key, or the
+	// leader pod IP for groups admitted before hostnames were assigned. The address
```
ahg-g: we don't need to change semantics and use the IP, just keep it consistent with the existing ordinal mode and set it to he dns address, there should be no case where we can't set this to the dns address. Also, lets be consistent and set this value for both modes.

pkg/webhooks/pod_webhook.go
```diff
@@ -99,6 +100,57 @@ func (p *PodWebhook) Default(ctx context.Context, pod *corev1.Pod) error {
...
+			}
+			// The group key also provides the leader's DNS identity: hostname is a
+			// prefix of the key and subdomain is the headless service the record
+			// lives under. Both fields are immutable, so admission is the only
+			// chance to set them.
+			hostPrefix := hashDNSPrefix(groupUniqueKey)
+			if pod.Spec.Hostname == "" {
+				pod.Spec.Hostname = hostPrefix
+			}
+			if pod.Spec.Subdomain == "" {
```
ahg-g: the leader pod subdomain should not be empty, just like the ordinal case, we set a default, which is the lws name, and then we override here if the subdomain policy is UniquePerReplica, this will help consolidate the code and minimize the diff between ordinal and hash.

pkg/webhooks/pod_webhook.go
```diff
@@ -99,6 +100,57 @@ func (p *PodWebhook) Default(ctx context.Context, pod *corev1.Pod) error {
 	}
 	// adding labels for pods
 	if podutils.LeaderPod(*pod) {
+		if pod.Annotations[leaderworkerset.GroupIdentityAnnotationKey] == string(leaderworkerset.GroupIdentityHash) {
```
ahg-g: this is duplicating a lot of code that also applies to the ordinal mode, if I am not mistaken, the only difference between the two is the values of group unique key and the index, and setting the hostname, the rest should be similar.

pkg/controllers/pod_controller.go
```diff
@@ -122,7 +123,14 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 	}
 
 	if leaderWorkerSet.Spec.NetworkConfig != nil && *leaderWorkerSet.Spec.NetworkConfig.SubdomainPolicy == leaderworkerset.SubdomainUniquePerReplica {
-		if err := controllerutils.CreateHeadlessServiceIfNotExists(ctx, r.Client, r.Scheme, &leaderWorkerSet, pod.Name, map[string]string{leaderworkerset.SetNameLabelKey: leaderWorkerSet.Name, leaderworkerset.GroupIndexLabelKey: pod.Labels[leaderworkerset.GroupIndexLabelKey]}, &pod); err != nil {
+		// In hash mode the per-replica service carries the group key derived name
+		// the webhook stamped as the leader's subdomain, since Service names must
+		// begin with a letter and the pod name is not known at admission.
+		serviceName := pod.Name
```
ahg-g: the service name should always equal to the leader's pod subdomain, so here we should keep the code as is and just change `pod.Name` to `pod.Spec.Subdomain` in the CreateHeadlessServiceIfNotExists call

pkg/controllers/leaderworkerset_hash.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+	podTemplateApplyConfiguration.WithAnnotations(podAnnotations)
+
+	// The gate keeps a leader pod not-ready until its worker statefulset is ready,
+	// so the Deployment's maxUnavailable budget counts whole groups.
+	if *lws.Spec.LeaderWorkerTemplate.Size > 1 {
+		if podTemplateApplyConfiguration.Spec == nil {
+			podTemplateApplyConfiguration.Spec = coreapplyv1.PodSpec()
+		}
+		podTemplateApplyConfiguration.Spec.WithReadinessGates(
```
ahg-g, line 172: we need to set the subdomain in the pod template to `lws.Name`; this will work for the Shared subdomain policy, and will be overriden in the webhook if the subdomain policy is `UniquePerReplica`

ahg-g:
@BenjaminBraunDev pls document the hostnames format for the hash mode just like we did for the ordinal mode; document it here pls: https://github.com/kubernetes-sigs/lws/blob/main/api/leaderworkerset/v1/leaderworkerset_types.go#L262

api/leaderworkerset/v1/leaderworkerset_types.go
BenjaminBraunDev: Done. Made pod controller stamp the annotation on worker templates in both modes

api/leaderworkerset/v1/leaderworkerset_types.go
BenjaminBraunDev: sg, removed the IP fallback + made annotation carry the DNS for both

pkg/utils/pod/pod_utils.go
BenjaminBraunDev: done, made AddLWSVariables take the address

pkg/controllers/pod_controller.go
BenjaminBraunDev: Yeah true, admission+deploy template always set both of them. Removed the fallback and made it error.

pkg/controllers/pod_controller.go
BenjaminBraunDev: done

pkg/webhooks/pod_webhook.go
BenjaminBraunDev: Done, made it always set to LWS name and only override for UniquePerReplica policy, in both modes.

pkg/controllers/pod_controller.go
BenjaminBraunDev: done

pkg/webhooks/pod_webhook.go
BenjaminBraunDev: Refactored, made branch only differ in how the group key, index, and hostname are derived, and the rest is shared with ordinal

pkg/controllers/leaderworkerset_hash.go
BenjaminBraunDev: done

BenjaminBraunDev:
@ahg-g Pushed a follow up that builds the leader pod template and workload metadata through one shared function for both the sts and deployment paths to further consolidate. The larger diff is just due to additional tests, the functional code is about -100 lines from before

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -306,11 +304,17 @@ const (
 	// will share. The host names look like:
 	// Replica 0: my-lws-0.my-lws, my-lws-0-1.my-lws
 	// Replica 1: my-lws-1.my-lws, my-lws-1-1.my-lws
+	// With groupIdentity Hash, the leader host name is an 8 character prefix of
+	// the group key and worker host names extend the leader pod name:
+	// Group a1b2c3d4...: a1b2c3d4.my-lws, my-lws-7d9f8b6c4-x2kkp-1.my-lws
```
ahg-g: should we prefix the leader with the lws name to be consistent with the workers, and the ordinal mode?

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -306,11 +304,17 @@ const (
...
 	SubdomainShared SubdomainPolicy = "Shared"
 	// UniquePerReplica will create a headless service per replica
 	// The pod host names look like:
 	// Replica 0: my-lws-0.my-lws-0,my-lws-0-1.my-lws-0, my-lws-0-2.my-lws-0
 	// Replica 1: my-lws-1.my-lws-1,my-lws-1-1.my-lws-1, my-lws-1-2.my-lws-1
+	// With groupIdentity Hash, the per replica service is named with the group
+	// key prefix instead of the replica ordinal:
+	// Group a1b2c3d4...: a1b2c3d4.my-lws-a1b2c3d4, my-lws-7d9f8b6c4-x2kkp-1.my-lws-a1b2c3d4
```
ahg-g: ditto, recommend to prefix with the lws name to be consistent with the workers and ordinal mode.

pkg/controllers/leaderworkerset_controller.go
```diff
@@ -752,13 +763,13 @@ func (r *LeaderWorkerSetReconciler) getOrCreateRevisionIfNonExist(ctx context.Co
...
-		recorder.Eventf(lws, newRevision, corev1.EventTypeNormal, CreatingRevision, Create, message)
+		r.Record.Eventf(lws, newRevision, corev1.EventTypeNormal, CreatingRevision, Create, message)
 	}
 	return newRevision, err
 }
 
-func (r *LeaderWorkerSetReconciler) getUpdatedRevision(ctx context.Context, sts *appsv1.StatefulSet, lws *leaderworkerset.LeaderWorkerSet, revision *appsv1.ControllerRevision) (*appsv1.ControllerRevision, error) {
-	if sts == nil {
+func (r *LeaderWorkerSetReconciler) getUpdatedRevision(ctx context.Context, leaderExists bool, lws *leaderworkerset.LeaderWorkerSet, revision *appsv1.ControllerRevision) (*appsv1.ControllerRevision, error) {
+	if !leaderExists {
```
ahg-g: This is a bit strange, lets just depend on the caller to not call the function when the leader doesn't exist instead of passing in a flag.

pkg/webhooks/pod_webhook.go
```diff
@@ -99,35 +100,68 @@ func (p *PodWebhook) Default(ctx context.Context, pod *corev1.Pod) error {
...
-			_, groupIndex := statefulsetutils.GetParentNameAndOrdinal(pod.Name)
-			if groupIndex == -1 {
-				return fmt.Errorf("parsing pod ordinal for pod %s", pod.Name)
+		hashIdentity := pod.Annotations[leaderworkerset.GroupIdentityAnnotationKey] == string(leaderworkerset.GroupIdentityHash)
+		var groupUniqueKey string
+		if hashIdentity {
+			// Hash-identity leaders are created through a Deployment: the pod name is
+			// not known at admission (generateName), so the group identity is a fresh
+			// random key rather than a name-derived ordinal.
+			if _, foundGroupKey := pod.Labels[leaderworkerset.GroupUniqueHashLabelKey]; !foundGroupKey {
```
ahg-g: why do we do all these if/else statements? just set and override the labels for these keys, things would break for the most part if the labels are not set to what we expect them to be, besides these labels are owned by this controller and are not expected to be set by anyone but us.

pkg/webhooks/pod_webhook.go
```diff
@@ -99,35 +100,68 @@ func (p *PodWebhook) Default(ctx context.Context, pod *corev1.Pod) error {
...
+			}
+			// The group key also provides the leader's hostname, a prefix of the key.
+			// The field is immutable, so admission is the only chance to set it. The
+			// subdomain default comes from the pod template.
+			if pod.Spec.Hostname == "" {
+				pod.Spec.Hostname = hashDNSPrefix(groupUniqueKey)
+			}
+		} else {
+			// add group index label to group pods
+			if _, found := pod.Labels[leaderworkerset.GroupIndexLabelKey]; !found {
```
ahg-g: ditto, pls simplify and just set the labels without all this if/else checking.

pkg/webhooks/pod_webhook.go
```diff
@@ -99,35 +100,68 @@ func (p *PodWebhook) Default(ctx context.Context, pod *corev1.Pod) error {
...
-			pod.Labels[leaderworkerset.GroupUniqueHashLabelKey] = groupUniqueKey
-		} else {
-			groupUniqueKey = pod.Labels[leaderworkerset.GroupUniqueHashLabelKey]
+			if hashIdentity {
+				// Service names must begin with a letter, so the per-replica
+				// service cannot be named the raw key.
+				pod.Spec.Subdomain = fmt.Sprintf("%s-%s", pod.Labels[leaderworkerset.SetNameLabelKey], pod.Spec.Hostname)
+			} else {
+				pod.Spec.Subdomain = pod.Name
+			}
```
ahg-g, line 133:
```suggestion
            pod.Spec.Subdomain = pod.Name
			if hashIdentity {
				// Service names must begin with a letter, so the per-replica
				// service cannot be named the raw key.
				pod.Spec.Subdomain = fmt.Sprintf("%s-%s", pod.Labels[leaderworkerset.SetNameLabelKey], pod.Spec.Hostname)
			}
```

ahg-g (commented):
a few more comments to simplify the code a bit more.

api/leaderworkerset/v1/leaderworkerset_types.go
BenjaminBraunDev: Yeah that's probably a good idea. Made leader lws name plus the key prefix.

api/leaderworkerset/v1/leaderworkerset_types.go
BenjaminBraunDev: This also makes the service name and the leader host name the same string so the webhook can set one from the other

pkg/controllers/leaderworkerset_controller.go
BenjaminBraunDev: Dropped the flag

pkg/webhooks/pod_webhook.go
BenjaminBraunDev: Removed the checks

pkg/webhooks/pod_webhook.go
BenjaminBraunDev: done

pkg/webhooks/pod_webhook.go
BenjaminBraunDev: With the prefixed host name change from the other comment, the per replica service name is already the leader host name, so I'm having the hash branch just sets the subdomain from the hostname.

api/leaderworkerset/v1/leaderworkerset_types.go
```diff
@@ -304,17 +304,18 @@ const (
...
 	// UniquePerReplica will create a headless service per replica
 	// The pod host names look like:
 	// Replica 0: my-lws-0.my-lws-0,my-lws-0-1.my-lws-0, my-lws-0-2.my-lws-0
 	// Replica 1: my-lws-1.my-lws-1,my-lws-1-1.my-lws-1, my-lws-1-2.my-lws-1
-	// With groupIdentity Hash, the per replica service is named with the group
-	// key prefix instead of the replica ordinal:
-	// Group a1b2c3d4...: a1b2c3d4.my-lws-a1b2c3d4, my-lws-7d9f8b6c4-x2kkp-1.my-lws-a1b2c3d4
+	// With groupIdentity Hash, the per replica service name matches the leader
+	// host name:
+	// Group a1b2c3d4...: my-lws-a1b2c3d4.my-lws-a1b2c3d4, my-lws-7d9f8b6c4-x2kkp-1.my-lws-a1b2c3d4
```
ahg-g, line 318: as a followup, pls update this to include the hash added by Deployment for the replicaset

ahg-g: actually, the dns name is I think correct as described here, what we need to document is the pod names.

ahg-g (commented):
/approve
/lgtm

PR #989, opened 2026-08-18
[docs] restructure documentation for LeaderWorkerSet and DisaggregatedSet concepts and overview

ahg-g:
## Summary

  This PR restructures and expands the documentation on the site to provide a clear, modular, and comprehensive guide for both LeaderWorkerSet (LWS) and DisaggregatedSet (DS) as complementary APIs for distributed AI/ML workloads.

## Key Changes

  1. Overview & Landing Page Updates
      • Updated site/content/en/docs/overview/_index.md to introduce both LeaderWorkerSet (core pod group replication) and DisaggregatedSet (multi-role disaggregated inference orchestration).
      • Aligned feature headings (Core Features: LeaderWorkerSet and Advanced Features: DisaggregatedSet) across overview/_index.md and README.md.
      • Embedded concept architecture diagrams (/images/lws-concept.svg and /images/ds-concept.svg).
  2. Top-Level Concepts Reorganization (site/content/en/docs/concepts/)
      • Restructured the top-level Concepts page into a unified architecture overview illustrating how DisaggregatedSet orchestrates underlying LeaderWorkerSets.
      • Added an API feature comparison matrix and a "When to Use Which API" decision guide.
      • Cleaned up redundant inline concept sections to rely on Docsy sidebar navigation.
  3. LeaderWorkerSet Concepts Modularization (site/content/en/docs/concepts/leaderworkerset/)
      • Created a dedicated leaderworkerset/ landing page explaining the two-tier StatefulSet composition architecture (Leader StatefulSet + Worker StatefulSets) and design rationale.
      • Split monolithic concept documentation into dedicated subpages:
          • Dual Pod Templates (pod-templates/): Configuring distinct specifications for leader and worker pods.
          • Startup Policy (startup-policy/): Worker creation timing relative to leader readiness (LeaderReady, LeaderCreated).
          • Exclusive Topology Placement (topology-placement/): Co-locating replica pods on dedicated topology domains (racks, hosts).
          • Subgroups (subgroups/): Subgroup sizing, independent placement, and LeaderOnly heterogeneous scheduling.
          • Volume Claim Templates (volume-claim-templates/): Dynamic per-pod storage provisioning for leaders and workers.
          • Rollout Strategy (rollout-strategy/): Rolling update mechanics, maxUnavailable, and maxSurge.
          • Failure Handling (failure-handling/): Group restart policies (RecreateGroupOnPodRestart, None, RecreateGroupAfterStart) and node failure recovery.

  4. DisaggregatedSet Concepts Modularization (site/content/en/docs/concepts/disaggregatedset/)
      • Created a dedicated disaggregatedset/ landing page detailing multi-role orchestration, child LWS naming, and core design principles.
      • Added dedicated concept subpages:
          • Roles (roles/): Defining serving phases (prefill, decode, encode), independent pod templates, and child LWS mapping.
          • Slices (slices/): Replicating entire multi-role topologies into N independent failure and rollout domains (KEP-846 https://github.com/kubernetes-sigs/lws/tree/main/keps/846-disaggregatedset-slices).
          • Placement Policy (placement-policy/): Intra-slice co-location, inter-slice spread, and domain exclusivity (ExclusiveSlice, ExclusiveTopology) (KEP-848 https://github.com/kubernetes-sigs/lws/tree/main/keps/848-disaggregatedset-placement-policy).
          • Independent Role Autoscaling (role-scaler/): Stable per-role /scale target via DisaggregatedSetRoleScaler for HPA/KEDA integration across rolling updates (KEP-849 https://github.com/kubernetes-sigs/lws/tree/main/keps/849-DisaggregatedSet-HPA).

  5. Site Navigation & Polish
      • Renamed "Management of LWS" section to Operations (site/content/en/docs/manage/_index.md).
      • Moved Contribution Guidelines to the bottom of the left sidebar navigation (weight: 100).

/kind documentation


site/content/en/docs/concepts/disaggregatedset/roles/_index.md
```diff
@@ -0,0 +1,95 @@
...
+```yaml
+apiVersion: disaggregatedset.x-k8s.io/v1
+kind: DisaggregatedSet
+metadata:
+  name: disaggregatedset-sample
+spec:
+  roles:
+  - name: prefill
+    replicas: 2
+    leaderWorkerTemplate:
```
yankay: Could we add the missing `spec` block under each role and move `replicas` and `leaderWorkerTemplate` beneath it? These fields belong to `roles[].spec`, so the current examples do not match the DisaggregatedSet schema.

The same issue appears in `roles/_index.md:57`, `slices/_index.md:41`, `placement-policy/_index.md:73`, and `role-scaler/_index.md:69`:

```yaml
spec:
  roles:
  - name: prefill
    spec:
      replicas: 2
      leaderWorkerTemplate:
        size: 4
        # ...
```

site/content/en/docs/concepts/disaggregatedset/role-scaler/_index.md
```diff
@@ -0,0 +1,149 @@
...
+  roles:
+  - name: prefill
+    scaling:
+      mode: External
+    leaderWorkerTemplate:
+      size: 4
+      workerTemplate:
+        spec:
+          containers:
+          - name: vllm-prefill
```
yankay: Although CPU-based scaling may be uncommon for GPU inference, could we add a CPU request to the selected Pod template since this example uses averageUtilization? Without a CPU request, HPA cannot calculate this metric and will not scale as intended.

site/content/en/docs/concepts/disaggregatedset/role-scaler/_index.md
```diff
@@ -0,0 +1,149 @@
...
+                               ▼
+┌─────────────────────────────────────────────────────────────┐
+│  Child LeaderWorkerSet: "<ds-name>-0-<revision>-<role-name>"│
+└─────────────────────────────────────────────────────────────┘
+```
+
+1. **Automatic Lifecycle:** When you set `scaling.mode: External` on a role, the DisaggregatedSet controller automatically creates and maintains a `DisaggregatedSetRoleScaler` named `<disaggregatedset-name>-<role-name>`. You do not need to author this resource manually.
+2. **Stable Target:** The scaler's name remains constant regardless of how many rolling updates or revision changes occur.
+3. **Reconciled Replicas:** The external autoscaler (HPA or KEDA) writes the desired replica count to the scaler's `/scale` endpoint. The DisaggregatedSet controller reads this value and scales the corresponding child LeaderWorkerSet accordingly.
+
```
yankay, line 50: Could we mention here that External scaling currently requires `spec.slices: 1`? Users may naturally combine these features, but admission rejects `slices > 1` when any role uses scaling.mode: External.

site/content/en/docs/concepts/disaggregatedset/placement-policy/_index.md
```diff
@@ -0,0 +1,113 @@
...
+### 2. `ExclusiveSlice`
+- **Co-location:** All roles (e.g. prefill and decode) belonging to the same slice are co-located in the same topology domain.
+- **Spread:** Different slices of the same DisaggregatedSet are scheduled onto separate topology domains.
+- **Sharing:** Slices from *other* DisaggregatedSets or other cluster workloads are permitted to share the domain if capacity allows.
+
+**Use Case:** Cost-efficient production serving where you want low-latency KV-cache handoff within each slice and fault isolation across your slices, while allowing dense cluster bin-packing.
+
+### 3. `ExclusiveTopology`
+- **Co-location & Spread:** Performs everything `ExclusiveSlice` does (co-locates roles within a slice and spreads slices across domains).
+- **Exclusivity (1:1 Domain Mapping):** Guarantees that the topology domain is exclusively reserved for that single slice. Pods from other DisaggregatedSets are strictly prevented from landing in the same domain.
```
yankay: This wording may imply that the entire topology domain is reserved for the slice. Could we clarify that the exclusivity applies only among DisaggregatedSet slices? Unrelated workloads are not matched by the injected anti-affinity and may still share the domain.

site/content/en/docs/concepts/leaderworkerset/failure-handling/_index.md
```diff
@@ -0,0 +1,92 @@
...
+        containers:
+        - name: worker
+          image: worker-image:latest
+```
+
+### RecreateGroupAfterStart
+
+When any pod in a group fails, the entire group is recreated **if and only if there are no pods currently pending** in the group.
+
+- **Pod Failures:** Recreates the entire group if a pod fails after initial startup is complete. If pods are still pending (e.g., during large container image pulls or initial scheduling), group recreation is deferred until the group stabilizes, preventing premature restart cascades.
```
yankay: Could we avoid describing this as deferred recreation? Since #725, when any Pod is Pending, the controller skips the current failure event without saving it or scheduling a retry; #757 only graduated this behavior to the formal API. It therefore does not automatically retry after the group stabilizes.

site/content/en/docs/overview/_index.md
```diff
@@ -1,58 +1,62 @@
...
-disaggregated workload deployment, with support for coordinated rollouts and failure handling.
-
-Key features of DisaggregatedSet include:
+### Advanced Features: DisaggregatedSet
 - **Disaggregated Architecture Support:** Specifically designed for workloads where different phases (e.g., prefill and decode) run on separate infrastructure.
-- **Coordinated N-Dimensional Rollouts:** Updates multiple roles (2–10) in lockstep, preserving capacity ratios throughout the update process.
+- **Coordinated N-Dimensional Rollouts:** Updates multiple roles (2-10) in lockstep, preserving capacity ratios throughout the update process.
 - **Unified Lifecycle Management:** Manages multiple underlying LeaderWorkerSets as a single logical unit.
 - **Automatic Service Orchestration:** Automatically creates and manages headless services for each role to facilitate discovery and revision-aware routing.
 - **Advanced Failure Handling:** Coordinated drain and restart policies across all roles in the disaggregated set.
```
yankay: Could we clarify what “coordinated restarts across roles” means here? The current wording suggests that DisaggregatedSet coordinates restarts across roles, but this behavior is not currently implemented; cross-role coordination is limited to rollout and drain. Perhaps “coordinated lifecycle operations, such as rollout and drain” would be more precise.

yankay (commented):
Thanks for putting this together. I left a few suggestions based on my understanding; I may have missed some context, but I hope they’re helpful.

site/content/en/docs/concepts/disaggregatedset/roles/_index.md
```diff
@@ -0,0 +1,95 @@
...
+
+```yaml
+apiVersion: disaggregatedset.x-k8s.io/v1
+kind: DisaggregatedSet
+metadata:
+  name: disaggregatedset-sample
+spec:
+  roles:
+  - name: prefill
+    replicas: 2
```
BenjaminBraunDev: I tested this example, and the corresponding DisaggregatedSet examples on the Slices, Independent Role Autoscaling, and Placement Policy pages, by applying each verbatim to a 1.36 kube-apiserver with the v1 CRDs from main (server-side dry-run). All four are rejected:

```
strict decoding error: unknown field "spec.roles[0].leaderWorkerTemplate", unknown field "spec.roles[0].replicas", unknown field "spec.roles[1].leaderWorkerTemplate", unknown field "spec.roles[1].replicas"
```

The cause is that `DisaggregatedRoleSpec` embeds `LeaderWorkerSetTemplateSpec`, so `replicas`, `leaderWorkerTemplate`, and `rolloutStrategy` must be nested under a `spec` key on each role. Only `name`, `scaling`, and `metadata` sit at the role level. With field validation set to ignore, the unknown fields get pruned and the CEL rule then fails with `no such key: spec`, so these manifests fail in every validation mode.

The correct shape, matching `config/samples/disaggregatedset_v1_disaggregatedset.yaml`:

```yaml
  roles:
  - name: prefill
    spec:
      replicas: 2
      leaderWorkerTemplate:
        size: 4
        ...
```

Two notes on the other pages: the `scaling` block in the autoscaling page example is already correctly placed at the role level, and I confirmed the corrected versions of these examples apply cleanly.

site/content/en/docs/concepts/disaggregatedset/roles/_index.md
```diff
@@ -0,0 +1,95 @@
...
+
+A `DisaggregatedSet` spec defines a `roles` list where each entry contains:
+
+| Field | Type | Description |
+| :--- | :--- | :--- |
+| `name` | `string` | Unique name for this role within the set (e.g., `prefill`, `decode`, `encode`). |
+| `replicas` | `int32` | Number of LWS replicas (pod groups) for this role per slice. |
+| `leaderWorkerTemplate` | `LeaderWorkerTemplate` | Full pod template defining the leader and worker pod containers, resource requests/limits, restart policies, and subgroup configurations for this role. |
+| `rolloutStrategy` | `RolloutStrategy` | Optional rolling update configuration for this role. DisaggregatedSet coordinates updates across roles to maintain capacity ratios. |
+| `scaling` | `RoleScalingMode` | Optional external scaling mode (e.g., enabling HPA via `DisaggregatedSetRoleScaler`). |
```
BenjaminBraunDev: Same nesting issue in this table: `replicas`, `leaderWorkerTemplate`, and `rolloutStrategy` live under `roles[].spec`, not directly on the role entry. Also, `scaling` is a `RoleScaling` struct containing a `mode` field. `RoleScalingMode` is the type of `scaling.mode`, not of `scaling`.

site/content/en/docs/concepts/disaggregatedset/role-scaler/_index.md
```diff
@@ -0,0 +1,149 @@
...
+```yaml
+apiVersion: disaggregatedset.x-k8s.io/v1
+kind: DisaggregatedSet
+metadata:
+  name: llm-serving
+spec:
+  roles:
+  - name: prefill
+    scaling:
+      mode: External
```
BenjaminBraunDev: Worth documenting the alpha restriction here: the webhook rejects `spec.slices > 1` when any role has `scaling.mode: External` (see `validateExternalScaling` in `pkg/webhooks/disaggregatedset/disaggregatedset_webhook.go`), and it warns when an External role also sets `spec.replicas > 1`. Since the Slices page presents slices and per-role autoscaling as freely composable, a note here and/or on the Slices page would save users a confusing rejection.

site/content/en/docs/concepts/disaggregatedset/slices/_index.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Naming and Identification
+
+Child resources belonging to a slice include the slice index in their name and labels:
+
+### Resource Naming Format
+`<DisaggregatedSet-name>-<slice-index>-<revision-hash>-<role-name>`
+
+Example for `slices: 2`:
+- `my-inference-0-7f9b8c-prefill`
```
BenjaminBraunDev: Nit: the revision hash is 8 hex characters (`revisionLength = 8` in `pkg/utils/disaggregatedset/utils.go`), this example shows 6.

site/content/en/docs/concepts/disaggregatedset/placement-policy/_index.md
BenjaminBraunDev: This overstates the guarantee. The anti-affinity term injected for `ExclusiveTopology` only matches pods carrying the `disaggregatedset.x-k8s.io/name` label, so it excludes other DisaggregatedSets' slices but not unrelated workloads. The code comment in `pkg/utils/disaggregatedset/affinity.go` states explicitly that unrelated workloads may still share the domain. The comparison matrix below is accurate, but "exclusively reserved for that single slice" here and "guaranteed hardware isolation and zero noisy-neighbor interference" in the use case line should be scoped to other DisaggregatedSets.

site/content/en/docs/concepts/disaggregatedset/placement-policy/_index.md
```diff
@@ -0,0 +1,113 @@
...
+                nvidia.com/gpu: "4"
+```
+
+In this example:
+- **Slice 0** (prefill + decode pods) lands together on **Rack A**.
+- **Slice 1** (prefill + decode pods) lands together on **Rack B**.
+
+---
+
+## How It Works Under the Hood
```
BenjaminBraunDev: One caveat worth adding in this section: affinity is injected when a child LeaderWorkerSet is created, so changing `placementPolicy` on a live DisaggregatedSet takes effect on the next rollout, not immediately. The API comment on `PlacementPolicy` in `disaggregatedset_types.go` calls this out.

site/content/en/docs/concepts/_index.md
```diff
@@ -3,197 +3,64 @@ title: "Concepts"
...
+| Feature / Dimension | LeaderWorkerSet (LWS) | DisaggregatedSet (DS) |
+| :--- | :--- | :--- |
+| **Primary Purpose** | Deploying a group of pods as a unit of replication | Orchestrating multi-role disaggregated serving topologies |
+| **Unit of Replication** | Replica = 1 leader Pod + *N* worker Pods | Set = Multiple roles, each mapped to a child LWS |
+| **Workload Type** | Homogeneous multi-node inference | Heterogeneous multi-role inference (prefill, decode, encode) |
+| **CRD** | `leaderworkerset.x-k8s.io/v1` | `disaggregatedset.x-k8s.io/v1` |
+| **Rollout Ownership** | LWS controller (`maxUnavailable`, `maxSurge`) | DisaggregatedSet controller (lockstep, ratio-preserving) |
+| **Scaling** | Horizontal Pod Autoscaler (HPA) via scale subresource | Independent per-role scaling & full topology slice scaling |
+| **Service Discovery** | Headless service per replica (`UniquePerReplica` subdomain policy) | Headless service per role with revision-aware routing |
+| **Placement & Topology** | Exclusive topology placement & subgroups per replica | Role-level placement policies & slice topology spread |
```
BenjaminBraunDev: "Role-level placement policies" is not quite accurate. `placementPolicy` is set-level (`spec.placementPolicy`) and its unit of co-location and spread is the slice, not the role. Suggest "Slice-level placement policy and topology spread".

site/content/en/docs/concepts/disaggregatedset/roles/_index.md
BenjaminBraunDev: This table has the same nesting issue as the example below: `replicas`, `leaderWorkerTemplate`, and `rolloutStrategy` live under `roles[].spec`. Also, the type of `scaling` is listed as `RoleScalingMode`, but that is the type of its nested `mode` field. The `scaling` field itself is a `RoleScaling` struct, so in the YAML this should be `scaling.mode: External` instead of `scaling: External`

site/content/en/docs/concepts/disaggregatedset/slices/_index.md
BenjaminBraunDev: nit: revision hashes are 8 hex characters not 6

site/content/en/docs/concepts/disaggregatedset/placement-policy/_index.md
BenjaminBraunDev, line 110: Probably worth noting that affinity is injected at LWS creation, so changing `placementPolicy` on a live set takes effect on the next rollout, not necessarily immediately.

site/content/en/docs/concepts/_index.md
BenjaminBraunDev: `placementPolicy` is set-level and co-locates / spreads slices, not roles. Suggest "slice-level placement policy and topology spread"

BenjaminBraunDev (dismissed):

BenjaminBraunDev:
Confirming @yankay's comment on the role examples: I applied the page examples to a kube-apiserver with the v1 CRDs from main, and they rejected with unknown fields `spec.roles[*].replicas` and `spec.roles[*].leaderWorkerTemplate`. The corrected `roles[].spec` nesting works. External scaling slices restriction and ExclusiveTopology scope also match the code, +1 to both

site/content/en/docs/concepts/disaggregatedset/placement-policy/_index.md
ahg-g: Updated, but we should actually make it immutable.

site/content/en/docs/concepts/disaggregatedset/role-scaler/_index.md
ahg-g: done.

site/content/en/docs/concepts/disaggregatedset/roles/_index.md
ahg-g: Good catch, done.

site/content/en/docs/concepts/disaggregatedset/role-scaler/_index.md
ahg-g: done.

site/content/en/docs/concepts/disaggregatedset/placement-policy/_index.md
ahg-g: actually exclusive topology is matched against other workloads (at least other disaggsets) and it is truly 1:1

site/content/en/docs/concepts/leaderworkerset/failure-handling/_index.md
ahg-g: Great catch, done.

site/content/en/docs/overview/_index.md
ahg-g: done.

site/content/en/docs/concepts/disaggregatedset/slices/_index.md
ahg-g: done.

site/content/en/docs/concepts/disaggregatedset/roles/_index.md
ahg-g: done.

ahg-g (commented):
Great feedback, thanks!

site/content/en/docs/concepts/_index.md
ahg-g: done.

BenjaminBraunDev:
looked over the latest changes and looks good

BenjaminBraunDev:
@ahg-g I believe the linkTitle/weight lines added here are in auto-generated files that the genref template doesn't emit, so pull-lws-verify-main now fails on the regen diff for PRs based off current main (I noticed this on on #986) Fix: #1005

PR #991, opened 2026-08-20
Fix DisaggregatedSet StatefulSet name validation budget (#972)

### LINKED ISSUE #972: DisaggregatedSet validation misses the StatefulSet name budget
**What happened**:

The validation in https://github.com/kubernetes-sigs/lws/pull/966 only ensures that the generated `<lws>-prv` Service name fits the 63-character limit. It does not reserve space for names generated later by StatefulSet.

For a 40-character DisaggregatedSet name and the role `prefill`, admission accepts a 59-character generated LWS name because its derived Service name is exactly 63 characters. LWS then creates a StatefulSet with the same 59-character name. The StatefulSet controller uses `<statefulset-name>-<revision-hash>` as the Pod's `controller-revision-hash` label, which exceeds the 63-character label-value limit, so Pods cannot be created.

DisaggregatedSet admission should enforce the tightest downstream name budget, including leader and worker StatefulSet names, group indexes, and ControllerRevision hashes.

1. Install an LWS build containing https://github.com/kubernetes-sigs/lws/pull/966.
2. Apply this DisaggregatedSet:

```yaml
apiVersion: disaggregatedset.x-k8s.io/v1
kind: DisaggregatedSet
metadata:
  name: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  namespace: default
spec:
  roles:
    - name: prefill
      spec:
        replicas: 1
        leaderWorkerTemplate:
          size: 1
          workerTemplate:
            spec:
              containers:
                - name: main
                  image: registry.k8s.io/pause:3.10
    - name: decode
      spec:
        replicas: 1
        leaderWorkerTemplate:
          size: 1
          workerTemplate:
            spec:
              containers:
                - name: main
                  image: registry.k8s.io/pause:3.10
```

3. Admission accepts the object. Find the generated `prefill` LWS and inspect its StatefulSet:

```bash
kubectl get leaderworkersets -n default
kubectl describe statefulset -n default <generated-prefill-lws-name>
```

The generated LWS/StatefulSet name is 59 characters. The StatefulSet reports `FailedCreate` because the generated `controller-revision-hash` label value is longer than 63 characters.

Related issues and precedents:

- https://github.com/kubernetes-sigs/lws/issues/959
- https://github.com/kubernetes-sigs/lws/issues/410
- https://github.com/kubernetes/kubernetes/issues/64023
- https://github.com/kubernetes-sigs/jobset/issues/408

**Environment**:

- Kubernetes version (use `kubectl version`): Not cluster-specific
- LWS version (use `git describe --tags --dirty --always`): PR 966 head `e4b1c7fdf1bdc286f28919c69635923ac1589688`
- Cloud provider or hardware configuration: Not applicable
- OS: Not applicable
- Kernel: Not applicable
- Install tools: Not applicable
- Others: None

Dasmat13:
#### What type of PR is this?
/kind bug

This updates `validateGeneratedNames` in the DisaggregatedSet webhook to enforce a tighter downstream name budget. 

Previously, it only accounted for the `-prv` Service suffix. However, it needs to account for the StatefulSet controller revision hash labels, specifically for worker StatefulSets which use the format `<lws>-<groupIndex>-<hash>`. This ensures that the generated labels never exceed the Kubernetes 63-character limit for label values.

Fixes #972

This introduces dynamic calculation of the required overhead per-role depending on the number of replicas (which determines the length of the group index digit), ensuring the tightest possible budget is used to validate the name length at admission time.

```release-note
Fixes a bug where DisaggregatedSet admission accepted names that were too long, causing the downstream StatefulSet to fail to create Pods due to controller-revision-hash label limits.

### DIFF
--- a/pkg/webhooks/disaggregatedset/disaggregatedset_webhook.go
+++ b/pkg/webhooks/disaggregatedset/disaggregatedset_webhook.go
@@ -126,26 +126,43 @@ func (w *DisaggregatedSetWebhook) validateGeneratedNames(obj *disaggv1.Disaggreg
 	sliceDigits := len(strconv.Itoa(int(maxSliceIndex)))
 
 	const (
-		dns1035MaxLen    = 63
-		revisionLen      = 8 // hex characters in the revision hash
-		serviceSuffixLen = 4 // len("-prv")
-		separators       = 3 // three "-" between dsName, slice, revision, roleName
+		dns1035MaxLen                     = 63
+		revisionLen                       = 8 // hex characters in the revision hash
+		serviceSuffixLen                  = 4 // len("-prv")
+		separators                        = 3 // three "-" between dsName, slice, revision, roleName
+		statefulSetRevisionLabelSuffixLen = 11 // len("-<10-char-hash>")
 	)
 
 	rolesPath := field.NewPath("spec", "roles")
 	for i, role := range obj.Spec.Roles {
 		lwsNameLen := len(obj.Name) + separators + sliceDigits + revisionLen + len(role.Name)
-		svcNameLen := lwsNameLen + serviceSuffixLen
+		
+		groupIndexDigits := 1
+		if role.Spec.Replicas != nil && *role.Spec.Replicas > 0 {
+			groupIndexDigits = len(strconv.Itoa(int(*role.Spec.Replicas - 1)))
+		}
 
-		if svcNameLen > dns1035MaxLen {
-			combinedLimit := dns1035MaxLen - separators - sliceDigits - revisionLen - serviceSuffixLen
+		// The worker StatefulSet pods get a label "controller-revision-hash" with value:
+		// <lwsName>-<groupIndex>-<hash>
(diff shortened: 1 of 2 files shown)


BenAyedMedAla (commented):
Verified the arithmetic here — the `controller-revision-hash` format and the 10-char max hash length are both right, and the tests match. Nice fix for the Static-replica case.

One gap though: for roles with `scaling.mode: External`, the actual replica/group count isn't driven by `role.Spec.Replicas` on the DisaggregatedSet at all — it's driven by the separate `DisaggregatedSetRoleScaler` object via its own `/scale` subresource (`disaggregatedsetrolescalers.yaml`: `subresources.scale`). `vdisaggregatedset.kb.io` only watches `disaggregatedsets`, and there's no validating webhook on `disaggregatedsetrolescalers` at all. So an External-scaled role can grow past whatever `groupIndexDigits` was computed at DisaggregatedSet-create time with zero re-validation, reproducing #972's failure mode just via the scaler path instead. Might be worth a follow-up issue if it's out of scope for this PR — happy to file one if useful.

PR #993, opened 2026-08-20
fix: keep kind_load_image archive cleanup in a subshell

### LINKED ISSUE #992: Upgrade E2E fails with "archive: unbound variable"
**What happened**:

The `pull-lws-test-e2e-upgrade-main` job exits before running the upgrade tests:

```text
hack/e2e-test.sh: line 136: archive: unbound variable
```

Example failure:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_lws/953/pull-lws-test-e2e-upgrade-main/2090340013316247552

The temporary image archive should be cleaned up and the upgrade tests should continue.

Trigger `pull-lws-test-e2e-upgrade-main` against a revision containing the current `kind_load_image` implementation.

This was introduced by #928 (`78ea4be`): a `RETURN` trap outlives the function-local `archive` variable, and `nounset` terminates the script when the trap runs again.

A fix exists in the unrelated open PR #877 (`ad9d95b`). It would be useful to extract it into a focused PR.

**Environment**:

- LWS revision: `main` at `3824359`
- Test environment: Kubernetes Prow presubmit
- Others: Bash with `nounset` enabled

AshSgDe29071999:
#### What type of PR is this?

/kind bug

`pull-lws-test-e2e-upgrade-main` exits before the upgrade tests:

```text
hack/e2e-test.sh: line 136: archive: unbound variable
```

`kind_load_image` registered a `RETURN` trap against a function-local
`archive`. The trap outlives that local variable and aborts a later function
return when `nounset` is enabled.

This change runs the image export and import in a subshell and uses an `EXIT`
trap, keeping the cleanup handler and temporary archive in the same scope.

Fixes #992

The original commit was replaced by a maintainer-authored, signed-off commit
because its EasyCLA check remained unresolved. Thanks to @AshSgDe29071999 for
opening the focused fix and documenting the failure.

This PR was written in part with the assistance of generative AI.

```release-note
NONE
```

### DIFF
--- a/hack/e2e-test.sh
+++ b/hack/e2e-test.sh
@@ -119,17 +119,18 @@ function deploy_gang_scheduler() {
 
 # Avoid kind's --all-platforms import path for multi-arch release images.
 function kind_load_image {
-    local image="$1"
-    local archive
-    archive="$(mktemp)"
-    docker save "$image" -o "$archive"
+    (
+        local image="$1"
+        local archive
+        archive="$(mktemp)"
+        trap 'rm -f "$archive"' EXIT
+        docker save "$image" -o "$archive"
 
-    while IFS= read -r node; do
-        docker exec -i "$node" ctr --namespace=k8s.io images import \
-            --digests --snapshotter=overlayfs - < "$archive"
-    done < <($KIND get nodes --name "$KIND_CLUSTER_NAME")
-
-    rm -f "$archive"
+        while IFS= read -r node; do
+            docker exec -i "$node" ctr --namespace=k8s.io images import \
+                --digests --snapshotter=overlayfs - < "$archive"
+        done < <($KIND get nodes --name "$KIND_CLUSTER_NAME")
+    )
 }
 
 function kind_load {


BenAyedMedAla (commented):
Traced through this — it fixes the actual root cause, not just the symptom. A `RETURN` trap set inside a function isn't scoped to that one call; it persists as the shell's active handler and fires on the *next* function return anywhere in the script, which is exactly how `$archive` (local to an already-finished call) went unbound. Moving the body into a subshell with an `EXIT` trap is the right fix: the trap lives and dies with the subshell process and can't leak elsewhere. It also happens to improve cleanup-on-failure — if `docker save` fails under `errexit`, the subshell's `EXIT` trap still runs and removes the temp archive.

Confirmed against CI: `pull-lws-test-e2e-upgrade-main` — the exact job #992 reported failing — now passes, along with everything else. Only red check is EasyCLA, which looks like the one remaining blocker before this is mergeable.

yankay:
Thanks for the fix. Could you please resolve the EasyCLA check by following the [[Kubernetes CLA guide](https://github.com/kubernetes/community/blob/master/CLA.md)](https://github.com/kubernetes/community/blob/master/CLA.md)? After signing, comment `/easycla` to refresh the status. :)

yindia:
@AshSgDe29071999 Hey, do you have an idea of when the SLA issue might be fixed? I have two PRs currently waiting on this, so I’d appreciate any update when you get a chance.

yindia:
@yankay

This failure isn't caused by your change. `pull-lws-test-e2e-upgrade-main` is broken on `main` right now.

The upgrade test's "before" phase installs the previous released controller (`v0.9.0`) and validates the workloads it created. PR #1002 ("Add leader/worker role label to managed StatefulSets") made `ExpectValidLeaderStatefulSet` in `test/testutils/validators.go:190` require the new `leaderworkerset.sigs.k8s.io/role: leader` label on the leader StatefulSet pod template. The `v0.9.0` controller predates that label and never sets it, so the validator retries until it times out at 120s.

It wasn't caught because this job didn't run on #1002 (it isn't in that PR's check set) and it's non-required (`Required | false` above), so the strict validator was never exercised against the old controller before merge.

I'll open a fix PR that makes the validator tolerate a missing role label on StatefulSets created by an older controller. Once it merges, a rebase (or `/retest`) will clear this job. Nothing to change on your side.

yankay:
> @yankay
> 
> This failure isn't caused by your change. `pull-lws-test-e2e-upgrade-main` is broken on `main` right now.
> 
> The upgrade test's "before" phase installs the previous released controller (`v0.9.0`) and validates the workloads it created. PR #1002 ("Add leader/worker role label to managed StatefulSets") made `ExpectValidLeaderStatefulSet` in `test/testutils/validators.go:190` require the new `leaderworkerset.sigs.k8s.io/role: leader` label on the leader StatefulSet pod template. The `v0.9.0` controller predates that label and never sets it, so the validator retries until it times out at 120s.
> 
> It wasn't caught because this job didn't run on #1002 (it isn't in that PR's check set) and it's non-required (`Required | false` above), so the strict validator was never exercised against the old controller before merge.
> 
> I'll open a fix PR that makes the validator tolerate a missing role label on StatefulSets created by an older controller. Once it merges, a rebase (or `/retest`) will clear this job. Nothing to change on your side.

Thanks @AshSgDe29071999 

There is an https://github.com/kubernetes-sigs/lws/issues/1016 opened .

yindia:
Opened #1017 with the validator fix. Once it merges, a rebase (or `/retest pull-lws-test-e2e-upgrade-main`) will clear the job on this PR. No change needed here.

Do you want to keep this PR open and rebase after #1017 lands, or close it and reopen once the job is green?

yankay:
> Opened #1017 with the validator fix. Once it merges, a rebase (or `/retest pull-lws-test-e2e-upgrade-main`) will clear the job on this PR. No change needed here.
> 
> Do you want to keep this PR open and rebase after #1017 lands, or close it and reopen once the job is green?

If it does not affect CI, this PR is not urgent. We can keep it open, but since the change is small and does not block other work, another PR may incorporate it later.

yindia:
@yankay this looks good to me, lets merge this

PR #995, opened 2026-08-20
Only group pod affinity

gaogaoSpark:
**What type of PR is this?**

Feature

Adds support for the `share-topology` annotation, which enables co-location scheduling for pods within the same replica without enforcing anti-affinity between different replicas.

Compared to the existing `exclusive-topology`:
- `exclusive-topology`: Pods in the same replica share a topology domain + anti-affinity between different replicas
- `share-topology`: Pods in the same replica share a topology domain, but different replicas can share the same domain

**Which issue(s) this PR fixes**

Fixes #

**Special notes for your reviewer**

- Introduces new annotation: `leaderworkerset.sigs.k8s.io/share-topology`
- Adds webhook validation to prevent setting both `exclusive-topology` and `share-topology` simultaneously
- Implements `SetShareAffinities` function that only sets pod affinity (no anti-affinity)

**Does this PR introduce a user-facing change?**

```release-note
Added support for share-topology annotation to enable co-location scheduling within the same replica without anti-affinity between different replicas.
```


ahg-g:
This is an API change, can we at least create a tracking issue to discuss the use case the solution. 

We need to move away from annotations, the existing annotation was supposed to be a stop gap, we should discuss a proper spec for this.

gaogaoSpark:
> This is an API change, can we at least create a tracking issue to discuss the use case the solution.
> 
> We need to move away from annotations, the existing annotation was supposed to be a stop gap, we should discuss a proper spec for this.

Here is a GitHub issue-style version:

---

## Requirement

Pods within the same replica need rack-level affinity, while replicas do not require rack-level anti-affinity.

In other words, we want all pods belonging to one replica to be scheduled within the same rack, but different replicas are allowed to share the same rack.

## About the API Change Concern

I don’t think this necessarily changes the API. It is more about supporting an additional scheduling scenario while keeping the implementation approach consistent with the existing annotation:

```text
leaderworkerset.sigs.k8s.io/exclusive-topology
```

If the existing annotation is intended to be only a temporary workaround, then the proper solution may require a larger design change. In that case, the proposed annotation:

```text
leaderworkerset.sigs.k8s.io/share-topology
```

can also be adjusted later to align with the new design.

## Use Case

### Scenario

Large language model deployment with PD disaggregation.

### Cluster topology

The cluster consists of:

```text
4 racks * 18 GB300 nodes
```

Each node has:

```text
4 GPUs
```

### Decode workload

For the decode workload:

```text
data-parallel-size = 16
```

This means one decode replica requires:

```text
16 GPUs = 4 nodes
```

Since one rack contains 18 nodes, multiple decode replicas need to be deployed, and up to 4 decode replicas can be placed within the same rack.

## Expected Behavior

- Pods within the same decode replica should be scheduled in the same rack.
- Different decode replicas do not need rack-level anti-affinity.
- Multiple decode replicas should be allowed to share the same rack when resources are available.

## Motivation

The existing `exclusive-topology` behavior enforces stronger isolation than needed for this use case. For PD-disaggregated LLM serving, decode replicas only require intra-replica rack affinity for better locality and performance, but enforcing inter-replica rack anti-affinity reduces scheduling flexibility and may lead to inefficient resource utilization.

---

ahg-g:
I suggest we address https://github.com/kubernetes-sigs/lws/issues/1064 first and then add this feature as a placement policy type.

gaogaoSpark:
> I suggest we address #1064 first and then add this feature as a placement policy type.

ok

PR #996, opened 2026-08-20
fix: support budget-driven combined rollout and scale-up

### LINKED ISSUE #717: Modifying GPU requirement with replica count change did not reflect in existing replica/ didn't succeed the deployment
**What happened**:

I have an EKS cluster with LWS operator deployed. I deployed a single replica with 8 GPU as resource requirement. When I made a change to the yaml with 2 replica and reduce my GPU requirement to 4 the existing deployed replica and the pod didn't change. 

This caused the second replica to not get deployed even if the total GPU requirement was satisfied.

The existing replica and its pods to get recreated and both replica and its pods to be deployed successfully.

- Should be easy to repro with first creating a single replica of model deployment and then making changes with 2 replica count and reducing GPU resource from 8 to 4 for each replica.

No

**Environment**:

- Kubernetes version (use `kubectl version`): 1.32
- LWS version (use `git describe --tags --dirty --always`): [v0.7.0](https://github.com/kubernetes-sigs/lws/releases/tag/v0.7.0) 
- Cloud provider or hardware configuration: AWS/EKS + GPU
- OS (e.g: `cat /etc/os-release`): AL2023

chenshoval:
#### What type of PR is this?

/kind bug
/kind api-change

This fixes combined template updates and scale-up getting stuck when the cluster does not have enough spare capacity. For example, replacing one 8-GPU group with two 4-GPU groups fits the same cluster, but creating the additional group before replacing the old one needs extra capacity that may not exist.

The PR now uses the existing `maxUnavailable` and `maxSurge` settings. It removes the proposed `updateOrder` field; there is no new `ScaleFirst` or `RolloutFirst` option.

It follows the same idea as DaemonSet rolling updates: replace an old group when the availability budget allows it, even if an additional group is stuck Pending. For LWS, the leader and all its workers must be Ready before the group counts as available. It is not an exact copy of DaemonSet behavior: LWS keeps StatefulSet RollingUpdate and partition protection.

In practice:

- With `maxUnavailable: 1` and `maxSurge: 0`, an old group can be replaced to free capacity while an additional group is Pending.
- With `maxUnavailable: 0` and `maxSurge: 1`, a healthy old group is kept until enough whole groups are Ready to replace it without dropping below the availability floor. A Ready leader with an unready worker is not enough.
- The controller records replacement reservations before allowing disruption, so repeated reconciliations do not spend the same availability credit twice. Leader deletion uses UID/resourceVersion checks and normal worker garbage collection.

Fixes #717

Design discussion: #1018. DaemonSet reference: [KEP-1591](https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/1591-daemonset-surge).

**This changes the default behavior for combined updates.** The existing default `maxUnavailable: 1` can now allow old-group replacement while additions are Pending. It is not an opt-in ordering setting, and it does not promise that every rollout happens in a fixed order. Ordinary scaling and rollouts outside the combined-update path keep their existing handling.

The availability floor is `max(0, min(initial non-surge replicas, desired replicas) - resolved maxUnavailable)`. The initial baseline is retained across superseding updates. Percentages use the desired replica count, with the existing LWS rounding rules. This bounds controller-authorized disruption, not independent Pod failures.

There is a known partition limitation: updating an unavailable lower ordinal may also expose a healthy higher ordinal to native replacement. If the budget cannot cover that, the controller waits and reports `CombinedRolloutBudgetBlocked`. This does not provide DaemonSet-style independent replacement of any group.

**Validation**

- On the rebased commit `692ab326`, CI unit and verification jobs passed. All eight E2E jobs passed: Kubernetes 1.33 through 1.36, cert-manager, gang scheduling, and both upgrade paths. These are the existing CI suites, not proof of every new combined-rollout edge case.
- On the follow-up commit `365b850`, **integration CI passed: 53 controller specs and 94 webhook specs, with no failures or skips** ([passing run](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_lws/996/pull-lws-test-integration-main/2097961912057729024)). The three failures on the previous commit were fixed in the test simulator: supply the leader DNS identity normally assigned by Kubernetes, and wait for the resized template to be published before simulating readiness. No production code or availability assertions were changed. Unit and verification CI also passed on this commit; E2E reruns were still in progress when this note was updated.
- Before the rebase, native AKS 1.35.7 tests passed for constrained-capacity replacement, zero-unavailability protection, whole-group readiness credit, worker cleanup and final convergence. These used size-two groups with 300m CPU per old Pod and 175m per new Pod, constrained to 750m available capacity. They tested the budget-driven implementation, not the earlier ordering prototype.
- Before the rebase, local validation passed 602 unit tests, 23 selected integration tests and Go lint. Those are separate, earlier results; the current full integration result is listed above.
- The additional native restart/partition scenarios remain incomplete. The restart harness checked leadership before lease handoff completed; later observations confirmed reservation preservation, but final convergence and the partition scenario were not completed.

The description has been updated to replace the original ordering proposal. Earlier discussion and test results for that prototype are historical. #1018 remains open for maintainer agreement on the approach.

```release-note
Combined LeaderWorkerSet template updates and scale-up now use the existing maxUnavailable and maxSurge budgets to decide when old groups can be replaced. Pending additions no longer block an affordable replacement, and availability credit requires the leader and all workers to be Ready. With the default maxUnavailable of 1, this can replace old groups earlier than before. No updateOrder field is added; StatefulSet RollingUpdate and partition protection are retained.
```

### DIFF
--- a/api/leaderworkerset/v1/leaderworkerset_types.go
+++ b/api/leaderworkerset/v1/leaderworkerset_types.go
@@ -321,6 +321,18 @@ const (
 
 // RollingUpdateConfiguration defines the parameters to be used for RollingUpdateStrategyType.
 type RollingUpdateConfiguration struct {
+	// updateOrder controls whether existing replicas are updated before scaling up
+	// when the pod template and replica count increase in the same update.
+	// ScaleFirst preserves the existing behavior of creating the additional replicas
+	// before updating existing replicas. RolloutFirst updates existing replicas before
+	// creating the additional replicas, allowing their old resources to be released.
+	// The default value is ScaleFirst.
+	//
+	// +optional
+	// +kubebuilder:default=ScaleFirst
+	// +kubebuilder:validation:Enum={ScaleFirst,RolloutFirst}
+	UpdateOrder UpdateOrderType `json:"updateOrder,omitempty"`
+
 	// partition indicates the ordinal at which the lws should be partitioned for updates.
 	// During a rolling update, all the groups from ordinal Partition to Replicas-1 will be updated.
 	// The groups from 0 to Partition-1 will not be updated.
@@ -366,6 +378,16 @@ type RollingUpdateConfiguration struct {
 	// +kubebuilder:default=0
 	MaxSurge intstr.IntOrString `json:"maxSurge,omitempty"`
 }
+
+type UpdateOrderType string
+
+const (
+	// ScaleFirst increases the replica count before updating existing replicas.
+	ScaleFirstUpdateOrder UpdateOrderType = "ScaleFirst"
+
+	// RolloutFirst updates existing replicas before increasing the replica count.
+	RolloutFirstUpdateOrder UpdateOrderType = "RolloutFirst"
+)
 
 type RolloutStrategyType string
 
--- a/api/leaderworkerset/v1/leaderworkerset_types.go
+++ b/api/leaderworkerset/v1/leaderworkerset_types.go
@@ -321,18 +321,6 @@ const (
 
 // RollingUpdateConfiguration defines the parameters to be used for RollingUpdateStrategyType.
 type RollingUpdateConfiguration struct {
-	// updateOrder controls whether existing replicas are updated before scaling up
-	// when the pod template and replica count increase in the same update.
-	// ScaleFirst preserves the existing behavior of creating the additional replicas
-	// before updating existing replicas. RolloutFirst updates existing replicas before
-	// creating the additional replicas, allowing their old resources to be released.
-	// The default value is ScaleFirst.
-	//
-	// +optional
-	// +kubebuilder:default=ScaleFirst
-	// +kubebuilder:validation:Enum={ScaleFirst,RolloutFirst}
-	UpdateOrder UpdateOrderType `json:"updateOrder,omitempty"`
-
 	// partition indicates the ordinal at which the lws should be partitioned for updates.
 	// During a rolling update, all the groups from ordinal Partition to Replicas-1 will be updated.
 	// The groups from 0 to Partition-1 will not be updated.
@@ -348,7 +336,7 @@ type RollingUpdateConfiguration struct {
 	Partition *int32 `json:"partition,omitempty"`
 
 	// maxUnavailable is the maximum number of replicas that can be unavailable during the update.
-	// Value can be an absolute number (ex: 5) or a percentage of total replicas at the start of update (ex: 10%).
+	// Value can be an absolute number (ex: 5) or a percentage of desired spec.replicas (ex: 10%).
 	// Absolute number is calculated from percentage by rounding down.
 	// This can not be 0 if MaxSurge is 0.
 	// By default, a fixed value of 1 is used.
@@ -357,6 +345,11 @@ type RollingUpdateConfiguration struct {
 	// can be scaled down further, followed by scaling up the new replicas, ensuring
 	// that at least 70% of original number of replicas are available at all times
 	// during the update.
+	// During combined template updates and scale-up, controller-authorized disruption
+	// uses a floor of max(0, min(initial non-surge replicas, desired replicas)-maxUnavailable).
+	// Whole Ready additional groups provide credit; Pending additions do not block
+	// affordable old-group replacement. This is not a guarantee against independent
+	// failures or a strict ordering policy. An unaffordable old suffix can block progress.
 	//
 	// +kubebuilder:validation:XIntOrString
 	// +kubebuilder:default=1
@@ -365,7 +358,7 @@ type RollingUpdateConfiguration struct {
 	// maxSurge is the maximum number of replicas that can be scheduled above the original number of
 	// replicas.
 	// Value can be an absolute number (ex: 5) or a percentage of total replicas at
-	// the start of the update (ex: 10%).
+	// desired spec.replicas (ex: 10%).
 	// Absolute number is calculated from percentage by rounding up.
 	// By default, a value of 0 is used.
 	// Example: when this is set to 30%, the new replicas can be scaled up by 30%
@@ -378,16 +371,6 @@ type RollingUpdateConfiguration struct {
 	// +kubebuilder:default=0
 	MaxSurge intstr.IntOrString `json:"maxSurge,omitempty"`
 }
-
-type UpdateOrderType string
-
-const (
-	// ScaleFirst increases the replica count before updating existing replicas.
-	ScaleFirstUpdateOrder UpdateOrderType = "ScaleFirst"
-
-	// RolloutFirst updates existing replicas before increasing the replica count.
-	RolloutFirstUpdateOrder UpdateOrderType = "RolloutFirst"
-)
 
 type RolloutStrategyType string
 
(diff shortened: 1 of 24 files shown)


bhks:
Thank you so much for coming up with this flag based clever idea.

chenshoval:
@Edwinhr716 @kerthcet is it possible to get a review for this before ill need another rebase?

chenshoval:
@kerthcet @Edwinhr716 @yankay copilot found added no comments twice.
can we move forward please?
much appreciated!

yankay:
HI @chenshoval 


Thanks for working on this. Since this PR adds a public API field and new rollout semantics, I think it needs a KEP or ISSUE for broader design discussion. I’m also not yet sure that `updateOrder` is the right abstraction, since there seems to be little precedent for it in other Kubernetes subprojects.

chenshoval:
Thanks @yankay , that makes sense. I agree that the problem and rollout semantics should be discussed before we commit to a public updateOrder API.

I will add a structured design proposal to #717 comparing the explicit ordering policy with a budget-driven approach using the existing maxUnavailable and maxSurge semantics. The goal is to support both capacity-constrained updates that must release old resources and demand-driven scale-up that must preserve existing availability. Please let me know if you would prefer a separate enhancement issue.

I will keep this PR and pause API changes until we agree on the abstraction. If the discussion indicates that a KEP is required, I can draft one using the agreed direction.

chenshoval:
Thanks @yankay. I opened #1018 to discuss the rollout semantics:

The proposal compares the current explicit updateOrder prototype with a budget-driven design using maxUnavailable and maxSurge, including compatibility, StatefulSet accounting, and both constrained-capacity and demand-driven scale-up cases.

I’ll keep this PR as the validated prototype and pause implementation changes until we agree on the design.

chenshoval:
@yankay I’ve changed the PR to use the existing maxUnavailable and maxSurge settings instead of adding updateOrder.

It follows the same idea as DaemonSet rolling updates: we can replace an old group when the budget allows it, even if a new group is stuck Pending. For LWS, the leader and all its workers have to be Ready before the group counts as available. We still keep StatefulSet partition behavior, so it’s not exactly the same as DaemonSet.

The main cases passed on AKS. Restart and partition testing isn’t finished yet. Could you take another look? I’ll leave this issue open until the maintainers agree on the approach.

P.S no more API changes so it is safer to look into.

chenshoval:
@Edwinhr716 @kerthcet @yankay  would appreciate a review

aliceco01:
Read this end to end. The core planner is careful work: the reservation model, the decode hardening on persisted state, and the whole-group readiness definition that refuses credit from terminating pods, stale owners and stale aggregate status all look right to me. The behavior change is declared in the release note and in bold in the PR description, which I appreciate, since that was my first concern going in.

Four things, roughly in order of how much they would hold me up.

**1. `Fixes #717` holds only for ordinal group identity.** `Reconcile` returns through `reconcileHash` before the combined hook is reached, and `reconcileCombinedRollout` returns `handled=false` when `sts == nil`, so an LWS with `groupIdentity: Hash` never enters this path. That is the right call for the implementation, since the whole design rests on StatefulSet partition semantics, but it means a user who hits #717 with hash identity gets no fix and no signal that the feature does not apply to them. #877 took the other route for a comparable limitation and rejected the combination in the webhook with an explicit message. Could the scope at least be stated in the docs page and the release note, or is a webhook warning worth it here?

**2. What clears an unusable persisted state?** The state lives in the `leaderworkerset.sigs.k8s.io/combined-rollout` annotation on the leader StatefulSet, and `combinedModelPlanUpdate` refuses to proceed on a state whose `Generation` is ahead of the observed one: "invalid or newer persisted state; do not apply or delete". The caller propagates that error, so as far as I can tell the rollout then stalls and re-errors on every reconcile with nothing that clears it. Annotations are user-writable and survive backup and restore, so this is reachable without a controller bug: restoring a StatefulSet from a backup, or a well-meaning `kubectl annotate`, wedges the rollout. Is there a recovery path I missed? If not, would you consider either surfacing it in a condition so an operator can see why nothing is moving, or treating an unusable state as "start over" rather than "stop"?

Related and more of an architectural question for the maintainers than for you: is an annotation on a child object the right home for controller state at all, versus `.status` on the LWS? I can see the argument that it is co-located with the partition it fences, and #877 also reaches for an annotation, so I would rather ask than assert.

**3. `docs/combined-rollout-investigation.md` is 278 lines of design rationale in `docs/`.** Design documents in this repo live in `keps/`, and #1018 is the open design issue for exactly this. Merging it under `docs/` puts a second design surface next to the KEP directory that nothing keeps in sync. Would it be better as `keps/1018-combined-rollout/`, or dropped from this PR and left in the issue?

**4. The PR removes 354 lines from `test/integration/controllers/leaderworkerset_test.go`.** Some of that is clearly correct: the deleted case asserted the old contract directly, including "When scaling up the Replicas, Partition will not change, so the new created Pods will apply with the new template" with the partition pinned at 4, and that contract is what this PR changes. Since the new coverage lands in separate files, though, it is hard to check by reading the diff whether every deleted assertion has an equivalent. Could you note in the PR description which new test covers each removed scenario, or keep the old cases with updated expectations where the scenario still applies?

Nothing above is a blocker on the planner logic itself. I would be glad to re-read once 1 and 2 have an answer.

chenshoval:
@aliceco01 sorry for the slow reply, and thanks for going through this in detail. I tested the state and Hash paths against the PR's current code with a real API server and production reconciliation.

1. You're right about the scope. This path only applies to `groupIdentity: Ordinal` (the default). Hash still uses its existing Deployment path. The test confirms that dispatch, but it doesn't establish that Hash has the same scheduling deadlock. I'm making the Ordinal-only scope explicit in the rollout docs and release note rather than rejecting an otherwise supported Hash configuration.

2. You didn't miss a recovery path. Malformed JSON, an unsupported state version, a missing phase, and a generation ahead of the LWS all remained blocked across fresh reconciler instances. The StatefulSet, Pods and LWS conditions stayed unchanged, and no Warning event was emitted. Restoring the exact known-good state resumed the reserved replacement. That is a test control, not advice to delete the annotation: starting over would lose the original baseline and outstanding reservations and could allow another disruption before the first replacement is ready. I agree the current diagnostic is inadequate; the future-generation error even ends in `<nil>`. I'd keep the fail-closed behavior and add an explicit invalid-state condition or warning rather than silently reset it. That diagnostic change is not implemented yet. Keeping the state with the partition allows them to be written atomically, but I'm happy to discuss the longer-term placement with the maintainers.

3. Agreed on avoiding two design documents. I'm removing the investigation write-up from this PR and keeping the design discussion in #1018. The user-facing behavior and limitations will stay in the rollout docs.

4. The four replaced scenarios still call `testCombinedScaleSurge`: template update plus growth, growth during rollout, template update plus growth with surge, and growth followed by downscale during a surge rollout. I'll add that mapping to the description. It isn't a one-for-one replacement, though: the two surge fixtures now use size-one groups, and the helper doesn't reproduce the earlier partially completed rollout or every intermediate status assertion. Those are coverage gaps, not assertions I'm claiming are equivalent.

The five review-characterization cases and the 43 existing focused combined-rollout tests passed locally. These are API-server tests with simulated native-controller acknowledgements, not new cluster E2E results. I'm also rebasing and rerunning checks; I'll keep those results separate from the tests above.

PR #997, opened 2026-08-21
Remove unnecessary pods create permission from controller RBAC

### LINKED ISSUE #994: Remove unnecessary Pod create RBAC permission
**What happened**:

#323 removed `pods/create` from generated RBAC files but missed the kubebuilder marker in `pod_controller.go`. Therefore, #325 regenerated `config/rbac/role.yaml` with `pods/create`, leaving Helm and Kustomize inconsistent.

LWS does not create Pods directly; the StatefulSet controller does.

**Expected fix**:

- Remove `create` from the Pod RBAC marker.
- Keep `create` for Services.
- Regenerate and test the manifests.
- Verify LWS works without Pod creation permission.

This should preferably follow #952, so CI keeps Helm and Kustomize RBAC synchronized.

aliceco01:
#### What type of PR is this?

/kind cleanup

Removes the `create` verb from the `pods` RBAC marker in `pkg/controllers/pod_controller.go` and regenerates `config/rbac/role.yaml`. LWS never creates Pods directly — the leader Pod's reconcile creates a worker **StatefulSet**, and Pods for both leader and workers are created by the built-in StatefulSet controller under its own RBAC. LWS only needs `get`, `list`, `watch`, `patch`, `update`, and `delete` on Pods.

Verified no code path creates Pods:
- `grep -rn "\.Create(" pkg/` — every production hit constructs a non-Pod object (StatefulSet, Service, ControllerRevision, PodGroup, DisaggregatedSetRoleScaler, LeaderWorkerSet)
- `grep -rn "&corev1.Pod{" pkg/` — the only non-test Pod construction is a transient in-memory object for a resource-sum calculation, never sent to the API server
- No `CreateOrUpdate`/`CreateOrPatch` usage
- The one production server-side-apply (`leaderworkerset_controller.go:407`) applies a StatefulSet

Background: this permission was already removed once in `a34330a` ("Remove pods create RBAC.", Jan 2025), but by hand-editing the generated YAML and Helm chart without updating the marker. A later `make manifests` in an unrelated PR regenerated `role.yaml` from the stale marker and silently reintroduced `create`. This PR fixes the marker — the source of truth — so regeneration stays correct. The Helm chart needs no change; `charts/lws/templates/rbac/clusterrole.yaml` still reflects the correct state

Fixes #994

```release-note
NONE
```

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -18,9 +18,7 @@ rules:
   - ""
   resources:
   - pods
-  - services
   verbs:
-  - create
   - delete
   - get
   - list
@@ -40,6 +38,18 @@ rules:
   verbs:
   - get
   - list
+  - update
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - services
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
   - update
   - watch
 - apiGroups:
--- a/pkg/controllers/pod_controller.go
+++ b/pkg/controllers/pod_controller.go
@@ -62,7 +62,7 @@ func NewPodReconciler(client client.Client, schema *runtime.Scheme, record event
 
 //+kubebuilder:rbac:groups="",resources=events,verbs=create;watch;update;patch
 //+kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;watch;update;patch
-//+kubebuilder:rbac:groups=core,resources=pods,verbs=create;delete;get;list;patch;update;watch
+//+kubebuilder:rbac:groups=core,resources=pods,verbs=delete;get;list;patch;update;watch
 //+kubebuilder:rbac:groups=core,resources=pods/finalizers,verbs=update
 //+kubebuilder:rbac:groups=core,resources=nodes,verbs=get;list;watch;update;patch
 


ahg-g:
Are these permissions updates aligned with the helm chart?

yankay:
Hi @aliceco01 

Could you add Fixes #994 and a `release-note` block with `NONE` to the PR description?

This is a historical issue: #323 removed the permission from the generated RBAC files, but the kubebuilder marker was not updated, so #325 later regenerated it. Thanks! :)

aliceco01:
@yankay Thanks for the review, Will update the PR description with `Fixes #994` and a `release-note` block set to `NONE` shortly. Appreciate the pointer to the #323/#325 history, that matches what I found when tracing how the permission got reintroduced.

@ahg-g Yes, the Helm chart is already aligned. The `pods` rule in the chart grants exactly `delete, get, list, patch, update, watch` (no `create`):

https://github.com/kubernetes-sigs/lws/blob/060ab4da17eea1f0865fcbd3236bb1075e2cea32/charts/lws/templates/rbac/clusterrole.yaml#L41-L51

The chart was fixed by hand in #323 and never regressed, since it's maintained manually rather than regenerated from the kubebuilder markers. Only the marker and the generated `config/rbac/role.yaml` were stale which is what this PR fixes. After merge, the marker, `role.yaml`, and the chart all agree, and can be confirmed with:

```
grep -A10 'pods$' charts/lws/templates/rbac/clusterrole.yaml config/rbac/role.yaml
```

yankay:
pending #992

aliceco01:
Rebased onto latest `main` to pick up a docs template change that was causing `pull-lws-verify-main` to fail (unrelated to this PR's diff , regenerated `site/content/en/docs/reference/*.v1.md` was missing `linkTitle`/`weight` front matter that main had already updated). No changes to the actual RBAC fix itself.

@ahg-g the rebase caused the `lgtm` label to be auto-removed by Prow, flagging so you can re-confirm when you get a chance.

PR #999, opened 2026-08-22
Fix RecreateGroupOnPodRestart deletion race

### LINKED ISSUE #998: RecreateGroupOnPodRestart is skipped when the worker StatefulSet recreates the deleted pod first
/kind bug

**What happened**:

Deleting a single worker pod from a group with `restartPolicy: RecreateGroupOnPodRestart` sometimes does not recreate the group. The deleted worker comes back with a new UID, but the leader and the other worker keep theirs and keep running.

We see this intermittently in Kueue's e2e on kind. Two runs:

- https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/14646/pull-kueue-test-e2e-extended-shard-1-release-0-18-1-36/2090160363050897408
- https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/14674/pull-kueue-test-e2e-extended-shard-1-release-0-18-1-35/2090384639033085952

From the first run, namespace `lws-e2e-md9rg`. The worker StatefulSet creates `lws-0-2`, then creates it again 1.2s later after we delete it:

```
19:43:00.152  statefulset-controller  SuccessfulCreate  Create Pod lws-0-2 in StatefulSet lws-0 successful
19:43:01.352  statefulset-controller  SuccessfulCreate  Create Pod lws-0-2 in StatefulSet lws-0 successful
```

The lws-controller-manager log has no `RecreateGroupOnPodRestart` event for that namespace at any point. It emits the event normally for two other namespaces seconds either side:

```
19:42:52  Worker pod lws-0-1 failed, deleted leader pod lws-0 to recreate group 0   ns=lws-e2e-7fznl
19:43:05  Worker pod lws-0-1 failed, deleted leader pod lws-0 to recreate group 0   ns=lws-e2e-2fhgp
```

At namespace teardown the surviving pods confirm the leader was never

```
Pod lws-0    uid 772cd5d7-4921-437d-8cb0-1458ce9af6a0   (unchanged)
Pod lws-0-1  uid c320e402-9433-417d-98e3-8ea30dbe5344   (unchanged)
Pod lws-0-2  uid 6edc6d99-4b2c-4f4a-af9e-71dff074f34a   (replaced)
```

Second run is the same shape: no `RecreateGroupOnPodRestart` event for emitted for others in the same log.

Our pods use `agnhost netexec` with `terminationGracePeriodSeconds: 1`IGTERM and the pod is gone quickly. We suspect the pod's terminatingwindow is short enough that the reconcile for the deletion lands after the StatefulSet has already recreated a pod of the same name, but we haven't confirmed that and
would rather leave the diagnosis to you.

Deleting a worker pod triggers group recreation, regardless of how quickly the StatefulSet replaces it.

1. Create an LWS with `restartPolicy: RecreateGroupOnPodRestart`, sizes on SIGTERM (`agnhost netexec`) with `terminationGracePeriodSeconds:
1`.
2. Wait for all groups ready.
3. `kubectl delete pod <lws>-0-2`.
4. Repeat on a loaded cluster. Intermittently the leader and the othero `RecreateGroupOnPodRestart` event is emitted.

Adding a few seconds of delay before the pod reconcile makes it reproduce every time.

**Environment**:

- Kubernetes version (use `kubectl version`): v1.35.5 and v1.36.1 (kin
- LWS version (use `git describe --tags --dirty --always`): v0.8.0
- Cloud provider or hardware configuration: kind on CI
- OS (e.g: `cat /etc/os-release`): n/a
- Kernel (e.g. `uname -a`): n/a
- Install tools: kind
- Others:

poorpaper:
#### What type of PR is this?

/kind bug

A worker StatefulSet may recreate a deleted Pod with the same namespace and
name before the Pod controller processes the deletion.

The default controller request only contains the namespaced name. As a result,
the delete and replacement create events may collapse into the same workqueue
key, and reconciliation may read the replacement Pod instead of observing the
deleted Pod. This causes `RecreateGroupOnPodRestart` to be skipped.

This change:

- carries the deleted Pod UID, relevant labels, and controller owner identity
  in a typed reconcile request;
- keeps delete and same-name replacement events as distinct workqueue entries;
- processes deletion metadata without fetching the replacement Pod;
- preserves the existing StatefulSet owner watch behavior;
- preserves the UID-based ownership checks that prevent stale workers from
  deleting a replacement leader.

No API, CRD, or RBAC changes are introduced.

Fixes #998

Validation performed:

- `go test ./pkg/controllers -count=1`
- `go vet ./pkg/controllers`
- `go test ./api/... ./pkg/... ./cmd/... -count=1`
- controller and webhook envtest suites: 144/144 specs passed
- focused Kind E2E for `RecreateGroupOnPodRestart`: 5/5 consecutive runs passed

The focused E2E verification confirmed that all original Pod UIDs in the
group were replaced after deleting a worker Pod.

```release-note
Fixed an issue where `RecreateGroupOnPodRestart` could be skipped if a worker StatefulSet recreated a deleted Pod before reconciliation.
```

### DIFF
--- a/pkg/controllers/pod_controller.go
+++ b/pkg/controllers/pod_controller.go
@@ -61,71 +61,29 @@ type PodReconciler struct {
 	SchedulerProvider schedulerprovider.SchedulerProvider
 }
 
-// podReconcileRequest keeps the identity of a deleted Pod in the workqueue.
-// A StatefulSet can create a replacement with the same namespace and name
-// before reconciliation starts, so a namespaced name alone is insufficient.
+// podReconcileRequest keeps the identity and snapshot of a deleted Pod in the
+// workqueue. A StatefulSet can create a replacement with the same namespace
+// and name before reconciliation starts, so a namespaced name alone is
+// insufficient.
 type podReconcileRequest struct {
 	types.NamespacedName
-	UID             types.UID
-	Deleted         bool
-	LWSName         string
-	WorkerIndex     string
-	GroupIndex      string
-	RevisionKey     string
-	OwnerAPIVersion string
-	OwnerKind       string
-	OwnerName       string
-	OwnerUID        types.UID
+	UID        types.UID
+	DeletedPod *corev1.Pod
 }
 
 func podReconcileRequestForPod(pod *corev1.Pod, deleted bool) podReconcileRequest {
 	request := podReconcileRequest{
 		NamespacedName: client.ObjectKeyFromObject(pod),
 		UID:            pod.UID,
-		Deleted:        deleted,
 	}
-	if !deleted {
-		return request
-	}
-
-	request.LWSName = pod.Labels[leaderworkerset.SetNameLabelKey]
-	request.WorkerIndex = pod.Labels[leaderworkerset.WorkerIndexLabelKey]
-	request.GroupIndex = pod.Labels[leaderworkerset.GroupIndexLabelKey]
-	request.RevisionKey = pod.Labels[leaderworkerset.RevisionKey]
-	if owner := metav1.GetControllerOf(pod); owner != nil {
-		request.OwnerAPIVersion = owner.APIVersion
-		request.OwnerKind = owner.Kind
-		request.OwnerName = owner.Name
-		request.OwnerUID = owner.UID
+	if deleted {
+		request.DeletedPod = pod.DeepCopy()
+		if request.DeletedPod.DeletionTimestamp == nil {
+			deletionTimestamp := metav1.Now()
+			request.DeletedPod.DeletionTimestamp = &deletionTimestamp
+		}
 	}
 	return request
-}
-
-func (r podReconcileRequest) deletedPod() corev1.Pod {
-	deletionTimestamp := metav1.Now()
-	pod := corev1.Pod{ObjectMeta: metav1.ObjectMeta{
-		Name:              r.Name,
-		Namespace:         r.Namespace,
-		UID:               r.UID,
-		DeletionTimestamp: &deletionTimestamp,
-		Labels: map[string]string{
-			leaderworkerset.SetNameLabelKey:     r.LWSName,
-			leaderworkerset.WorkerIndexLabelKey: r.WorkerIndex,
-			leaderworkerset.GroupIndexLabelKey:  r.GroupIndex,
-			leaderworkerset.RevisionKey:         r.RevisionKey,
-		},
-	}}
-	if r.OwnerUID != "" {
-		controller := true
-		pod.OwnerReferences = []metav1.OwnerReference{{
-			APIVersion: r.OwnerAPIVersion,
-			Kind:       r.OwnerKind,
-			Name:       r.OwnerName,
-			UID:        r.OwnerUID,
-			Controller: &controller,
-		}}
-	}
-	return pod
 }
 
 func NewPodReconciler(client client.Client, schema *runtime.Scheme, record events.EventRecorder, sp schedulerprovider.SchedulerProvider) *PodReconciler {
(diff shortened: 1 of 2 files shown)


yankay:
Thanks for the fix. I verified it with a deterministic informer-to-reconcile test: the identical test fails on the parent commit because the leader remains, and passes consistently on this PR. This confirms the race is fixed.

/lgtm

pkg/controllers/pod_controller.go
```diff
@@ -67,8 +139,14 @@ func NewPodReconciler(client client.Client, schema *runtime.Scheme, record event
 //+kubebuilder:rbac:groups=core,resources=nodes,verbs=get;list;watch;update;patch
 
 func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
ahg-g: now that we have a custom constructor, in what cases does Reconcile get invoked? can it be removed if it is not relevant anymore?

pkg/controllers/pod_controller.go
```diff
@@ -56,19 +61,92 @@ type PodReconciler struct {
...
 }
 
+// podReconcileRequest keeps the identity of a deleted Pod in the workqueue.
+// A StatefulSet can create a replacement with the same namespace and name
+// before reconciliation starts, so a namespaced name alone is insufficient.
+type podReconcileRequest struct {
+	types.NamespacedName
+	UID             types.UID
+	Deleted         bool
+	LWSName         string
```
ahg-g: instead of manually picking specific fields, how about we deep copy the pod object; to minimize the cloning overhead, we do that only in the case the pod was deleted since in the other cases Reconcile will do a Get anyway, so this struct can be something like:

```
type podReconcileRequest struct {
	types.NamespacedName
	UID             types.UID
	Deleted         bool
    Pod corev1.Pod
}
```

The reason is because I am concerned that the logic in reconcile may change to access a field/label/annotation that is not captured here, and so this could be a source of bugs.

pkg/controllers/pod_controller.go
poorpaper: You're right, `Reconcile` isn't used by the controller anymore. I removed it and updated the remaining test in 6f23e18.

pkg/controllers/pod_controller.go
poorpaper: I changed delete requests to keep a `pod.DeepCopy()`. I used `*corev1.Pod` since the typed request needs to stay comparable; non-delete requests still fetch the current Pod. Addressed in 6f23e18.

poorpaper:
@ahg-g  Thanks for the review. I addressed both comments in 6f23e18:

- removed the unused `Reconcile` wrapper and updated its remaining unit-test caller;
- replaced the hand-picked deletion metadata with a complete deleted Pod snapshot, while keeping the typed request comparable by storing it as `*corev1.Pod`.

yankay:
> @ahg-g Thanks for the review. I addressed both comments in 6f23e18:
> 
> * removed the unused `Reconcile` wrapper and updated its remaining unit-test caller;
> * replaced the hand-picked deletion metadata with a complete deleted Pod snapshot, while keeping the typed request comparable by storing it as `*corev1.Pod`.

HI @poorpaper 
It looks like you may have forgotten to push code. Could you please check? Thanks :)

poorpaper:
> > @ahg-g Thanks for the review. I addressed both comments in 6f23e18:
> > 
> > * removed the unused `Reconcile` wrapper and updated its remaining unit-test caller;
> > * replaced the hand-picked deletion metadata with a complete deleted Pod snapshot, while keeping the typed request comparable by storing it as `*corev1.Pod`.
> 
> HI @poorpaper It looks like you may have forgotten to push code. Could you please check? Thanks :)

🥲Sorry about that. I created the commit locally but forgot to push it before replying. It has now been pushed as 6f23e18. Thanks for pointing it out. @yankay

yankay:
Thanks @poorpaper  for the /lgtm!

ahg-g:
/approve

@poorpaper you may need to rebase just to get Trivy to pass

PR #1000, opened 2026-08-22
fix: handle pod replacement race in RecreateGroupOnPodRestart

### LINKED ISSUE #998: RecreateGroupOnPodRestart is skipped when the worker StatefulSet recreates the deleted pod first

amirialy:
#### What type of PR is this?

/kind bug

Fix a race where `RecreateGroupOnPodRestart` can miss a deleted worker Pod when the StatefulSet quickly recreates it with the same name.

Fixes #998

```release-note
NONE

### DIFF
--- a/pkg/controllers/pod_controller.go
+++ b/pkg/controllers/pod_controller.go
@@ -61,7 +61,7 @@ type PodReconciler struct {
 	Record            events.EventRecorder
 	SchedulerProvider schedulerprovider.SchedulerProvider
 	deletedPodsMu     sync.Mutex
-	deletedPods       map[types.NamespacedName][]*corev1.Pod
+	deletedPods       map[types.NamespacedName]*corev1.Pod
 }
 
 func NewPodReconciler(client client.Client, schema *runtime.Scheme, record events.EventRecorder, sp schedulerprovider.SchedulerProvider) *PodReconciler {
@@ -70,8 +70,7 @@ func NewPodReconciler(client client.Client, schema *runtime.Scheme, record event
 		Scheme:            schema,
 		Record:            record,
 		SchedulerProvider: sp,
-		deletedPods:       make(map[types.NamespacedName][]*corev1.Pod),
-	}
+		deletedPods:       make(map[types.NamespacedName]*corev1.Pod)}
 }
 
 //+kubebuilder:rbac:groups="",resources=events,verbs=create;watch;update;patch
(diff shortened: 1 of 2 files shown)


ahg-g:
/close

https://github.com/kubernetes-sigs/lws/pull/999 fixes the issue

PR #1003, opened 2026-08-22
Add Helm-based upgrade E2E test

### LINKED ISSUE #962: Add Helm upgrade E2E coverage

Add a Kind-based Helm upgrade E2E test that:

1. Installs the previous LWS release with Helm.
2. Creates LeaderWorkerSet and DisaggregatedSet workloads.
3. Reconciles CRDs using the documented upgrade procedure.
4. Runs `helm upgrade` with the current chart and controller image.
5. Verifies that the Helm release, controller, and webhooks are healthy, existing workloads are preserved, and new workloads can be created.

Run the test in Prow for changes affecting the Helm chart.

**Why is this needed**:

LWS has previously had Helm upgrade regressions:

- https://github.com/kubernetes-sigs/lws/issues/529 reported a webhook panic caused by an older CRD remaining after upgrade.
- https://github.com/kubernetes-sigs/lws/issues/880 reported that upgrading from v0.7.0 to v0.8.0 deleted the CRD and cascaded to deleting existing LeaderWorkerSets.

The existing upgrade test added by https://github.com/kubernetes-sigs/lws/pull/928 covers manifests/Kustomize, not Helm release upgrades. A real `helm install` followed by `helm upgrade` is needed to cover Helm-specific resource ownership and lifecycle behavior.

yindia:
#### What type of PR is this?

/kind feature
/kind failing-test

We have hit Helm upgrade regressions before (#529, #880), but the upgrade E2E from #928 only covers the kustomize path. Helm behaves differently on upgrade, most importantly it never touches the `crds/` directory, so CRDs have to be reconciled by hand first. Nothing tests that today.

This adds a Kind based Helm upgrade test. It installs the previous released chart, creates LeaderWorkerSet and DisaggregatedSet workloads, applies the current CRDs the way the docs tell users to, runs helm upgrade in place with the freshly built image, and then checks that the release, controller and webhooks are healthy, the old workloads survived, and new ones still come up.

There is no new Go test. The existing phase driven upgrade test only talks to the cluster through the client, so the same before and after phases work for Helm too. Everything Helm specific lives in the shell script.

The changes:

1. `hack/e2e-test.sh` gets a helm branch behind `LWS_UPGRADE_METHOD=helm`. It installs the old chart from the OCI registry and, on upgrade, applies `charts/lws/crds` before running helm upgrade. The manifests path is untouched.
2. A new `test-e2e-upgrade-helm` make target, folded into `test-e2e-upgrade`.
3. `LWS_UPGRADE_FROM_VERSION` now defaults to the newest released git tag instead of a hardcoded value, so both upgrade tests track releases on their own. Still overridable.

Fix #962

The Prow presubmit is not in this repo. The plan is a job running `make test-e2e-upgrade-helm` scoped to changes under `charts/`. Happy to open the test-infra PR once this looks right.

I left the v0.7 to v0.8 CRD move case (the one behind #880) out on purpose. It needs the extra resource policy annotation step, so it fits better as a second entry in the matrix later rather than bloating this one.

Validated as far as a cluster free run allows: `bash -n` passes, the make target parses, `helm show chart` confirms the old chart resolves from the registry, and `helm template` confirms the image split and the DisaggregatedSet webhook render correctly.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,9 @@ INTEGRATION_TARGET ?= ./test/integration/...
 E2E_KIND_VERSION ?= kindest/node:v1.36.1
 CERT_MANAGER_VERSION ?= v1.17.0
 USE_EXISTING_CLUSTER ?= false
-LWS_UPGRADE_FROM_VERSION ?= v0.9.0
+# Newest released tag to upgrade from; falls back to v0.9.0 on shallow clones
+# without tags. Override to pin a specific version.
+LWS_UPGRADE_FROM_VERSION ?= $(shell git tag -l 'v[0-9]*.[0-9]*.[0-9]*' | sort -V | tail -1 | grep . || echo v0.9.0)
 UPGRADE_KIND_CLUSTER_NAME ?= lws-upgrade
 
 # For local testing, we should allow user to use different kind cluster name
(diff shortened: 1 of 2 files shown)


yindia:
@yankay it's ready, Do you have any other feedback ?

yankay:
Overall, this looks good to me. Let's wait for #993 (fixing #992) and rerun the upgrade E2E before merging. Please also follow up with the test-infra PR for the Helm upgrade Prow job.

/lgtm

yindia:
@yankay I’m not getting any response on PR https://github.com/kubernetes-sigs/lws/pull/993. Is there anything I can do to unblock myself?

yindia:
@yankay Now we can merge this PR

PR #1004, opened 2026-08-23
Use LWS ownership for DisaggregatedSet Services

ErikJiang:
#### What type of PR is this?

/kind bug

Makes each DisaggregatedSet `-prv` Service owned by its corresponding LeaderWorkerSet, allowing Kubernetes GC to clean it up with the LWS instead of using manual deletion paths.

Existing Services are migrated to the matching LWS owner. Legacy slice-0 migration waits for Service GC before creating sibling slices.

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/disaggregatedset/disaggregatedset_controller.go
+++ b/pkg/controllers/disaggregatedset/disaggregatedset_controller.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"slices"
 	"strconv"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -47,6 +48,11 @@ type DisaggregatedSetReconciler struct {
 	ScalerManager  *ScalerManager
 }
 
+// legacyServiceGCRequeue paces the retries while a legacy Service is being garbage
+// collected. A finalizer on that Service blocks slice creation for as long as it is
+// held, so this must not turn into a per-second busy loop.
+const legacyServiceGCRequeue = 5 * time.Second
+
 // +kubebuilder:rbac:groups=disaggregatedset.x-k8s.io,resources=disaggregatedsets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=disaggregatedset.x-k8s.io,resources=disaggregatedsets/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=disaggregatedset.x-k8s.io,resources=disaggregatedsets/finalizers,verbs=update
@@ -54,6 +60,8 @@ type DisaggregatedSetReconciler struct {
 // +kubebuilder:rbac:groups=disaggregatedset.x-k8s.io,resources=disaggregatedsetrolescalers/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=leaderworkerset.x-k8s.io,resources=leaderworkersets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=leaderworkerset.x-k8s.io,resources=leaderworkersets/status,verbs=get
+// Needed to set blockOwnerDeletion on the Services this controller hands to a LeaderWorkerSet.
+// +kubebuilder:rbac:groups=leaderworkerset.x-k8s.io,resources=leaderworkersets/finalizers,verbs=update
 // +kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
 
@@ -81,9 +89,23 @@ func (r *DisaggregatedSetReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	revision := disaggregatedsetutils.ComputeRevision(disaggregatedSet.Spec.Roles)
 	sliceCount := int(disaggregatedsetutils.GetSlices(disaggregatedSet))
 
-	// Step 2: Delete LWS/services for slices beyond the desired count (slice
-	// scale-down). Per-revision drained cleanup runs per slice in reconcileSlice.
-	if err := r.cleanupRemovedSlices(ctx, disaggregatedSet, sliceCount); err != nil {
+	// One listing feeds both steps below; the migration only patches Services, so
+	// the slice scale-down still sees an accurate view of the LWS objects.
+	allLWS, err := r.LWSManager.List(ctx, disaggregatedSet, -1, "")
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Migrate Services created by older controllers before any LWS deletion, so
+	// foreground deletion garbage-collects them instead of leaving DS-owned orphans.
+	if err := r.ServiceManager.migrateLegacyServices(ctx, disaggregatedSet, allLWS); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Step 2: Delete LWS for slices beyond the desired count (slice scale-down).
+	// Their Services are garbage-collected through the LWS owner reference.
+	// Per-revision drained cleanup runs per slice in reconcileSlice.
+	if err := r.cleanupRemovedSlices(ctx, disaggregatedSet, allLWS, sliceCount); err != nil {
 		return ctrl.Result{}, err
 	}
 
@@ -114,13 +136,21 @@ func (r *DisaggregatedSetReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	// and recreate an existing deployment. When slices > 1, the pre-slices (label-less)
 	// slice-0 still at the target revision must become slice-aware before any sibling
 	// slice is created, otherwise its slice-agnostic service would also select the
-	// siblings' same-revision pods. Delete the legacy slice-0 objects here (service
-	// first); the slice loop below then recreates slice 0 in slice-aware form. Running
-	// before the loop, and returning on error, keeps the legacy service from ever
-	// coexisting with sibling slices.
+	// siblings' same-revision pods. Delete the legacy slice-0 LWS objects here and
+	// wait for their Services to be garbage-collected before the slice loop recreates
+	// slice 0 and adds siblings.
 	if sliceCount > 1 {
-		if err := r.recreateLegacySlice0(ctx, disaggregatedSet, revision, roleNames); err != nil {
+		pendingServices, err := r.deleteLegacySlice0(ctx, disaggregatedSet, revision, roleNames)
+		if err != nil {
 			return ctrl.Result{}, err
+		}
+		if len(pendingServices) > 0 {
+			// A finalizer on one of these Services stalls slice creation indefinitely,
+			// so name them instead of only reporting that we are waiting.
+			log.Info("Waiting for legacy Service garbage collection before recreating slice 0", "services", pendingServices)
+			r.Record.Eventf(disaggregatedSet, nil, corev1.EventTypeWarning, EventReasonWaitingForServiceGC,
+				"Migrate", "Waiting for legacy Services %v to be garbage-collected before creating slice-aware LeaderWorkerSets", pendingServices)
+			return ctrl.Result{RequeueAfter: legacyServiceGCRequeue}, nil
 		}
 	}
 
@@ -422,7 +452,7 @@ func (r *DisaggregatedSetReconciler) reconcileSlice(
 	}
 	revisionRoles := disaggregatedsetutils.GroupByRevision(allLWS)
 
-	if err := r.ServiceManager.ReconcileServices(ctx, disaggregatedSet, slice, revisionRoles, revision); err != nil {
+	if err := r.ServiceManager.ReconcileServices(ctx, disaggregatedSet, revisionRoles, revision); err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to reconcile services: %w", err)
 	}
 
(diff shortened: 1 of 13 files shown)


pkg/controllers/disaggregatedset/service_manager.go
```diff
@@ -213,6 +231,36 @@ func (manager *ServiceManager) cleanupDrainedServices(
...
+
+	for i := range serviceList.Items {
+		service := &serviceList.Items[i]
+		if disaggregatedsetutils.SliceLabelMatches(service.Labels, slice) {
+			if _, exists := currentRoles[service.Labels[disaggregatedsetv1.RoleLabelKey]]; exists {
+				continue
+			}
+			if !metav1.IsControlledBy(service, deployment) {
+				continue
+			}
```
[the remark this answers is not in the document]
ErikJiang: Fixed. DeleteLegacyService now verifies ownership before deletion. Foreign or ownerless collisions are preserved, and regression tests were added.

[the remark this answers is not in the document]
ErikJiang: Fixed. CleanupRemovedSlices now skips Services not controlled by the DisaggregatedSet. Tests cover owned, foreign, and ownerless Services.

BenAyedMedAla (commented):
Reviewed the ownership-check additions here — the pattern (list once, filter by `metav1.IsControlledBy` in every cleanup path: removed-role, drained-revision, removed-slice, and legacy) is consistent with the same safety model added to `LeaderWorkerSetManager` in #983. The reasoning for decoupling removed-role cleanup from the overall revision-readiness gate — so an orphaned Service for a deleted role doesn't linger just because unrelated roles haven't finished rolling out — makes sense and is well covered by the new foreign/ownerless/removed-role test cases.

Two small things:
1. The PR body's "Fixes #" is empty — was there a tracking issue for this, or should the placeholder be dropped?
2. The new `cleanupRemovedRoleServices` tests all use slice 0 — a case with a removed-role service belonging to a *different* slice than the one being reconciled would make the per-slice scoping explicit (currently only implied via `SliceLabelMatches`).

ahg-g:
Why are we manually deleting the Service objects, the Service should be owned by the object it's lifecycle is associated with, and when that object is deleted the Service should automatically be deleted too. That is how the workers statefulset is managed, it is owned by the leader pod, and when the leader is deleted, it's worker sts and headless service are automatically deleted.

/cc @hasB4K

ErikJiang:
@ahg-g  Good point. 
These are the per-(slice, revision, role) -prv Services created by the DisaggregatedSet controller, not the headless Services managed by the LWS controller. They are currently owned by the DisaggregatedSet, which is why explicit cleanup is needed.
Since each -prv Service maps 1:1 to an LWS, we propose making that LWS its owner and relying on Kubernetes GC when the LWS is deleted. This should remove most manual cleanup for drained revisions, removed slices, and removed roles.
Legacy slice-0 migration may still require synchronous deletion because GC is asynchronous and the slice-agnostic Service must disappear before sibling slices are created.
One point to confirm: should a removed-role Service remain until its LWS finishes draining and is deleted, rather than being removed immediately when the role leaves the spec? If so, we will follow that lifecycle model.

ahg-g:
> One point to confirm: should a removed-role Service remain until its LWS finishes draining and is deleted, rather than being removed immediately when the role leaves the spec? If so, we will follow that lifecycle model.

I don't think the time to delete it matters, leaving it to GC is better than doing this manually

ErikJiang:
@ahg-g 
Thanks, updated accordingly. Each `-prv` Service is now owned by its corresponding LWS, and the manual cleanup paths have been removed. Existing Services are migrated to the LWS owner before deletion, allowing Kubernetes GC to handle cleanup.

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -589,50 +600,78 @@ func (r *DisaggregatedSetReconciler) cleanupDrainedLWS(ctx context.Context, disa
...
+				if err := r.Delete(ctx, service); err != nil && !apierrors.IsNotFound(err) {
+					return false, fmt.Errorf("failed to delete legacy service %s-prv: %w", legacyName, err)
+				}
+				continue
+			}
+			owner := metav1.GetControllerOf(service)
+			if owner == nil || owner.APIVersion != leaderworkersetv1.GroupVersion.String() || owner.Kind != "LeaderWorkerSet" || owner.Name != legacyName {
+				return false, fmt.Errorf("legacy service %s is not controlled by the expected LeaderWorkerSet %s", service.Name, legacyName)
+			}
+			waitingForGC = true
```
ahg-g: We should use foreground deletion of lws to ensure that child objects are deleted first before the lws object itself is deleted. See this example: https://github.com/kubernetes-sigs/lws/blob/main/pkg/controllers/pod_controller.go#L258

pkg/controllers/disaggregatedset/service_manager.go
```diff
@@ -142,124 +150,55 @@ func (manager *ServiceManager) buildService(
...
-	log := logf.FromContext(ctx)
-
-	readyRevisionSet := make(map[string]bool)
-	for _, group := range revisionRoles {
-		if revisionReadyOnAllRoles(group, roleNames) {
-			readyRevisionSet[group.Revision] = true
+	lws *leaderworkersetv1.LeaderWorkerSet,
+) (bool, bool, error) {
+	service := &corev1.Service{}
+	if err := manager.client.Get(ctx, client.ObjectKey{Name: lws.Name + "-prv", Namespace: deployment.Namespace}, service); err != nil {
```
ahg-g: should we create a function that returns the service name, I see we add "-prv" everywhere we create the service name (what does "prv" mean btw?)

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -589,50 +600,78 @@ func (r *DisaggregatedSetReconciler) cleanupDrainedLWS(ctx context.Context, disa
...
-// lingers alongside siblings), then the legacy LWS. The caller runs this before the slice
-// loop, which then recreates slice 0 in slice-aware form. This restarts slice 0 once,
-// which we accept in place of an in-place migration. A legacy slice-0 at an older revision
-// is left alone: the normal rolling update migrates it without a restart.
+// the new siblings' same-revision pods. For each role it transfers ownership of an older
+// DS-owned Service to the legacy LWS, deletes that LWS, and waits for garbage collection
+// before the caller enters the slice loop. This restarts slice 0 once, which we accept in
+// place of an in-place migration. A legacy slice-0 at an older revision is left alone:
+// the normal rolling update migrates it without a restart.
 func (r *DisaggregatedSetReconciler) recreateLegacySlice0(
```
ahg-g: is this function actually "recreateLegacySlice0" or just "deleteLegacySlice0"? if the latter, I recommend to rename the func.

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -576,6 +584,9 @@ func (r *DisaggregatedSetReconciler) cleanupDrainedLWS(ctx context.Context, disa
 		}
 
 		for _, lws := range roles {
+			if _, _, err := r.ServiceManager.transferServiceOwnership(ctx, disaggregatedSet, lws); err != nil {
+				return fmt.Errorf("failed to transfer Service ownership for LWS %s: %w", lws.Name, err)
+			}
 			log.Info("Deleting drained LWS", "name", lws.Name)
 			if err := r.LWSManager.Delete(ctx, disaggregatedSet.Namespace, lws.Name); err != nil {
```
ahg-g: make sure this delete does a foreground deletion

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -452,6 +457,9 @@ func (r *DisaggregatedSetReconciler) cleanupRemovedSlices(ctx context.Context, d
 		if parseErr != nil || sliceIdx < desiredSlices {
 			continue
 		}
+		if _, _, err := r.ServiceManager.transferServiceOwnership(ctx, disaggregatedSet, lws); err != nil {
+			return fmt.Errorf("failed to transfer Service ownership for LWS %s: %w", lws.Name, err)
+		}
```
ahg-g: instead of doing this everywhere we call `r.LWSManager.Delete`, can we do the transfer of ownership inside Delete? this helps ensure proper coverage.

ahg-g (commented):
thanks!

BenAyedMedAla (commented):
Went through the redesign fully. It correctly implements the GC-based ownership Abdullah suggested, and the hard part — `recreateLegacySlice0` waiting on asynchronous GC before creating slice-aware siblings — is handled correctly across multiple reconcile passes (verified by tracing it through: transfer → delete → wait → confirm gone → proceed), with a sensible fallback for the DS-owned-directly case.

Worth calling out: removing the old "no flip-flop" test isn't a coverage loss — a Service's lifetime is now bounded by its LWS's lifetime via GC, so that scenario is structurally impossible rather than something that needs separate testing.

My earlier comments on the previous version (empty "Fixes #", slice-scoping tests) no longer apply — that code doesn't exist anymore. Good rework.

yankay:
HI @ErikJiang 

Thanks for reworking the Service ownership and migration flow.

**Could we wait for the legacy Service to reach `NotFound` before creating slice-aware siblings?**

I reproduced this on Kind v1.35.0 with a finalizer on the legacy Service. `Delete` only moved it to `Terminating`, but reconciliation continued and created the slice-aware LWS objects and Pods. Since the legacy selector is slice-agnostic, its EndpointSlice contained prefill Pods from both slice 0 and slice 1, violating the KEP isolation invariant.

Could we requeue after deletion and proceed only after a subsequent `Get` returns `NotFound`? A controller/integration test could hold the Service with a finalizer, verify that no sibling is created while it is terminating, then remove the finalizer and verify that reconciliation proceeds.

yankay:
HI @ErikJiang 
Thanks for adding the legacy ownership migration. Could there still be an upgrade compatibility gap if the old controller stops after deleting a drained LWS but before deleting its DS-owned Service?

The new controller no longer lists Services for orphan cleanup, so an old-revision Service without a matching LWS remains indefinitely. I reproduced this with a controller test on the current head.

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
ErikJiang: Fixed. All controller-owned LWS deletion paths now use foreground propagation, so dependent Services are removed before the LWS disappears.

pkg/controllers/disaggregatedset/service_manager.go
ErikJiang: Added PrivateServiceName and PrivateServiceSuffix. `prv` is the existing suffix for the private headless Service, distinguishing it from the Service using the bare LWS name.

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
ErikJiang: Renamed it to deleteLegacySlice0. Recreation happens later through the normal reconciliation flow.

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
ErikJiang: Fixed. This path now goes through the shared deleteLWS helper, which uses foreground deletion.

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
ErikJiang: Centralized this in DisaggregatedSetReconciler.deleteLWS. It validates or transfers Service ownership and then foreground-deletes the LWS. All production deletion paths now use this helper, while LeaderWorkerSetManager remains independent of ServiceManager.

test/e2e/disaggregatedset/e2e_test.go
```diff
@@ -200,6 +200,11 @@ var _ = Describe("DisaggregatedSet E2E Tests", Ordered, func() {
 			By("waiting for rolling update to complete")
 			kubectl.ForSingleActiveRevision(deploymentName, oldRevision)
 
+			By("verifying the old revisions Services are garbage collected via their owning LWS")
```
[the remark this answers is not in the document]
ErikJiang: fixed.

pkg/controllers/disaggregatedset/disaggregatedset_controller.go
```diff
@@ -589,50 +600,78 @@ func (r *DisaggregatedSetReconciler) cleanupDrainedLWS(ctx context.Context, disa
...
+					return false, fmt.Errorf("failed to delete legacy service %s-prv: %w", legacyName, err)
+				}
+				continue
+			}
+			owner := metav1.GetControllerOf(service)
+			if owner == nil || owner.APIVersion != leaderworkersetv1.GroupVersion.String() || owner.Kind != "LeaderWorkerSet" || owner.Name != legacyName {
+				return false, fmt.Errorf("legacy service %s is not controlled by the expected LeaderWorkerSet %s", service.Name, legacyName)
+			}
+			waitingForGC = true
+			continue
```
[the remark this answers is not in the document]
ErikJiang: Fixed. The migration now reads the legacy LWS directly and returns a clear error when a live foreign LWS owns the Service. It waits only when the expected owner is actually gone. Regression tests cover both foreign-LWS and foreign-Service collisions.

ErikJiang:
@ahg-g @yankay Thanks for the detailed feedback. I’ve addressed the remaining issues:

- All controller-owned LWS deletion paths now use one shared helper that validates or transfers Service ownership, then performs foreground deletion.
- Legacy slice-0 migration waits until the legacy Service reaches NotFound before creating slice-aware siblings.
- Upgrade migration now removes DS-owned private Services whose matching LWS is already gone, while leaving foreign or ownerless Services untouched.
- Added the Service-name helper, renamed the legacy deletion function, and covered the collision and migration cases with regression tests.

PTAL. Thanks!

pkg/controllers/disaggregatedset/service_manager.go
```diff
@@ -47,10 +47,62 @@ func NewServiceManager(k8sClient client.Client, scheme *runtime.Scheme) *Service
...
+			continue
+		}
+
+		if lws := lwsByService[service.Name]; lws != nil {
+			if _, controlled, err := manager.transferServiceOwnership(ctx, deployment, lws); err != nil {
+				return err
+			} else if !controlled {
+				return fmt.Errorf("legacy service %s could not be transferred to LeaderWorkerSet %s", service.Name, lws.Name)
+			}
+			continue
```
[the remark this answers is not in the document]
ErikJiang: done.

ahg-g:
@ErikJiang the test seems to be consistently failing, could it be related to this PR?

ahg-g:
hopefully https://github.com/kubernetes-sigs/lws/pull/1026 addresses the flake

/retest

ahg-g:
/lgtm
/approve

Thanks!

PR #977, opened 2026-08-11
Add serviceMonitor.extraLabels to values and extend serviceMonitor template to use them

k-caps:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it
As it exists now, the chart hardcodes its servicemonitor. If an existing Prometheus only scrapes servicemonitors with specific labels, this allows simple injection of those labels for gitops style deployment.

#### Special notes for your reviewer
Tested with local `helm lint` and `helm template --debug`

#### Does this PR introduce a user-facing change?
```release-note
Allows injecting custom extraLabels to servicemonitor
```

PR #1005, opened 2026-08-25
Emit linkTitle and weight from the genref template

BenjaminBraunDev:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

#989 added `linkTitle` and `weight` frontmatter to the auto-generated API reference files, but the genref template doesn't emit those fields, so `make verify` regenerates them without it and `pull-lws-verify-main` now fails for any PR based on current main. This adds the two fields to the template. Verified `make generate-apiref` now produces a clean diff against the checked-in files.

**Does this PR introduce a user-facing change?**

```release-note
NONE
```

### DIFF
--- a/hack/genref/markdown/pkg.tpl
+++ b/hack/genref/markdown/pkg.tpl
@@ -5,6 +5,8 @@
   {{- if .IsMain -}}
 ---
 title: {{ .Title }}
+linkTitle: {{ .Title }}
+weight: {{ if eq .GroupName "leaderworkerset.x-k8s.io" }}10{{ else }}20{{ end }}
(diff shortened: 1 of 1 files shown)


aliceco01:
Independently verified this fix: checked out this branch, ran `make generate-apiref`, and `git diff --exit-code site/` comes back clean, byte-identical to what's checked in for both `leaderworkerset.v1.md` (weight 10) and `disaggregatedset.v1.md` (weight 20). Also confirmed this is currently breaking `pull-lws-verify-main` on unrelated PRs whose CI reruns against current `main` (#997, #977).

PR #1007, opened 2026-08-26
docs: restore autoscaling coverage lost in the examples restructure

aliceco01:

Reworked on top of #1021. The original version of this PR extracted inline YAML into `site/static/examples/` and pulled it back with the `{{< include >}}` shortcode. #1021 solved that same problem a different way, with canonical files under `docs/examples/<workload>/<scenario>/` referenced by `curl`, and deleted both pages this PR was editing. That approach is gone and this PR now follows the merged convention instead.

What is left is the material the restructure dropped. The new `leaderworkerset/autoscaling/` guide covers vLLM and SGLang scaling on CPU. The old HPA page also had:

- a minimal deployment needing no GPU and no Hugging Face token, so the behavior can be seen on kind
- memory-based and CPU-plus-memory HPA variants, the latter with `behavior` tuning
- why resource requests are mandatory, how to generate load the HPA actually reads, and the common failure modes

**What this adds**

Three manifests alongside the existing engine examples, following the same one-file bundle convention:

| File | Contents |
|---|---|
| `nginx.yaml` | LeaderWorkerSet + CPU HPA, no GPU required |
| `nginx-memory-hpa.yaml` | memory-based HPA |
| `nginx-multi-metric-hpa.yaml` | CPU + memory HPA with `behavior` tuning |

The two variants replace rather than supplement the CPU HPA from `nginx.yaml`, since two HPAs pointing at one LeaderWorkerSet fight over its replica count. Both the manifests and the page say so explicitly.

The prose is folded back into `site/content/en/docs/examples/leaderworkerset/autoscaling/_index.md` as sections on resource requests, alternative metrics, load generation, troubleshooting, and cleanup. Scaling behavior is called out for LWS specifically: a replica is a whole group, so an aggressive scale-down tears down a leader and its workers together.

**Testing**

All three manifests validate with `kubectl apply --dry-run=client`. Shortcode and code-fence nesting in the page is balanced.

**Note on review state**

The force-push to rework this dropped the earlier `/lgtm`. The change is substantially different from what was reviewed in August, so it needs a fresh look rather than a reinstated label.

/cc @yankay

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/examples/leaderworkerset/autoscaling/nginx-memory-hpa.yaml
+++ b/docs/examples/leaderworkerset/autoscaling/nginx-memory-hpa.yaml
@@ -1,0 +1,26 @@
+# Memory-based autoscaling for the nginx example in this directory.
+#
+# This REPLACES the CPU HorizontalPodAutoscaler shipped in nginx.yaml. Two HPAs
+# targeting one LeaderWorkerSet will fight over its replica count, so delete
+# the CPU one first:
+#
+#   kubectl delete hpa lws-hpa
+#   kubectl apply -f nginx-memory-hpa.yaml
+apiVersion: autoscaling/v2
+kind: HorizontalPodAutoscaler
+metadata:
+  name: lws-memory-hpa
+spec:
+  scaleTargetRef:
+    apiVersion: leaderworkerset.x-k8s.io/v1
+    kind: LeaderWorkerSet
+    name: leaderworkerset-sample
+  minReplicas: 1
+  maxReplicas: 5
+  metrics:
+    - type: Resource
+      resource:
+        name: memory
+        target:
+          type: Utilization
+          averageUtilization: 70
--- a/docs/examples/leaderworkerset/autoscaling/nginx-multi-metric-hpa.yaml
+++ b/docs/examples/leaderworkerset/autoscaling/nginx-multi-metric-hpa.yaml
@@ -1,0 +1,53 @@
+# CPU + memory autoscaling, with scaling behavior tuned to damp oscillation,
+# for the nginx example in this directory.
+#
+# The HPA computes a desired replica count per metric and takes the largest, so
+# either metric alone can scale the LeaderWorkerSet up.
+#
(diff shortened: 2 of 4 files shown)


yankay:
Thanks @aliceco01 

From the perspective of extracting YAML from Markdown into standalone downloadable files, this PR looks reasonable and follows the JobSet/Kueue site pattern.
/lgtm

aliceco01:
Thanks @yankay

Small heads up: the `/lgtm` doesn't seem to have registered. 
Could you post `/lgtm` as a new comment on its own line when you get a
chance? Thanks!

aliceco01:
Friendly ping. This has had `lgtm` since Aug 31 and CI is green, it just needs an `/approve` from a `site/OWNERS` approver to merge.

@ahg-g would you have a moment? It is docs-only: the YAML manifests are moved out of the markdown into `site/static/examples/` and pulled back in with the existing `{{< include >}}` shortcode, so the rendered pages are unchanged.

yindia:
Hey @aliceco01, I just merged a [PR](https://github.com/kubernetes-sigs/lws/pull/1021) that restructures the examples. Please rebase your branch and align your example with the new structure.

aliceco01:
Rebased onto #1021 and reworked to match the new structure.

@yindia the original approach here (YAML in `site/static/examples/` pulled in with `{{< include >}}`) is dropped entirely, since #1021 covers that with `docs/examples/<workload>/<scenario>/` and `curl`. Both pages this PR edited were deleted in that restructure.

What it does now is fill the gaps the restructure left in the autoscaling guide: a no-GPU nginx example so the behavior is reachable without a Hugging Face token, memory and CPU-plus-memory HPA variants, and the resource-request, load-generation and troubleshooting notes from the old HPA page. Title and description updated to match.

The force-push dropped the `/lgtm`. That is the right outcome here, the change is substantially different from what was reviewed in August and should get a fresh read rather than the label back.

PR #601, opened 2025-08-05
fix: don't increase API surface to allow resizing

synthe102:
#### What type of PR is this?
/kind cleanup

#### What this PR does / why we need it

Following the discussion in #557, relaxing the webhook to allow resizing doesn't justify increasing the API surface. This PR cleans up the `resizePolicy` field but keeps the tests ensuring resizing works.

#### Which issue(s) this PR fixes

Fixes #

#### Does this PR introduce a user-facing change?

```release-note

```

PR #1008, opened 2026-08-26
Document resizing groups via leaderWorkerTemplate.size

almogtavor:
#### What type of PR is this?

/kind documentation

`.spec.leaderWorkerTemplate.size` became mutable in KEP-552 (#557, #576, #601), but the website never documented it. `pod-templates` only says that `size` defines the group, and `rollout-strategy` covers template rollouts. So a user has no way to learn from the docs that the field can be edited at all, and no way to predict what happens when they do.

This adds a Resizing Groups concept page under `concepts/leaderworkerset/` covering:

- that `size` can be changed on a running LeaderWorkerSet
- that a resize is not in-place: `size` is part of `leaderWorkerTemplate`, so changing it cuts a new `ControllerRevision` and every pod in every group is recreated, leaders included. The page explains why, which is that each pod's `LWS_GROUP_SIZE` is populated at creation time from the size annotation and cannot be updated in place
- that the replacement runs through the configured `rolloutStrategy`, so `maxUnavailable`, `maxSurge` and `partition` all apply
- the subgroup interaction, which is currently a trap: `subGroupSize` is immutable while `size` is not, so a resize that breaks the divisibility rule is rejected with an error reported against `subGroupSize`, naming a field the user did not touch
- the `size: 1` boundary, where no worker StatefulSet exists

Also adds a one-line cross-link from the subgroups and pod templates pages.

None.

Two things I checked against the code rather than against the existing docs, worth a reviewer's eye:

1. The `size: 1` section follows the controller, not the API docstring. `leaderworkerset_types.go` says that at `size: 1` a "0-replica StatefulSet for the workers" is created, but `pod_controller.go` returns before creating one. The page says no worker StatefulSet exists. If that reading is right, the docstring is stale and I am happy to fix it in a separate PR, since it is an API comment change and does not belong in a docs PR.

2. The page states that going from N to 1 removes the worker StatefulSet. That follows from each worker StatefulSet being owned by its leader pod, so replacing the leader garbage-collects it, rather than from any explicit deletion path.

I deliberately did not quote the webhook's rejection verbatim, since the value in the `field.Invalid` call is a `*int32` and I did not want to guess how it renders. The page quotes only the detail string.

```release-note
NONE
```

### DIFF
--- a/site/content/en/docs/concepts/leaderworkerset/pod-templates/_index.md
+++ b/site/content/en/docs/concepts/leaderworkerset/pod-templates/_index.md
@@ -10,7 +10,7 @@ LeaderWorkerSet supports defining distinct specifications for leader and worker 
 
 ## Leader and Worker Templates
 
-An LWS replica consists of one leader pod and *N - 1* worker pods (where *N* is defined by `.spec.leaderWorkerTemplate.size`).
+An LWS replica consists of one leader pod and *N - 1* worker pods (where *N* is defined by `.spec.leaderWorkerTemplate.size`). `size` can be changed on a running LeaderWorkerSet, which recreates every pod in every group. See [Resizing](../resizing/).
 
 You can configure pod specifications using two template fields:
 
--- a/site/content/en/docs/concepts/leaderworkerset/resizing/_index.md
+++ b/site/content/en/docs/concepts/leaderworkerset/resizing/_index.md
@@ -36,9 +36,21 @@ If you need to add capacity without disturbing pods that are already serving, sc
 
 ## Resizing With Subgroups
 
-`subGroupSize` is immutable, but `size` is not. The two are validated against each other on every update, so a resize that breaks the [subgroup sizing rules](../subgroups/) is rejected by the webhook with `size or size - 1 must be divisible by subGroupSize`.
+`subGroupSize` is immutable, but `size` is not, and the two are validated against each other on every update. A resize that breaks the [subgroup sizing rules](../subgroups/) is rejected, and which sizes are valid depends on the subgroup policy type.
 
-The rejection is reported against `spec.leaderWorkerTemplate.subGroupPolicy.subGroupSize`, even though `size` is the field you changed. With `subGroupSize: 8`, valid sizes are 8, 9, 16, 17, 24, 25 and so on. To move to a size that does not fit, recreate the LeaderWorkerSet with both fields set together.
+Under `LeaderWorker`, the default, either `size` or `size - 1` must be divisible by `subGroupSize`. With `subGroupSize: 8` that allows 8, 9, 16, 17, 24, 25 and so on, and anything else is rejected with:
+
+```
+size or size - 1 must be divisible by subGroupSize
+```
+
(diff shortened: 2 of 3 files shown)


aliceco01 (commented):
Thanks for writing this up, the "resize is a rollout, not a scale" framing is the part users most need and it comes across clearly.

Two things I hit while checking this against the webhook:

**The `Resizing With Subgroups` example does not hold for `LeaderExcluded`.** The "valid sizes are 8, 9, 16, 17, 24, 25" list is correct for the default `LeaderWorker` policy, but `validateUpdateSubGroupPolicy` adds a separate check for `LeaderExcluded` (`pkg/webhooks/leaderworkerset_webhook.go:255-258`) requiring `size-1` to be divisible on its own. So under `LeaderExcluded` with `subGroupSize: 8`, `size: 8` is rejected and only 9, 17, 25 are valid, and the message differs: `size-1 must be divisible by subGroupSize when using LeaderExcluded`. Might be worth splitting the example by policy type, since the page already links to the subgroups doc where both types are described.

**Error path capitalization.** The page says the rejection is reported against `spec.leaderWorkerTemplate.subGroupPolicy.subGroupSize`, but the webhook builds the path with `SubGroupPolicy` capitalized, so the emitted error reads `spec.leaderWorkerTemplate.SubGroupPolicy.subGroupSize`. The JSON tag is `subGroupPolicy` (`api/leaderworkerset/v1/leaderworkerset_types.go:185`), so the webhook looks like the thing that is actually inconsistent here, but until that is fixed the doc will not match what users see. Worth a separate issue?

Everything else I spot-checked lines up: the `leaderworkerset.sigs.k8s.io/size` annotation name, the "read once at pod creation" behaviour (`pkg/utils/pod/pod_utils.go:148` errors if the annotation is missing, so an existing pod genuinely cannot learn a new size), and `weight: 65` slotting between rollout-strategy and failure-handling without a collision.

almogtavor:
@aliceco01 thanks! I'll fix these issues

almogtavor:
Thanks, both confirmed and both fixed in 173426f.

On LeaderExcluded you are right my original was actively misleading. `size: 8` with `subGroupSize: 8` passes the general check on line 249 since `8%8==0`, then gets rejected by the LeaderExcluded check on line 255 on `(8-1)%8 != 0`, so the only message the user sees is the one I did not quote. The section now has a paragraph per policy type, each with its own list of valid sizes and its own error, and calls out explicitly that `size: 8` is valid under the default and rejected under `LeaderExcluded`.

On the capitalization, agree that the webhook is inconsistent. Worth noting the path is inconsistent within itself: `leaderWorkerTemplate` and `subGroupSize` are both serialized casing and only the middle segment is the Go field name. opened #1015 to change all seven paths to `subGroupPolicy` rather than an issue.

That leaves the two PRs coupled: this page currently documents the capitalized path, which is correct today and wrong the moment #1015 merges. I can merge this as-is and let #1015 carry the one line doc update.

almogtavor:
@ahg-g rebased on current main and the documented field path now matches the casing #932 settled on, so this is ready for a look whenever you have a moment.

PR #1014, opened 2026-08-30
Migrate config/default from deprecated commonLabels to labels

### LINKED ISSUE #866: config/default: migrate deprecated commonLabels to labels (kustomize v5)

aliceco01:
Migrates `config/default/kustomization.yaml` from the deprecated `commonLabels` field to `labels` with `includeSelectors: true`, preserving the selector injection that `commonLabels` performed implicitly.

`kustomize build config/default` output is byte-identical before and after this change, so no CRD or chart regeneration is required.

Note: `includeSelectors: true` is required here. A plain `labels:` migration (as `kustomize edit fix` would produce) is metadata-only and would drop the label from selectors, mutating immutable selector fields on already-deployed Deployments and Services.

Fixes #866

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -9,9 +9,13 @@ namespace: lws-system
 namePrefix: lws-
 
 # Common labels to add to all resources and selectors (matches Helm behavior).
-commonLabels:
-  app.kubernetes.io/name: lws
-  app.kubernetes.io/instance: lws
+# includeSelectors is required to preserve the selector injection that the
+# deprecated commonLabels field performed implicitly.
+labels:
+- includeSelectors: true
+  pairs:
+    app.kubernetes.io/name: lws
+    app.kubernetes.io/instance: lws
 
 resources:
 - ../crd


aliceco01:
Friendly ping on this one. All CI is green and it merges cleanly against main. The change is confined to `config/default/kustomization.yaml` and `kustomize build config/default` output is byte-identical before and after, so there is no CRD or chart regeneration to review.

@yankay you kicked off testing here, would you or @ahg-g have a moment to take a look?

Issue #942, opened 2026-07-28
Add a new annotation/label to indicate the leader vs worker in LWS


YuleiF:
**What would you like to be added**: A new annotation/label to indicate the leader vs worker for sts managed by LWS

**Why is this needed**: 
Improved Observability & Filtering: Allows users, custom controllers, and operational scripts to easily query, inspect, or filter leader vs. worker StatefulSets via kubectl selectors without having to parse pod-level details or infer status based on naming conventions.

Flexibility for External Integrations: Enables downstream ecosystem tools (such as custom schedulers, monitoring agents, or gang scheduling controllers) to quickly identify the role of each managed StatefulSet and apply role-specific logic.

**Completion requirements**: 

This enhancement requires the following artifacts:

- [ ] Design doc
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

Dasmat13:
+1, having an explicit role label on managed StatefulSets/Pods makes filtering a lot cleaner. 

I'd be glad to take this on and open a PR for it if you'd like!

yindia:
@Dasmat13 @YuleiF  I dug into the code a bit to see how hard this would be, and the good news is the role is basically already there in the StatefulSets, it's just not spelled out anywhere, so you can't reliably select on it.

Right now the leader StatefulSet gets `leaderworkerset.sigs.k8s.io/name` and the revision hash, and the worker ones additionally carry `group-index` and `group-key`. So in practice the only way to tell them apart today is "the worker STS has a `group-index` label and the leader one doesn't", which is exactly the kind of implicit thing this issue is asking us to stop relying on.

What I'd suggest is just adding one controller-applied label:

```
leaderworkerset.sigs.k8s.io/role = leader | worker
```

It follows the same convention as the labels the controller already sets, so there's no API/CRD/webhook change involved — we'd just add it in the two places the StatefulSet labels are built. I'd lean towards putting it on the pod templates too, so `kubectl get pods -l leaderworkerset.sigs.k8s.io/role=leader` works the same way.

A few things I'd want a maintainer's call on before opening a PR:

- Is `.../role` the name you want, or something like `.../set-role`?
- Are `leader`/`worker` fine as the values?
- StatefulSets only, or should pods (and maybe the headless Services) get it too for consistency?


Happy to put up the PR once the naming/scope is settled, it's a pretty small change plus tests and a docs line.

PR #1017, opened 2026-08-31
Keep role label on StatefulSet metadata, not pod templates

### LINKED ISSUE #1016: Adding role labels to Pod templates causes workload rollout during controller upgrade
**What happened**:

PR #1002 added the `leaderworkerset.sigs.k8s.io/role` label to managed StatefulSets and their Pod templates.

When upgrading from LWS v0.9.0 to a controller containing #1002, reconciliation adds the label to `.spec.template.metadata.labels` of existing StatefulSets. This changes the Pod template and triggers a rollout even though the LeaderWorkerSet spec is unchanged.

In the upgrade E2E, all six existing workload Pod UIDs changed. Replacing a leader Pod also caused its worker StatefulSet and Pods to be recreated.

#1002 also changed the shared validator to require the new label during the old-controller phase. Since v0.9.0 does not produce it, the test initially fails before reaching the upgrade check. Making that validation backward-compatible exposes the actual rollout.

#993 does not cause this behavior; it only fixes the test harness so the upgrade job can run.

A controller upgrade should preserve existing workload Pods when the LeaderWorkerSet spec has not changed.

The requirement in #942 is to identify managed leader and worker StatefulSets. Keeping the role label on StatefulSet metadata satisfies that requirement without modifying Pod templates.

1. Install LWS v0.9.0 and create a LeaderWorkerSet.
2. Record the workload Pod UIDs.
3. Upgrade to a controller containing #1002.
4. Wait for reconciliation and compare the StatefulSet generations and Pod UIDs.

The StatefulSet generations increase and the Pods are replaced.

For PR CI, the upgrade job must be triggered explicitly:

```text
/test pull-lws-test-e2e-upgrade-main
```

Suggested fix:

- Keep the role label on managed StatefulSet metadata.
- Do not inject it into StatefulSet Pod templates.
- Keep upgrade validation compatible with resources created by older controllers.
- Verify that Pod UIDs and restart counts remain unchanged in the v0.9.0-to-current upgrade E2E.

Related: #281, #797, and #926.

**Environment**:

- Kubernetes version: v1.36.1
- LWS version: v0.9.0 to `v0.10.0-19-g0a4f9cb`
- Cloud provider or hardware configuration: Kind
- OS: Ubuntu 24.04.4 LTS
- Kernel: 7.0.0-30-generic
- Install tools: Kind v0.32.0, kubectl v1.36.1, Kustomize v5.8.1
- Others: Introduced by commit `1929dd614e8da71854537dd1f6e7e1aa3345c3c9`

yindia:
#### What type of PR is this?

/kind failing-test
/kind regression

#1002 added the `leaderworkerset.sigs.k8s.io/role` label to managed StatefulSets and their pod templates. Adding it to the pod template mutates the template, so on a controller upgrade the existing workloads roll even when the LeaderWorkerSet spec has not changed. In the upgrade e2e all existing Pods were replaced.

Selecting leader vs worker StatefulSets (the goal of #942) only needs the label on StatefulSet metadata. This PR keeps the label on leader and worker StatefulSet metadata and drops it from both pod templates, so an upgrade no longer rolls running workloads. The shared validator is updated to match, which also lets it validate workloads created by an older controller.

It also removes a leaking `RETURN` trap in `hack/e2e-test.sh`. The trap fired again when the caller returned, with `$archive` out of scope, and aborted the upgrade job under `set -o nounset` before any test ran.

Fixes #1016. Related to #942. Broken job: pull-lws-test-e2e-upgrade-main.

The role label is no longer applied to Pods. Leader vs worker Pods can still be told apart by `worker-index` (0 is the leader). If pod-level selection by role is wanted later, it should be stamped by the pod webhook, not the pod template, to avoid the rollout.

```release-note
NONE
```

### DIFF
--- a/hack/e2e-test.sh
+++ b/hack/e2e-test.sh
@@ -122,7 +122,6 @@ function kind_load_image {
     local image="$1"
     local archive
     archive="$(mktemp)"
-    trap 'rm -f "$archive"' RETURN
     docker save "$image" -o "$archive"
 
     while IFS= read -r node; do
--- a/pkg/controllers/leaderworkerset_controller.go
+++ b/pkg/controllers/leaderworkerset_controller.go
@@ -794,7 +794,6 @@ func constructLeaderStatefulSetApplyConfiguration(lws *leaderworkerset.LeaderWor
 		leaderworkerset.WorkerIndexLabelKey: "0",
 		leaderworkerset.SetNameLabelKey:     lws.Name,
 		leaderworkerset.RevisionKey:         revisionKey,
-		leaderworkerset.RoleLabelKey:        leaderworkerset.RoleLeader,
 	})
 	podAnnotations := make(map[string]string)
 	podAnnotations[leaderworkerset.SizeAnnotationKey] = strconv.Itoa(int(*lws.Spec.LeaderWorkerTemplate.Size))
--- a/pkg/controllers/leaderworkerset_controller_test.go
+++ b/pkg/controllers/leaderworkerset_controller_test.go
@@ -118,7 +118,6 @@ func TestLeaderStatefulSetApplyConfig(t *testing.T) {
 								"leaderworkerset.sigs.k8s.io/name":                   "test-sample",
 								"leaderworkerset.sigs.k8s.io/worker-index":           "0",
 								"leaderworkerset.sigs.k8s.io/template-revision-hash": revisionKey2,
-								"leaderworkerset.sigs.k8s.io/role":                   "leader",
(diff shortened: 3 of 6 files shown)


yindia:
cc: @yankay

yindia:
@yankay Finally green. This PR fought merging harder than the workloads it protects. 🙂 Ready to merge.

ahg-g:
/approve
/lgtm

Great catch, this would have been a big regression!

PR #1021, opened 2026-09-02
Refresh examples for vLLM and SGLang

### LINKED ISSUE #1020: Refresh all the examples

Most examples we currently have are legacy and use old container versions. We should clean them up, focus on two servers, vllm and sglang only, just so we can keep them updated.

We also should organize the examples into: 
- LeaderWorkerSet: to show multi-node deployment patterns with autoscaling and topology-aware placement
- DisaggregatedSet: to show disaggregated deployments with disagg-aware rollouts, autoscaling and topology placement

**Why is this needed**:

Existing examples are all legacy and mostly not relevant.

yindia:
#### What type of PR is this?

/kind cleanup
/kind documentation

The old examples were mostly legacy and pinned outdated container images. This drops llamacpp, tensorrt-llm, and the sample manifests, keeps only vLLM and SGLang, and pins both to current releases.

Examples are now organized by API:
- LeaderWorkerSet: multi-node serving with autoscaling and topology-aware placement.
- DisaggregatedSet: prefill/decode roles with per-role rollouts, autoscaling via DisaggregatedSetRoleScaler, and topology placement.

Verified on a kind cluster: all manifests pass webhook admission and the controller reconciles them into the expected child LeaderWorkerSets, role scalers, and HPAs.

Fixes #1020

Runtime images are pinned (vllm/vllm-openai:v0.28.0, lmsysorg/sglang:v0.5.18). The disaggregated examples show the wiring; the KV connector and router setup are noted as deployment-specific.

```release-note
Refreshed the examples: removed legacy examples (llamacpp, tensorrt-llm, sample) and reorganized the vLLM and SGLang examples under docs/examples/leaderworkerset and docs/examples/disaggregatedset.
```


site/content/en/docs/examples/disaggregatedset/sglang.md
```diff
@@ -0,0 +1,41 @@
...
+This example runs SGLang in a disaggregated prefill/decode topology using
+SGLang's `--disaggregation-mode`. The prefill role starts with
+`--disaggregation-mode prefill` and the decode role with
+`--disaggregation-mode decode`.
+
+{{% alert title="A router wires the roles together" color="warning" %}}
+SGLang PD disaggregation needs a router/load balancer (for example
+`sglang-router` in mini-lb mode) to connect prefill and decode. That wiring is
+deployment-specific and omitted here. See the
+[SGLang PD disaggregation docs](https://docs.sglang.ai/references/disaggregation.html).
```
ahg-g: can we pls reference llm-d disagg guide instead here

site/content/en/docs/examples/leaderworkerset/sglang.md
```diff
@@ -0,0 +1,56 @@
+---
+title: "SGLang"
+linkTitle: "SGLang"
+weight: 2
+description: >
+  Multi-node inference with SGLang on LeaderWorkerSet.
+---
+
+[SGLang](https://docs.sglang.ai/) provides native support for distributed
+tensor-parallel inference, with no Ray required. Nodes coordinate directly through
```
ahg-g:
```suggestion
tensor-parallel inference. Nodes coordinate directly through
```

site/content/en/docs/examples/leaderworkerset/vllm.md
```diff
@@ -0,0 +1,65 @@
+---
+title: "vLLM"
+linkTitle: "vLLM"
+weight: 1
+description: >
+  Multi-node inference with vLLM on LeaderWorkerSet.
+---
+
+[vLLM](https://docs.vllm.ai/) supports distributed tensor- and pipeline-parallel
```
ahg-g: Can we pls use the non-ray mode in the example?

ahg-g (commented):
I am think we need to organize the examples in a way that allows us to show different features, like autoscaling and topology-aware placement. What if we organize them as the following (where sglang/vllm continue to be presented as different yamls but the README.md for each guide is shared:

Image

yindia:
@ahg-g I've updated the example as you suggested. Please take a look and let me know what you think. Once we've finalized the structure and the example, I'll go through and verify all the examples again

ahg-g:
> @ahg-g I've updated the example as you suggested. Please take a look and let me know what you think. Once we've finalized the structure and the example, I'll go through and verify all the examples again

Thanks, looks great!

yindia:
Thanks! Now that the structure is settled, I verified all the examples on a kind cluster running the controller built from this branch.

**Control plane (all 14 examples):**
- 14/14 pass strict schema validation against the CRDs.
- 14/14 are admitted by the webhooks.
- 14/14 reconcile: child LeaderWorkerSets are created, the `DisaggregatedSetRoleScaler` is auto-created for the External-scaling example, and multi-slice fans out to 4 child LWS.

**Wiring and behavior (verified against generated pods):**
- The `LWS_*` env vars resolve as expected: leader gets `rank=0`, `nnodes=<size>`, `master=<leader-fqdn>`; workers get `rank=1`. So the non-Ray vLLM and SGLang command args are wired correctly.
- Topology-aware placement injects the expected affinity: the LWS `exclusive-topology` annotation produces pod anti-affinity on the topology key, and the DisaggregatedSet `placementPolicy` produces pod affinity plus anti-affinity.
- End-to-end group formation, Service routing, scaling via the LWS scale subresource, and a RollingUpdate all work (checked with a CPU stand-in image, since the real images need GPUs).

**What I could not run:** actual model serving. Every example requires `nvidia.com/gpu` and I do not have GPU access, so the vLLM/SGLang containers never run. On kind the pods stay Pending with `Insufficient nvidia.com/gpu`. If someone with GPUs wants a real smoke test, `disaggregatedset/basic` needs the fewest (2 GPUs); the vLLM LeaderWorkerSet examples need 16 (tp=8, pp=2, 405B).

Open question: should the vLLM basic example use a smaller model than 405B so it is runnable on a more common GPU count?

yankay:
HI @yindia 
Thanks for refreshing the examples and reorganizing the documentation. The new structure looks clear to me. A few points:

1. All eight DisaggregatedSet examples omit `readinessProbe`, while the LeaderWorkerSet examples already include it. Could we add readiness probes to the prefill and decode containers so they are not marked ready before the serving endpoints are available?
2. The sentence “KEDA can target the role's child LeaderWorkerSet directly” appears incorrect. Could we point users to `DisaggregatedSetRoleScaler` instead? Direct changes to the child LWS replica count would be overwritten by the DisaggregatedSet controller.
3. [[KEP-849](https://github.com/kubernetes-sigs/lws/blob/main/keps/849-DisaggregatedSet-HPA/README.md?plain=1#L55)](https://github.com/kubernetes-sigs/lws/blob/main/keps/849-DisaggregatedSet-HPA/README.md?plain=1#L55) still references the old HPA page, which this PR removes. Could we update that link as part of the move?

non-blocking suggestions:

- Queue depth or in-flight requests, such as `vllm:num_requests_waiting` or suitable EPP metrics, may represent inference load better than CPU utilization. Integrating those metrics could expand the scope, for example through KEDA or an existing llm-d setup, so this seems suitable for a follow-up.

yindia:
@yankay Good point, queue-depth / in-flight signals (e.g. `vllm:num_requests_waiting`, or EPP metrics) do model inference load better than CPU. The CPU-based HPA here is meant as the minimal, dependency-light example (just metrics-server). Metric-driven scaling via KEDA is already shown in the `DisaggregatedSetRoleScaler` concepts, which uses `vllm_num_requests_waiting`. Agreed this is a good follow-up rather than in-scope for this refresh — I'll open a tracking issue to add a queue-depth autoscaling example (KEDA / llm-d). wdyt ?

ahg-g:
A couple of suggestions if I may:

1) It would be nice if we can accompany each guide with a visual/diagram that demonstrates what is being deployed.
2) for sglang and vllm, we can present the two options using tabs instead of listing them back-to-back?

yindia:
@ahg-g Thanks. Tabs are done for vLLM/SGLang. I'll open a separate follow-up issue for the diagrams and pick that up next, if that works for you

ahg-g:
/lgtm
/approve

This is dropping LWS with Kueue TAS, can we follow up with a guide for that (also DisaggSet with Kueue TAS)?

Issue #1022, opened 2026-09-02
[DisaggregatedSet] Support scaling during rolling updates


hasB4K:
**What would you like to be added**:

Allow a DisaggregatedSet role to scale up or down while a rolling update is active. Scale-up should add replicas to the new revision. Scale-down should remove old-revision replicas first, then shrink the new revision only when necessary.

The rollout must preserve the planner minimum and availability limits. Every role in an old revision must remain alive until that revision can be drained as a unit. This should build on the Spec/Ready accounting and replica-fraction planning introduced by #907.

**Why is this needed**:

DisaggregatedSet rollouts can take several minutes to a few hours depending on which model need to startup, and how long it takes to terminate old ones and how many replicas each roles have. There is effort in vLLM and other inference engine for that time to be reduced (Model Express is one), but we still need to consider that it might still be too slow to allow blocking scaling during a rolling update.

During that time, HPA, KEDA, or an operator must still be able to react to changing load. Today scale-up can wait for the new revision to become fully Ready, while scale-down is deferred until the rollout completes.

**Completion requirements**:

- [x] KEP / design doc
- [ ] Controller implementation
- [ ] Tests and documentation

/kind feature

PR #1023, opened 2026-09-02
KEP-1022: DisaggregatedSet scaling during rolling updates

hasB4K:
#### What type of PR is this?

/kind feature
/kind documentation

Adds KEP-1022 describing how DisaggregatedSet roles can scale up or down while a rolling update is active.

Scale-up adds capacity to the new revision without waiting for every pending replica. Scale-down removes old-revision capacity first while preserving the planner floor, availability limits, and the invariant that every role remains alive until an old revision can be retired as a unit.

The design builds on the Spec/Ready accounting and replica-fraction planning introduced by #907. This PR contains the design only; implementation will follow separately.

Refs #1022

This PR introduces no API or controller changes.

```release-note
NONE
```

### DIFF
--- a/keps/1022-DisaggregatedSet-Scaling-During-Rollout/README.md
+++ b/keps/1022-DisaggregatedSet-Scaling-During-Rollout/README.md
@@ -1,0 +1,542 @@
+# KEP-1022: DisaggregatedSet Scaling During Rolling Updates
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Current behavior](#current-behavior)
+  - [What PR #907 changes](#what-pr-907-changes)
+  - [Goals](#goals)
+  - [Non-goals](#non-goals)
+- [Proposal](#proposal)
+  - [Scale up during a rollout](#scale-up-during-a-rollout)
+  - [Scale down during a rollout](#scale-down-during-a-rollout)
+  - [If the target changes again](#if-the-target-changes-again)
+- [Design details](#design-details)
+  - [Safety rules](#safety-rules)
+  - [How one reconcile works](#how-one-reconcile-works)
+  - [How replica fractions help](#how-replica-fractions-help)
+  - [Safe old-revision drain](#safe-old-revision-drain)
+  - [Multiple old revisions](#multiple-old-revisions)
+  - [Completion and status](#completion-and-status)
+  - [Required code changes](#required-code-changes)
+  - [API and compatibility](#api-and-compatibility)
+  - [Test plan](#test-plan)
+    - [Unit tests](#unit-tests)
+    - [Integration tests](#integration-tests)
+    - [End-to-end tests](#end-to-end-tests)
+  - [Graduation criteria](#graduation-criteria)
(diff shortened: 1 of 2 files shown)


keps/1022-DisaggregatedSet-Scaling-During-Rollout/kep.yaml
```diff
@@ -0,0 +1,17 @@
+title: DisaggregatedSet scaling during rolling updates
+kep-number: 1022
+authors:
+  - "hasB4K"
+status: provisional
+creation-date: 2026-09-02
+reviewers:
+  - TBD
+approvers:
+  - TBD
```
ahg-g, line 10: update these please

keps/1022-DisaggregatedSet-Scaling-During-Rollout/kep.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+see-also:
+  - "/keps/766-DisaggregatedSet"
+  - "/keps/849-DisaggregatedSet-HPA"
+
+# The target maturity stage in the current dev cycle for this KEP.
+stage: alpha
```
ahg-g, line 17: are you planning to gate it behind a feature flag?

keps/1022-DisaggregatedSet-Scaling-During-Rollout/README.md
```diff
@@ -0,0 +1,542 @@
...
+The equality for new Spec is important. The existing `new Spec >= target`
+check assumes that new Spec never shrinks; it would incorrectly call a
+downscaled rollout complete while excess replicas still exist.
+
+`DisaggregatedSetRoleScaler.status.replicas` continues to report observed
+replicas across all revisions. It can temporarily be higher than Spec while
+pods terminate. The DisaggregatedSet remains Progressing until it reaches the
+latest target and old revisions are gone.
+
+### Required code changes
```
ahg-g, line 409: are you planning to put this new behavior behind an opt-in flag initially or will this behavior just be enabled by default form the get go?

ahg-g (commented):
looks good to me, but how do

anneheartrecord:
One thing I could not find in the KEP: when the target changes mid-rollout, what happens to the initial-replicas annotation on the new-revision LWS objects? GetTotalInitialReplicasPerRole (pkg/utils/disaggregatedset/utils.go) prefers that annotation and only falls back to spec.replicas when it is missing, so if the annotation is not rewritten on a target change the planner's min(initial, target) would keep using the old value. Is the intent to update the annotation, ignore it for the new side, or something else?

hasB4K:
> One thing I could not find in the KEP: when the target changes mid-rollout, what happens to the initial-replicas annotation on the new-revision LWS objects?

Yes, good point. This PR is dependant of #907. Regarding initial-replicas annotation, it is continually updated to the latest resolved target in #907 (even if the autoscaler update the number of replicas). On main, it is saved only just before a rollout, which is not ideal.

We also need since #907 since I think it will be easier to implement #1022 with the coordination window.

PR #1026, opened 2026-09-03
fix(test): prevent scaler race in DisaggregatedSet HPA e2e tests

ahg-g:
Sequential specs in test/e2e/disaggregatedset/hpa_test.go shared the same DisaggregatedSet and DisaggregatedSetRoleScaler names ("test-hpa" and "test-hpa-prefill") in an Ordered container.

When the first spec completed, CleanupDeployment deleted the DS but did not delete or await garbage collection of the scaler. When the second spec started immediately after, it matched the stale, dying scaler from the first spec and scaled it to 3, while the DS controller detected the ownership mismatch, refused to adopt it (ScalerConflict), and created the LWS with 0 replicas. Once GC deleted the stale scaler, the controller created a new scaler seeded at 0, causing the test to time out waiting for replicas >= 3.

Fix this by:
1. Isolating each spec into its own Context with unique deployment and scaler names (test-hpa-get, test-hpa-scale, test-hpa-cascade), matching the convention used across the rest of the e2e test suite.
2. Removing the artificial Ordered constraint on the suite.
3. Adding explicit dsrs cleanup to kubectl.CleanupDeployment.
4. Verifying ownerRef when waiting for the scaler to exist.

Fixes #

```release-note

```

### DIFF
--- a/test/e2e/disaggregatedset/hpa_test.go
+++ b/test/e2e/disaggregatedset/hpa_test.go
@@ -38,105 +38,131 @@ import (
 //     HPA reads /scale before writing; a 500 here deadlocks the HPA loop).
 //   - kubectl scale on the scaler propagates through to the underlying LWS.
 //   - Deleting the parent DS cascades and removes the scaler via the ownerRef.
-var _ = Describe("DisaggregatedSet HPA External Scaling", Ordered, func() {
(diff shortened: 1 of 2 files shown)


hasB4K (approved):

hasB4K:
Thanks for the patch!
/lgtm

PR #1029, opened 2026-09-07
Guard nil replicas and size in LeaderWorkerSet webhook validation

aliceco01:
#### What type of PR is this?

/kind bug

`generalValidate` reads `replicas` and `leaderWorkerTemplate.size` as bare pointer dereferences, and `validateUpdateSubGroupPolicy` dereferences `size` again on the path reached when `subGroupPolicy` is set. None of the three is nil-guarded, so an object that reaches validation without its defaults applied panics the webhook instead of being accepted or rejected normally.

The two fields differ in how they are defaulted, and neither is safe on its own:

| field | CRD schema default | set by `Default()` |
| --- | --- | --- |
| `spec.replicas` | `+kubebuilder:default=1` | yes |
| `spec.leaderWorkerTemplate.size` | `+kubebuilder:default=1` | no |

On a normal admission path the schema default keeps both non-nil, so this is hardening rather than a panic reachable from a plain `kubectl apply`. It is reachable when an object never picks the schema default up, for example a CRD installed at an older revision, or an in-process caller invoking `ValidateCreate` directly, which this repository's own tests do. `size` is the weaker of the two, since the mutating webhook does not backfill it the way it backfills `replicas`.

Both are now read through a nil check that falls back to `1`, matching the schema default. No behavior change for any currently-passing case: the validation rules are unchanged, and only previously-panicking states now produce a clean result.

Fixes #

Left blank deliberately. See the note below on the overlap with the change that already landed.

- **Scope of this PR shrank after `d1a1de0` landed.** This branch originally also fixed the `subGroupPolicy` nil dereference and the `size % subGroupSize` divide-by-zero. Both were fixed upstream by #932, so those hunks are gone and only the `replicas` and `size` hardening remains. It has been rebased onto current `main` and the description rewritten to match. Reviewing it fresh rather than against the earlier discussion will be less confusing.
- `d1a1de0` replaced the previous `if lws.Spec.Replicas != nil` guard in `generalValidate` with a bare `replicas := *lws.Spec.Replicas`, relying on `Default()` backfilling it. That is sound on the admission path. This restores the guard so the function does not depend on defaulting having run, which also makes it safe for the direct callers.
- New test file `pkg/webhooks/leaderworkerset_webhook_nil_defaults_test.go` calls `ValidateCreate` directly, no envtest needed, and wraps the call in `recover()` so a regression surfaces as a normal failed assertion rather than crashing the test binary.
- All three tests fail against unpatched `main` with `runtime error: invalid memory address or nil pointer dereference`, and pass here. Verified on a clean worktree at `546910f` with only the test file added:

  ```
  --- FAIL: TestValidateCreateNilReplicas (0.00s)
  --- FAIL: TestValidateCreateNilSize (0.00s)
  --- FAIL: TestValidateCreateNilSizeWithSubGroupPolicy (0.00s)
  ```

```release-note
Fixed a bug where the LeaderWorkerSet validating webhook could panic instead of
returning a validation result when `spec.replicas` or
`spec.leaderWorkerTemplate.size` reached validation unset.
```

### DIFF
--- a/pkg/webhooks/leaderworkerset_webhook.go
+++ b/pkg/webhooks/leaderworkerset_webhook.go
@@ -142,18 +142,30 @@ func (r *LeaderWorkerSetWebhook) generalValidate(lws *v1.LeaderWorkerSet) field.
 	// Since the lws name is used as the name for headless service, it must be DNS-1035 compliant
 	ValidateName := apivalidation.NameIsDNS1035Label
 	allErrs := apivalidation.ValidateObjectMeta(&lws.ObjectMeta, true, apivalidation.ValidateNameFunc(ValidateName), field.NewPath("metadata"))
-	// Ensure replicas and groups number are valid
-	replicas := *lws.Spec.Replicas
-	if replicas < 0 {
-		allErrs = append(allErrs, field.Invalid(specPath.Child("replicas"), lws.Spec.Replicas, "replicas must be equal or greater than 0"))
+	// Ensure replicas and groups number are valid.
+	// replicas and size are backfilled to 1 by the CRD's OpenAPI schema default
+	// and by Default(), but this function can also run against an object that
+	// never picked either up (a CRD installed from an older revision, or a
+	// caller invoking the webhook logic directly), so neither is dereferenced
+	// unconditionally.
+	replicas := int32(1)
+	if lws.Spec.Replicas != nil {
+		replicas = *lws.Spec.Replicas
+		if replicas < 0 {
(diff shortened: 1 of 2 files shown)


aliceco01:
Closing in favour of #1034.

This was opened before #932 merged, and #932 covers the two panics it targeted (the `subGroupPolicy: {}` nil dereference and the `subGroupSize: 0` divide-by-zero), along with the `rollingUpdateConfiguration` and update-path guards. I re-ran all six tests from this branch against current `main` and they all pass, so nothing here demonstrates a live bug any more.

What was left after rebasing is the `replicas` and `size` hardening, which #932 did not cover. That is now #1034 as a self-contained change, with tests that do fail against `main`.

Issue #1030, opened 2026-09-08
Add topology-aware scheduling guides using Kueue (LWS + DisaggregatedSet)


yindia:
PR #1021 moved the topology-aware scheduling guides to native placement (LWS `exclusive-topology`, DisaggregatedSet `placementPolicy`) and deleted the old Kueue guide at `site/content/en/docs/examples/tas.md`. Native placement and Kueue TAS are separate paths, so the Kueue guides should live alongside the native ones rather than replace them. This issue tracks bringing Kueue TAS coverage back.

## Scope

1. **LWS + Kueue TAS (vLLM)** — restore the deleted guide. It was removed in #1021 and can be recovered from its last version: https://github.com/kubernetes-sigs/lws/blob/f6cae0b7773ab3cfc31f99598525a0c4f9cf3d11/site/content/en/docs/examples/tas.md
2. **LWS + Kueue TAS (SGLang)** — new guide, no prior example.
3. **DisaggregatedSet + Kueue TAS** — new guide. The only prior art is an nginx API sample from the old standalone tree (removed in #876), so there is no real inference example to restore.

## Notes

- Items 2 and 3 need authoring and validation on a real Kueue + GPU cluster before merge.
- Placement is the only difference from the existing guides; the workloads themselves match the basic vLLM and SGLang examples.

yindia:
@ahg-g The old tas.md guide only covered LWS + Kueue TAS with vLLM (from #682), so that one is a straightforward restore.

SGLang + Kueue TAS never existed. DisaggregatedSet + Kueue TAS only existed as an nginx API sample in the old standalone tree (removed in #876), not as a real inference example.

Do you want me to add SGLang and DisaggregatedSet Kueue TAS guides too, or keep this scoped to restoring the vLLM guide and handle the rest separately?

ahg-g:
I don't think sglang vs vllm matters here, we can use one of them. What is important is to cover both LWS and DisaggSet

PR #1032, opened 2026-09-08
Restore LWS Kueue topology-aware scheduling example

yindia:
#### What type of PR is this?

/kind documentation

#1021 dropped the LWS + Kueue topology-aware scheduling guide when it moved the examples to native placement. Native placement and Kueue TAS are separate paths, so this restores the vLLM guide alongside the native ones.

Part of #1030

Scoped to restoring the existing vLLM guide. SGLang and DisaggregatedSet Kueue TAS guides are tracked separately in #1030.

```release-note
NONE
```

### DIFF
--- a/docs/examples/disaggregatedset/topology-aware-scheduling/vllm-kueue.yaml
+++ b/docs/examples/disaggregatedset/topology-aware-scheduling/vllm-kueue.yaml
@@ -1,0 +1,80 @@
+# DisaggregatedSet + Kueue topology-aware scheduling: vLLM.
+# Requires Kueue with TAS and a LocalQueue named "default" (see the guide).
+apiVersion: disaggregatedset.x-k8s.io/v1
+kind: DisaggregatedSet
+metadata:
+  name: vllm-disagg
+  labels:
+    # Admit this workload through Kueue.
+    kueue.x-k8s.io/queue-name: default
+spec:
+  roles:
+    - name: prefill
+      spec:
+        leaderWorkerTemplate:
+          size: 1
+          workerTemplate:
+            metadata:
+              labels:
+                role: prefill
+              annotations:
+                # Co-locate the prefill role within one topology block.
+                # Set the value to your topology level's node label.
+                kueue.x-k8s.io/podset-required-topology: cloud.google.com/gce-topology-block
+            spec:
+              containers:
+                - name: vllm
+                  image: vllm/vllm-openai:v0.28.0 # pinned; bump to the latest release
+                  env:
+                    - name: HF_TOKEN
+                      value: $HF_TOKEN
+                  command:
+                    - sh
+                    - -c
+                    - "python3 -m vllm.entrypoints.openai.api_server --port 8080
+                       --model meta-llama/Meta-Llama-3.1-8B-Instruct
+                       --kv-transfer-config '{\"kv_connector\":\"NixlConnector\",\"kv_role\":\"kv_producer\"}'"
+                  resources:
+                    limits:
+                      nvidia.com/gpu: "1"
+                  ports:
+                    - containerPort: 8080
+                  readinessProbe:
+                    tcpSocket:
+                      port: 8080
+                    initialDelaySeconds: 15
+                    periodSeconds: 10
+    - name: decode
+      spec:
+        leaderWorkerTemplate:
+          size: 1
+          workerTemplate:
+            metadata:
+              labels:
+                role: decode
+              annotations:
+                # Co-locate the decode role within one topology block.
+                kueue.x-k8s.io/podset-required-topology: cloud.google.com/gce-topology-block
+            spec:
+              containers:
+                - name: vllm
+                  image: vllm/vllm-openai:v0.28.0 # pinned; bump to the latest release
+                  env:
+                    - name: HF_TOKEN
+                      value: $HF_TOKEN
+                  command:
+                    - sh
+                    - -c
+                    - "python3 -m vllm.entrypoints.openai.api_server --port 8080
+                       --model meta-llama/Meta-Llama-3.1-8B-Instruct
+                       --kv-transfer-config '{\"kv_connector\":\"NixlConnector\",\"kv_role\":\"kv_consumer\"}'"
+                  resources:
+                    limits:
+                      nvidia.com/gpu: "1"
+                  ports:
+                    - containerPort: 8080
+                  readinessProbe:
+                    tcpSocket:
+                      port: 8080
+                    initialDelaySeconds: 15
+                    periodSeconds: 10
--- a/docs/examples/leaderworkerset/topology-aware-scheduling/vllm-kueue.yaml
+++ b/docs/examples/leaderworkerset/topology-aware-scheduling/vllm-kueue.yaml
@@ -1,0 +1,107 @@
+# LeaderWorkerSet + Kueue topology-aware placement: vLLM.
+apiVersion: leaderworkerset.x-k8s.io/v1
+kind: LeaderWorkerSet
+metadata:
+  name: vllm
+  labels:
+    # Kueue admits this workload through the "default" LocalQueue.
+    kueue.x-k8s.io/queue-name: default
+spec:
+  replicas: 1
+  leaderWorkerTemplate:
+    size: 2
+    # RecreateGroupAfterStart recreates the whole group on a pod restart, but
+    # only after the group has reached initial deployment stability.
+    restartPolicy: RecreateGroupAfterStart
+    leaderTemplate:
+      metadata:
+        labels:
+          role: leader
+        annotations:
+          # Groups the leader and workers of this replica into one Kueue
+          # scheduling unit so they are admitted and placed together.
+          kueue.x-k8s.io/podset-group-name: "vllm-multi-host"
+          # Require the whole group inside one topology domain (all-or-nothing:
+          # Kueue only admits the group if it fits entirely within one block).
+          kueue.x-k8s.io/podset-required-topology: "cloud.google.com/gce-topology-block"
+      spec:
+        containers:
+          - name: vllm-leader
+            image: vllm/vllm-openai:v0.28.0 # pinned; bump to the latest release
+            env:
+              - name: HF_TOKEN
+                value: $HF_TOKEN
+            command:
+              - vllm
+              - serve
+              - zai-org/GLM-4.6
+              - --port
+              - "8080"
+              - --tensor-parallel-size
+              - "8"
+              - --pipeline-parallel-size
+              - "2"
+              - --nnodes
+              - $(LWS_GROUP_SIZE)
+              - --node-rank
+              - $(LWS_WORKER_INDEX)
+              - --master-addr
+              - $(LWS_LEADER_ADDRESS)
+              - --trust-remote-code
+            resources:
+              limits:
+                nvidia.com/gpu: "8"
+            ports:
+              - containerPort: 8080
+            readinessProbe:
+              tcpSocket:
+                port: 8080
+              initialDelaySeconds: 15
+              periodSeconds: 10
+            volumeMounts:
+              - mountPath: /dev/shm
+                name: dshm
+        volumes:
+          - name: dshm
+            emptyDir:
+              medium: Memory
+              sizeLimit: 15Gi
+    workerTemplate:
+      metadata:
+        annotations:
+          kueue.x-k8s.io/podset-group-name: "vllm-multi-host"
+          kueue.x-k8s.io/podset-required-topology: "cloud.google.com/gce-topology-block"
+      spec:
+        containers:
+          - name: vllm-worker
+            image: vllm/vllm-openai:v0.28.0 # pinned; bump to the latest release
+            command:
+              - vllm
+              - serve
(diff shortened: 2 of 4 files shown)


yindia:
cc: @ahg-g

ahg-g:
can we put it under the existing topology-aware scheduling guide? basically present two options, one that uses the placement policies and one that uses kueue and discuss when to use each approach.

ahg-g:
Trivy is failing also

yankay:
Thanks for adding this example. As I understand it, Kueue applies TAS independently to the generated LeaderWorkerSets, so prefill may be scheduled in topology block A while decode is scheduled in block B. Could we clarify this behavior to avoid implying that the two roles are co-located?

docs/examples/leaderworkerset/topology-aware-scheduling/vllm-kueue.yaml
```diff
@@ -0,0 +1,81 @@
...
+      metadata:
+        labels:
+          role: leader
+        annotations: 
+          kueue.x-k8s.io/podset-required-topology: "cloud.google.com/gce-topology-block"
+          kueue.x-k8s.io/podset-group-name: "vllm-multi-host"
+      spec:
+        containers:
+          - name: vllm-leader
+            image: vllm/vllm-openai:v0.8.5
```
ahg-g: this is a very old image, [pls see the llm-d guides](https://github.com/llm-d/llm-d/blob/add_namespace_as_parameter_for_cicd/docs/api-reference/artifacts.md#3-model-servers-and-extensions), we are now at 26. it would be great if we can update all the guides to use the latest image.

docs/examples/leaderworkerset/topology-aware-scheduling/vllm-kueue.yaml
```diff
@@ -0,0 +1,81 @@
...
+kind: LeaderWorkerSet
+metadata:
+  name: vllm
+  labels:
+    kueue.x-k8s.io/queue-name: default
+spec:
+  replicas: 1
+  leaderWorkerTemplate:
+    size: 2
+    restartPolicy: RecreateGroupOnPodRestart
```
ahg-g: here and in all other guides, I think we should recommend RecreateGroupAfterStart

site/content/en/docs/examples/leaderworkerset/topology-aware-scheduling/_index.md
```diff
@@ -33,4 +48,54 @@ curl https://raw.githubusercontent.com/kubernetes-sigs/lws/refs/heads/main/docs/
 {{% /tab %}}
 {{% /tabpane %}}
 
+## Option 2: Kueue topology-aware scheduling
+
+Use this when Kueue already manages quota and admission for your cluster.
```
ahg-g, line 53: pls reference https://kueue.sigs.k8s.io/docs/concepts/topology_aware_scheduling/

docs/examples/leaderworkerset/topology-aware-scheduling/vllm-kueue.yaml
```diff
@@ -0,0 +1,81 @@
...
+    leaderTemplate:
+      metadata:
+        labels:
+          role: leader
+        annotations: 
+          kueue.x-k8s.io/podset-required-topology: "cloud.google.com/gce-topology-block"
+          kueue.x-k8s.io/podset-group-name: "vllm-multi-host"
```
ahg-g, line 23: pls document that this is only to group the leader and workers into one scheduling unit.

docs/examples/leaderworkerset/topology-aware-scheduling/vllm-kueue.yaml
```diff
@@ -0,0 +1,81 @@
...
+              - name: HUGGING_FACE_HUB_TOKEN
+                valueFrom:
+                  secretKeyRef:
+                    name: hf-secret
+                    key: hf_api_token
+            command:
+              - sh
+              - -c
+              - "bash /vllm-workspace/examples/online_serving/multi-node-serving.sh leader --ray_cluster_size=$(LWS_GROUP_SIZE);
+                python3 -m vllm.entrypoints.openai.api_server --port 8080 --model deepseek-ai/DeepSeek-R1 --tensor-parallel-size 8 --pipeline-parallel-size 2 --trust-remote-code --max-model-len 4096"
```
ahg-g: I would switch to glm 5.3

ahg-g (commented):
I think we need to clarify what kueue TAS offers compared to the baked in exclusive-placement feature, I think for the most part is gives us all-or-nothing behavior when using "kueue.x-k8s.io/podset-required-topology"

ahg-g:
> Thanks for adding this example. As I understand it, Kueue applies TAS independently to the generated LeaderWorkerSets, so prefill may be scheduled in topology block A while decode is scheduled in block B. Could we clarify this behavior to avoid implying that the two roles are co-located?

This is true. I think only when we have the integration with CompositePodGroup we can express scheduling constraints to colocate the prefill and decode (apart from the slice placement policy which we have already)

docs/examples/leaderworkerset/topology-aware-scheduling/vllm-kueue.yaml
yindia: Switched to `zai-org/GLM-4.6` — couldn't find a "GLM 5.3" release, so I used the latest GLM. Happy to change the tag if you meant a specific one.

site/content/en/docs/examples/disaggregatedset/topology-aware-scheduling/_index.md
```diff
@@ -33,3 +46,56 @@ curl https://raw.githubusercontent.com/kubernetes-sigs/lws/refs/heads/main/docs/
 
 See the [placement policy concepts](../../../concepts/disaggregatedset/placement-policy/)
 for the full set of placement types.
+
+## Option 2: Kueue topology-aware scheduling
+
+Kueue admits the DisaggregatedSet and asks the scheduler to pack each role
+(prefill and decode) into one topology domain. Use this when Kueue already
+manages quota and admission for your cluster.
```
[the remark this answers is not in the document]
yindia: Intentional. Per @ahg-g's request the two guides are folded into one topology-aware-scheduling page presenting native placement vs Kueue TAS, so the DisaggregatedSet Kueue option belongs here alongside its native counterpart rather than in #1030.

site/content/en/docs/examples/leaderworkerset/topology-aware-scheduling/_index.md
```diff
@@ -3,21 +3,36 @@ title: "Topology-aware scheduling"
...
+bandwidth matters. The bandwidth between pods depends on how close their nodes
+sit in the data center. Topology-aware scheduling keeps each replica group
+within one topology domain to maximize it, which raises bandwidth for tensor and
+pipeline parallelism.
+
+There are two ways to do this. Pick based on how your cluster admits workloads:
+
+- **Native placement policies** — LWS's built-in
+  `leaderworkerset.sigs.k8s.io/exclusive-topology` annotation. No extra
+  components. Use this when LWS schedules directly against the cluster.
```
ahg-g, line 21: I think here we need to clarify that this is limited to exclusive 1:1 placement of lws replicas (pod group) to a domain, while Kueue allows binpacking more than one group in a domain. That is the key thing the Kueue enables for us.

site/content/en/docs/examples/leaderworkerset/topology-aware-scheduling/_index.md
```diff
@@ -33,4 +48,57 @@ curl https://raw.githubusercontent.com/kubernetes-sigs/lws/refs/heads/main/docs/
...
 
+## Option 2: Kueue topology-aware scheduling
+
+Use this when Kueue already manages quota and admission for your cluster.
+
+[Kueue topology-aware scheduling (TAS)](https://kueue.sigs.k8s.io/docs/concepts/topology_aware_scheduling/)
+places the group through `kueue.x-k8s.io/podset-required-topology`. Unlike the
+native `exclusive-topology` feature, this is **all-or-nothing**: Kueue only
+admits the group if it fits entirely within one topology domain, and holds it
+otherwise. Native placement instead schedules the group without that admission
```
ahg-g, line 59: clarify that unlike LWS's native exclusive-topology feature, more than one group can be placed into a topology domain

docs/examples/disaggregatedset/topology-aware-scheduling/vllm-kueue.yaml
```diff
@@ -0,0 +1,80 @@
...
+      spec:
+        leaderWorkerTemplate:
+          size: 1
+          workerTemplate:
+            metadata:
+              labels:
+                role: prefill
+              annotations:
+                # Co-locate the prefill role within one topology block.
+                # Set the value to your topology level's node label.
```
ahg-g, line 22: This is true, but only in the case where we have a size of 1 (since kueue will see the different roles as independent statefulsets), but this is not generalizable.

I suggest we make this example similar to the LWS one; document that Kueue is currently not integrated directly with DisaggregatedSet (i.e., Kueue is not aware of the concept of a disagg Role or Slice), but since Kueue is integrated with LWS, we can apply the same annotations to get all-or-nothing scheduling and binpacking more than on LWS groups into a topology domain.

site/content/en/docs/examples/disaggregatedset/topology-aware-scheduling/_index.md
```diff
@@ -3,19 +3,32 @@ title: "Topology-aware scheduling"
...
+to raise that bandwidth.
+
+There are two ways to do this. Pick based on how your cluster admits workloads:
+
+- **Native placement policies** — `spec.placementPolicy` built into
+  DisaggregatedSet. No extra components. Use this when LWS schedules directly
+  against the cluster.
+- **Kueue** — hand admission and placement to
+  [Kueue topology-aware scheduling (TAS)](https://kueue.sigs.k8s.io/docs/concepts/topology_aware_scheduling/).
+  Use this when Kueue already manages quota and admission for your cluster.
```
ahg-g, line 21: Clarify that unlike the LeaderWorkerSet API, the DisaggregatedSet API doesn't currently integrate directly with Kueue; however, one can use Kueue to manage the underlying LWS pod groups: colocate a pod group into one topology domain, and allow more than one group to coexist into the same topology domain

yindia: Done in 6211bdc. Updated the intro bullet and the Option 2 opening to state that the DisaggregatedSet API doesn't integrate directly with Kueue; instead Kueue manages the underlying LWS pod groups (colocate a pod group into one topology domain, allow multiple groups to coexist in the same domain).

ahg-g: Thanks a lot, very clear now! The other comments are not addressed yet, do you still plan to address them?

ahg-g:
Hi @yindia , any chance to address the remaining couple of comments?

site/content/en/docs/examples/leaderworkerset/topology-aware-scheduling/_index.md
yindia: The Kueue TAS link is referenced in both the intro bullet and the Option 2 body.

docs/examples/leaderworkerset/topology-aware-scheduling/vllm-kueue.yaml
yindia: Documented in the annotation comment: podset-group-name groups the leader and workers into one Kueue scheduling unit.

site/content/en/docs/examples/leaderworkerset/topology-aware-scheduling/_index.md
yindia: Done. Bullet now states native = exclusive 1:1 group→domain and Kueue can binpack >1 group into one domain.

site/content/en/docs/examples/leaderworkerset/topology-aware-scheduling/_index.md
yindia: Done. Option 2 now clarifies that, unlike native exclusive-topology, Kueue can place more than one group into a topology domain.

docs/examples/disaggregatedset/topology-aware-scheduling/vllm-kueue.yaml
yindia: Done. Header comment now mirrors the LWS example: Kueue is not integrated directly with DisaggregatedSet, but the same LWS annotations give all-or-nothing scheduling and binpacking of LWS groups into a domain.

yindia: Good point on size > 1 not being generalizable. Updated in e0f1328: each role now carries `kueue.x-k8s.io/podset-group-name` so Kueue groups the role's leader+workers into one LWS pod group (all-or-nothing + binpacking) rather than treating them as independent pods, mirroring the LWS example.

ahg-g:
ah, Trivy is failing

PR #1034, opened 2026-09-09
Guard nil replicas and size in leaderworkerset webhook validation

aliceco01:

`generalValidate` reads `replicas` as a bare dereference, relying on the CRD schema default and on `Default()` having run. `leaderWorkerTemplate.size` is dereferenced unguarded in both `generalValidate` and `validateUpdateSubGroupPolicy`, and is not backfilled by `Default()` at all.

An object that reaches validation without those defaults, from a CRD installed at an older revision or from a caller invoking `ValidateCreate` directly, panics the webhook instead of being accepted or rejected normally.

Both are now read through a nil check that falls back to `1`, matching the schema default.

**Relationship to #932**

This is a follow-up on the same function. #932 fixed the `subGroupPolicy` nil-pointer and divide-by-zero panics, and moved `replicas` from a guarded read to a bare one, on the basis that `Default()` now backfills it. That holds on the normal admission path but not when validation runs without defaulting, which is the case these three tests cover.

The existing `nil replicas should be defaulted` test does not catch it: it calls `Default()` before `ValidateCreate`, so the bare dereference always sees a backfilled value.

**Testing**

Three tests calling `ValidateCreate` directly, with a `recover()` so a regression fails as a normal assertion instead of crashing the test binary. All three panic against `main` without this change:

```
--- FAIL: TestValidateCreateNilReplicas (0.00s)
    ValidateCreate panicked: runtime error: invalid memory address or nil pointer dereference
--- FAIL: TestValidateCreateNilSize (0.00s)
    ValidateCreate panicked: runtime error: invalid memory address or nil pointer dereference
--- FAIL: TestValidateCreateNilSizeWithSubGroupPolicy (0.00s)
    ValidateCreate panicked: runtime error: invalid memory address or nil pointer dereference
```

Supersedes #1029, which was opened before #932 merged and is now mostly redundant with it.

/cc @ahg-g

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/webhooks/leaderworkerset_webhook.go
+++ b/pkg/webhooks/leaderworkerset_webhook.go
@@ -142,18 +142,30 @@ func (r *LeaderWorkerSetWebhook) generalValidate(lws *v1.LeaderWorkerSet) field.
 	// Since the lws name is used as the name for headless service, it must be DNS-1035 compliant
 	ValidateName := apivalidation.NameIsDNS1035Label
 	allErrs := apivalidation.ValidateObjectMeta(&lws.ObjectMeta, true, apivalidation.ValidateNameFunc(ValidateName), field.NewPath("metadata"))
-	// Ensure replicas and groups number are valid
-	replicas := *lws.Spec.Replicas
-	if replicas < 0 {
-		allErrs = append(allErrs, field.Invalid(specPath.Child("replicas"), lws.Spec.Replicas, "replicas must be equal or greater than 0"))
+	// Ensure replicas and groups number are valid.
+	// replicas and size are backfilled to 1 by the CRD's OpenAPI schema default
+	// and by Default(), but this function can also run against an object that
+	// never picked either up (a CRD installed from an older revision, or a
+	// caller invoking the webhook logic directly), so neither is dereferenced
+	// unconditionally.
+	replicas := int32(1)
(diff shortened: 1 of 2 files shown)


pkg/webhooks/leaderworkerset_webhook.go
```diff
@@ -142,18 +142,30 @@ func (r *LeaderWorkerSetWebhook) generalValidate(lws *v1.LeaderWorkerSet) field.
...
-	// Ensure replicas and groups number are valid
-	replicas := *lws.Spec.Replicas
-	if replicas < 0 {
-		allErrs = append(allErrs, field.Invalid(specPath.Child("replicas"), lws.Spec.Replicas, "replicas must be equal or greater than 0"))
+	// Ensure replicas and groups number are valid.
+	// replicas and size are backfilled to 1 by the CRD's OpenAPI schema default
+	// and by Default(), but this function can also run against an object that
+	// never picked either up (a CRD installed from an older revision, or a
+	// caller invoking the webhook logic directly), so neither is dereferenced
+	// unconditionally.
```
ahg-g, line 150: I don't think we cases like this, we should consolidate defaulting in one place, if this is defaulted by CEL, then this is enough.

PR #1025, opened 2026-09-03
[docs] Document supported Kubernetes versions

w3lld1:
#### What type of PR is this?

/kind documentation

#### What this PR does / why we need it

I documented the tested Kubernetes support window (1.33 through 1.36) in both the README and installation guide. I linked each supported minor version to its required E2E presubmit job on TestGrid. I also removed the obsolete Kubernetes 1.26 Start Ordinal guidance and clarified when the MaxUnavailableStatefulSet feature gate is needed.

#### Which issue(s) this PR fixes

Fixes #1024

#### Special notes for your reviewer

I confirmed the four linked TestGrid tabs match the current required presubmit job names in `kubernetes/test-infra`.

Validation:

- `uvx --from codespell codespell README.md site/content/en/docs/installation/_index.md`
- `git diff --check upstream/main...HEAD`
- `mise exec hugo-extended@0.152.2 -- hugo --source site --gc --minify --destination /tmp/hermes-lws-1024-public-final`

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #1036, opened 2026-09-09
docs: state the supported Kubernetes versions

### LINKED ISSUE #1024: Document Kubernetes .33-1.36 (or 1.34-1.37) support policy

Document the supported Kubernetes versions in the README and installation guide.

- Replace the outdated Kubernetes `>= 1.26` requirement with `>= 1.33`.
- State that LWS supports Kubernetes 1.33 through 1.36, matching the four existing required E2E presubmit jobs.
- Link the corresponding TestGrid jobs.
- Follow the public support-policy pattern used by [JobSet](https://github.com/kubernetes-sigs/jobset#installation).

A similar documentation update was made in [Kueue](https://github.com/kubernetes-sigs/kueue/pull/13638).

**Why is this needed**:

The current installation documentation still lists Kubernetes 1.26 as the minimum requirement, while CI validates Kubernetes 1.33, 1.34, 1.35, and 1.36. Users currently have no clear public statement describing the tested and supported Kubernetes versions.

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [x] Docs update

/help wanted

anneheartrecord:

The installation guide still asked for Kubernetes `>= 1.26`, while CI only
exercises 1.33 through 1.36. Users had no accurate public statement of what is
actually tested.

This replaces the stale requirement with the real range and makes the claim
verifiable:

- Lists the four Kind node images the required e2e presubmits actually run
  against (`v1.33.12`, `v1.34.8`, `v1.35.5`, `v1.36.1`), taken from
  `config/jobs/kubernetes-sigs/lws/lws-presubmit.yaml` in kubernetes/test-infra
  rather than restated from memory.
- Links TestGrid so the table can be checked against reality.
- Adds the same statement to the README next to the installation pointer,
  following the pattern JobSet uses.

It also drops the note about manually enabling the Start Ordinal feature gate on
1.26 — that cannot apply once 1.33 is the floor. The two link definitions it
used (`[feature_gate]`, `[start_ordinal]`) are removed with it, since nothing
else in the file referenced them.

Not included: the `max_unavailable` note is left as-is. MaxUnavailableStatefulSet
is still alpha, so that caveat is independent of the version floor.

This PR was written in part with the assistance of generative AI.

Fixes kubernetes-sigs/lws#1024

## Special notes for your reviewer

The version table is derived from the presubmit config, not from the issue text —
worth a second pair of eyes on whether you want the exact patch versions listed
(they drift when test-infra bumps the Kind images) or only the minor versions.
Happy to switch to minors only if you prefer lower maintenance.

```release-note
NONE

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,26 @@ Read the [documentation](https://lws.sigs.k8s.io/docs/) or watch the [related ta
 
 ## Installation
 
+**Requires a Kubernetes cluster running version 1.33 through 1.36.** These are the versions
+exercised by the required end-to-end presubmit jobs; see the [support policy](#kubernetes-version-support-policy).
+
 Read the [installation guide](https://lws.sigs.k8s.io/docs/installation/) to learn more.
+
+### Kubernetes version support policy
+
+LWS is tested against the following Kubernetes versions on every pull request, via the
+required `pull-lws-test-e2e-main-1-<minor>` presubmit jobs:
+
+| Kubernetes | Kind node image used in CI |
+| --- | --- |
+| 1.33 | `kindest/node:v1.33.12` |
+| 1.34 | `kindest/node:v1.34.8` |
+| 1.35 | `kindest/node:v1.35.5` |
+| 1.36 | `kindest/node:v1.36.1` |
+
+Results for these jobs are published on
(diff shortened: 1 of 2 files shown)


ahg-g:
See https://github.com/kubernetes-sigs/lws/pull/1025

PR #1040, opened 2026-09-10
fix: bound disaggregatedset kubectl queries

### LINKED ISSUE #1039: Flaky DisaggregatedSet progressive rollout e2e timeout
**What happened**:

The `Add role with progressive rolling update` e2e test timed out in:

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_lws/1007/pull-lws-test-e2e-main-1-34/2097645231800324096

The rollout completed, but `LWSNotRevision` generated the invalid selector `revision!==<revision>`. Query errors were treated as zero, and an unbounded `kubectl` call eventually blocked until the suite timeout.

The test should detect rollout completion and all `kubectl` calls should have bounded timeouts.

Run the DisaggregatedSet progressive rollout e2e test.

PR #907 already fixes the invalid selector by using `LabelNot` (commit `d3d2ef6`). This failure is unrelated to the documentation-only PR #1007.

**Environment**:

- Kubernetes version: v1.34
- LWS version: PR #1007 (`9456ad49`)
- Cloud provider or hardware configuration: Prow Kind presubmit
- OS: See linked Prow artifacts
- Kernel: See linked Prow artifacts
- Install tools: LWS e2e test infrastructure
- Others: Job `pull-lws-test-e2e-main-1-34`

amirialy:
#### What type of PR is this?

/kind bug
/kind failing-test
/kind flake

Fixes flaky DisaggregatedSet progressive rollout E2E tests and adds bound timeouts

Fixes #1039 

NONE

```release-note
NONE
```

### DIFF
--- a/test/testutils/disaggregatedset/kubectl/queries.go
+++ b/test/testutils/disaggregatedset/kubectl/queries.go
@@ -44,7 +44,7 @@ func LWSByRevision(deploymentName, revision string) *Builder {
 func LWSNotRevision(deploymentName, revision string) *Builder {
 	return Get("lws").
 		Label(labelName, deploymentName).
-		Label(labelRevision+"!=", revision).
+		LabelNot(labelRevision, revision).
 		Namespace(defaultNS)
 }
 
@@ -98,7 +98,10 @@ func EndpointSliceByRole(deploymentName, role string) *Builder {
 
 // CountPods returns the number of pods for a deployment.
 func CountPods(deploymentName string) int {
-	output, err := Pods(deploymentName).Output("name").RunQuiet()
+	output, err := Pods(deploymentName).
+		Output("name").
+		Timeout("30s").
+		RunQuiet()
 	if err != nil {
 		return 0
 	}
@@ -107,7 +110,10 @@ func CountPods(deploymentName string) int {
 
 // CountRunningPods returns the number of running pods for a deployment.
 func CountRunningPods(deploymentName string) int {
-	output, err := RunningPods(deploymentName).NoHeaders().RunQuiet()
+	output, err := RunningPods(deploymentName).
+		NoHeaders().
+		Timeout("30s").
+		RunQuiet()
 	if err != nil {
 		return 0
 	}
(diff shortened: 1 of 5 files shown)


test/testutils/disaggregatedset/kubectl/queries.go
```diff
@@ -98,7 +98,10 @@ func EndpointSliceByRole(deploymentName, role string) *Builder {
 
 // CountPods returns the number of pods for a deployment.
 func CountPods(deploymentName string) int {
-	output, err := Pods(deploymentName).Output("name").RunQuiet()
+	output, err := Pods(deploymentName).
+		Output("name").
+		Timeout("30s").
```
anneheartrecord: `kubectl get` doesn't take `--timeout`, that flag only exists on delete/wait/rollout:

    $ kubectl get pods --timeout=30s
    error: unknown flag: --timeout

With RunQuiet errors mapped to 0/"", all eight queries in this file would return zero. The global `--request-timeout=30s` is the one that applies to get; a parallel RequestTimeout() on the Builder would do it.

anneheartrecord (commented):
Thanks for picking this up so fast. One flag issue inline that would make the queries fail in CI. Also, the LWSNotRevision hunk is the same as in #907, so it may be worth rebasing on that or dropping it here. And #1039 mentions query errors being treated as zero, is that in scope here or a follow-up?

amirialy:
> Thanks for picking this up so fast. One flag issue inline that would make the queries fail in CI. Also, the LWSNotRevision hunk is the same as in #907, so it may be worth rebasing on that or dropping it here. And #1039 mentions query errors being treated as zero, is that in scope here or a follow-up?

Hi @anneheartrecord !
For LWSNotRevision, I don't think it really matters whether I drop it and rebase on #907 or keep it here, so I'm fine with either.
For the query errors, I'd prefer to keep them in scope here. Since a failed query currently gets treated as 0/"", it can make the rollout checks think everything is fine when the query actually failed, it seems related to the issue

anneheartrecord:
Thanks for the quick turnaround. The request-timeout change and surfacing the query errors through the waiters both look right to me, and CI is green across the matrix now. My comments are addressed.

PR #1046, opened 2026-09-13
docs: render concept examples from canonical files and validate all example manifests

### LINKED ISSUE #990: Reuse canonical YAML examples in website documentation

Avoid duplicating Kubernetes YAML manifests across Markdown pages. Add a Hugo `include` shortcode and render documentation snippets from canonical sample files, such as files under `config/samples` or `site/static/examples`.

Suggested scope:

- Add an `include` shortcode for YAML examples.
- Select one canonical location for runnable samples.
- Migrate repeated LWS and DisaggregatedSet snippets incrementally.
- Make the site build fail when an included example is missing.

Prior art:

- [JobSet include shortcode](https://github.com/kubernetes-sigs/jobset/blob/main/site/layouts/shortcodes/include.html)
- [JobSet documentation usage](https://github.com/kubernetes-sigs/jobset/blob/main/site/content/en/docs/tasks/failure_policy.md)
- [Kueue include shortcode](https://github.com/kubernetes-sigs/kueue/blob/main/site/layouts/shortcodes/include.html)

**Why is this needed**:

Copied YAML can drift from the CRD schema and runnable examples. Review of #989 highlighted how repeated DisaggregatedSet snippets can diverge from the required `roles[].spec` structure.

Using a single source of truth reduces maintenance, applies fixes consistently, and allows CI to validate the same manifests shown in the documentation.

aliceco01:
#### What type of PR is this?

/kind documentation
/kind cleanup

Concept pages carried 19 inline YAML manifests that nothing checked against the API. This PR:

- Moves each full manifest to `site/static/examples/<api>/<topic>/` and renders it with the existing `include` shortcode. The shortcode fails the Hugo build when a referenced file is missing. The two partial snippets that only illustrate a single field stay inline.
- Adds `test/examples`, wired into `make test`, which decodes every manifest under `site/static/examples`, `docs/examples` and `config/samples` with strict field checking and runs LeaderWorkerSet and DisaggregatedSet objects through the admission webhooks' defaulting and validation. Kinds outside this project's scheme (KEDA, Kueue) only need to be well-formed YAML with an apiVersion and kind.
- Removes `config/samples/lws_v1alpha1_configuration.yaml`, which the new test flagged: it uses the non-existent group `lws.x-k8s.io` and wraps the fields in a `spec` block the `Configuration` type does not have. The real configuration is `config/manager/controller_manager_config.yaml`.
- Documents the convention in the contribution guidelines page.

Rendered output is unchanged apart from the download link the shortcode adds above each manifest. Runnable engine examples stay under `docs/examples` per #1021.

Fixes #990

The Netlify deploy preview is the fastest way to see the concept pages. `go test ./test/examples/... -race -count=3` passes locally; the test takes about a second.

```release-note
NONE
```


BenAyedMedAla:
Since this hasn't had `/ok-to-test` yet, I checked out the branch and ran the new suite locally before relying on the "passes locally" claim:

```
go test ./test/examples/... -count=3 -v
```

All 26 manifests across `site/static/examples`, `docs/examples`, and `config/samples` pass, including the pre-existing `docs/examples/**` and `config/samples/**` files this PR didn't touch — so the new strict-decode + webhook defaulting/validation pass doesn't just work on the freshly-extracted examples, it holds for the whole existing corpus too. `keda.sh/v1alpha1 ScaledObject` correctly falls back to syntax-only checking as designed (not in the project's scheme). `go build ./...` is also clean on this branch.

Also compared each extracted `site/static/examples/**/*.yaml` byte-for-byte against the inline block it replaced in the corresponding `_index.md` — they match, so this is genuinely a refactor with no silent content change riding along.

One thing worth double-checking before merge, unrelated to correctness: removing `config/samples/lws_v1alpha1_configuration.yaml` because it uses the non-existent group `lws.x-k8s.io` looks right (the real `Configuration` type is `config.lws.x-k8s.io/v1alpha1`, unwrapped, per `cmd/main_test.go`) — but since that sample file predates this PR, it might be worth a one-line release note (currently `NONE`) so anyone who copied that stale sample notices it's gone, rather than just silently disappearing from `config/samples`.

PR #1050, opened 2026-09-17
[feat] Add groupReplacementPolicy to serialize hash group replacement

BenjaminBraunDev:
Follow-up to #987.

## Problem

In Hash mode the leader Deployment creates a replacement leader as soon as the old one starts terminating. The new group goes through scheduling while the old group still holds its capacity, so on a full cluster it preempts running workloads. Ordinal mode does not do this: the leader StatefulSet cannot recreate a pod until the old one is fully deleted.

## Change

New optional field `spec.groupReplacementPolicy`:

- `PostTermination` (default): the replacement group waits until the old group is fully deleted, matching Ordinal behavior.
- `Immediate`: the replacement group is created right away, the previous Hash behavior. Hash mode only.

## How it works

The pod webhook adds a scheduling gate to every Hash leader at creation. A gated pod is invisible to the scheduler, so it takes no node and triggers no preemption.

When the pod controller reconciles a gated leader it counts, within the same LeaderWorkerSet:

- `T`: groups still tearing down (any pod of the group exists while its leader is terminating or gone).
- `G`: gated leader pods, oldest first.

It admits the oldest `G - T` leaders by removing the gate and defers the rest, creating nothing for a deferred group (no worker StatefulSet) and requeuing after 10s. Deleting or creating any pod in the set re-queues its gated leaders, so admission happens as soon as the last pod of an old group is gone.

`T` counts pods, not just the leader object, because the leader is not always deleted foreground. A group restart deletes it foreground (leader outlives workers), but a rollout or scale down deletes it through the ReplicaSet in the background, so the leader can vanish a second before its workers release capacity.

Under `Immediate` the controller skips the count and removes the gate on first reconcile. The webhook runs on create only, so it never re-adds a gate the controller removed. DisaggregatedSet passes the field through and syncs it in place, since changing it does not roll pods.

## What is in the diff

- Logic, about 255 lines: API field (37), webhook defaulting and validation (22), pod controller gate handling (150), DS sync (33), helpers and event constants (14).
- Generated, about 135 lines: CRD schema in 4 YAML files (80), client-go apply configuration (22), reference doc page (33).
- Tests, about 670 lines: unit (346), integration (251), e2e (74).
- KEP section, 39 lines.


ahg-g:
/lgtm
/approve
This is pretty clean!

Can we pls update https://github.com/kubernetes-sigs/lws/tree/main/site/content/en/docs/concepts/leaderworkerset as a followup with the hash based group identity feature? I guess we should do that right after we release.

BenjaminBraunDev:
> /lgtm /approve This is pretty clean!
> 
> Can we pls update https://github.com/kubernetes-sigs/lws/tree/main/site/content/en/docs/concepts/leaderworkerset as a followup with the hash based group identity feature? I guess we should do that right after we release.

Yes, will do. Made a draft: #1062

